PR #1, opened 2017-08-03
add stub for proposals

erikzhang:
NEP-2: Passphrase-protected private key
---------------------------------------
This NEP describes a method for encrypting and encoding a passphrase-protected private key record.

NEP-3: Wallet Standard
----------------------
This NEP designs a wallet standard that allows the wallet files to be shared between the various implementations of NEO.

NEP-4: URI Scheme
-----------------
This NEP describes a URI scheme for making NEO based payments. The difference between BIP-21 and NEP-4 is that, BIP-21 makes payments that using bitcoin itself, and NEP-4 makes payments that using any asset on NEO.

NEP-5: Token Standard
---------------------
This NEP describes standard functions a token contract can implement.

NEP-6: Superconductive Exchange
-------------------------------
This NEP describes a mechanism for NEO based decentralized exchange.

NEP-7: Dynamic Sharding
-----------------------
This NEP describes a sharding mechanism that allows contracts to be executed in parallel.

NEP-8: Compact Block Relay
--------------------------
Compact blocks on the wire as a way to save bandwidth for nodes on the P2P network.

NEP-9: Peer Authentication
--------------------------
This NEP describes a way for peers to authenticate to other peers to guarantee node ownership and/or allow peers to access additional or limited node services, without the possibility of fingerprinting.

NEP-10: Aggregate Signature
---------------------------
An aggregate signature scheme is a digital signature that supports aggregation: Given n signatures on n distinct messages from n distinct users, it is possible to aggregate all these signatures into a single short signature. This single signature (and the n original messages) will convince the verifier that the n users did indeed sign the n original messages.

NEP-11: SM2 Cryptography
------------------------
Implement the [SM2 Digital Signature Algorithm](http://www.oscca.gov.cn/UpFile/2010122214822692.pdf) for NEO.

NEP-12: Homomorphic Encryption
------------------------------
Homomorphic encryption is a form of encryption that allows computations to be carried out on ciphertext, thus generating an encrypted result which, when decrypted, matches the result of operations performed on the plaintext.

NEP-13: Zero-knowledge proof
----------------------------
In cryptography, a zero-knowledge proof or zero-knowledge protocol is a method by which one party (the prover) can prove to another party (the verifier) that a given statement is true, without conveying any information apart from the fact that the statement is indeed true.

NEP-14: Stealth Addresses
-------------------------
Stealth addresses are a way for a payer and a payee to have a private exchange of funds.

NEP-15: NeoID
-------------
Digital identity refers to the identity information of individuals, organizations, and other entities that exist in electronic form. The more mature digital identity system is based on the PKI (Public Key Infrastructure) X.509 standard. In NEO, we will implement a set of X.509 compatible digital identity standards. This set of digital identity standards, in addition to compatible X.509 level certificate issuance model, will also support Web Of Trust point-to-point certificate issuance model. Our verification of identity when issuing or using digital identities includes the use of facial features, fingerprint, voice, SMS and other multi-factor authentication methods. At the same time, we will also use the blockchain to replace the Online Certificate Status Protocol (OCSP) to manage and record the X.509 Certificate Revocation List (CRL).

NEP-16: NeoFS
-------------
NeoFS is a distributed storage protocol that utilizes Distributed Hash Table technology. NeoFS indexes the data through file content (Hash) rather than file path (URI). Large files will be divided into fixed-size data blocks that are distributed and stored in many different nodes.

NEP-17: NeoX
------------
NeoX is a protocol that implements cross-chain interoperability. NeoX is divided into two parts: "cross-chain asset exchange protocol" and "cross-chain distributed transaction protocol."

NEP-18: NeoQS
-------------
NeoQS (Quantum Safe) is a lattice-based cryptographic mechanism. At present, quantum computers do not have the ability to quickly solve the Shortest Vector Problem (SVP) and the Closest Vector Problem (CVP), which is considered to be the most reliable algorithm for resisting quantum computers.

### DIFF
--- a/README.mediawiki
+++ b/README.mediawiki
@@ -20,4 +20,106 @@ First review [[nep-1.mediawiki|NEP-1]]. Then clone the repository and add your N
 | Erik Zhang
 | Meta
 | Active
+|-
+| 2
+| Passphrase-protected private key
+| Erik Zhang
+| Standard
+| Stub
+|-
+| 3
+| Wallet Standard
+| Erik Zhang
+| Standard
+| Stub
+|-
+| 4
+| URI Scheme
+| Erik Zhang
+| Standard
+| Stub
+|-
+| 5
+| Token Standard
+| Erik Zhang, Da Hongfei
+| Standard
+| Stub
+|-
+| 6
+| Superconductive Exchange
+| Erik Zhang, Da Hongfei
+| Standard
+| Stub
+|-
+| 7
+| Dynamic Sharding
+| Erik Zhang, Da Hongfei
+| Standard
+| Stub
+|-
+| 8
+| Compact Block Relay
+| Erik Zhang
+| Standard
+| Stub
+|-
+| 9
+| Peer Authentication
+| Erik Zhang
+| Standard
+| Stub
+|-
+| 10
+| Aggregate Signature
+| Erik Zhang
+| Standard
+| Stub
+|-
+| 11
+| SM2 Cryptography
+| Erik Zhang
+| Standard
+| Stub
+|-
+| 12
+| Homomorphic Encryption
+| Erik Zhang
+| Standard
+| Stub
+|-
+| 13
+| Zero-knowledge proof
+| Erik Zhang
+| Standard
+| Stub
+|-
+| 14
+| Stealth Addresses
+| Erik Zhang
+| Standard
+| Stub
+|-
+| 15
+| NeoID
+| Erik Zhang, Da Hongfei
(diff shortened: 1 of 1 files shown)


lllwvlvwlll:
Fast Sync would also be one to add to the list. Maybe add a brief description for each (1 sentence).

canesin:
@lllwvlvwlll NEP-8 ?

a7594li:
good job

revett:
👍

PR #2, opened 2017-08-06
NEP: Passphrase-protected private key

erikzhang:
A method is proposed for encrypting and encoding a passphrase-protected private key record in the form of a 58-character Base58Check-encoded printable string. Encrypted private key records are intended for use on paper wallets. Each record string contains all the information needed to reconstitute the private key except for a passphrase, and the methodology uses salting and scrypt to resist brute-force attacks.


canesin:
With all the talk about quantum crypto shouldn't we take advantage of the clean slate and setup a standard using something like vector lattice ?

erikzhang:
This method can resist the quantum computer.
Because in the final step of decryption, the decrypter needs to compute the NEO address to verify that the decryption result is correct. In this process, hashing algorithms are not easily implemented by quantum computers.

canesin:
@erikzhang You didn't merge the PR but the branch ... Could you close this PR them ? (considering the NEP-2 design is final)

erikzhang:
Each pull request for NEP will be merged when it is in a final status. The current status of it is accepted.
You can check the work flow [here](https://github.com/neo-project/proposals/blob/master/nep-1/process.png).

PR #4, opened 2017-08-11
NEP: Token Standard 

lllwvlvwlll:


erikzhang:
Do we really need the methods (transferFrom, approve, allowance) from ERC20?

lllwvlvwlll:
They allow you to transfer funds on behalf of someone else which can be pretty valuable.  

Use case: Ability for people to set up 'joint' accounts where two users share access to the tokens in the contract while retaining their own distinct accounts.  

It could also be used as an interface mechanism for systems to interact with the tokens on behalf of someone else.  A user would 'allow' an application to access their tokens/data.

luodanwg:
Very pretty document!  I have 3 suggestions and you can consider about.
1. Deploy function.
    I  think the `deploy` function should not be the standard token function, it will depends on the specific business needs. Maybe someone need to set the ico start/end time, maybe need to set exchange rate. etc. so I think we should not control it in the protocol.

2. methods (transferFrom, approve, allowance)
   As ERC-23(https://github.com/ethereum/EIPs/issues/223) has removed these functions, so i also think retaining these methods will seems much complicated and the recent ICO seems have no use of these methods. Should we also remove them?

3. originator
   I noticed that the ‘originator’ defined as a required parameter of every function, some function such as` totalSupply`，`name`，`symbol`，`decimals`，`balanceOf ` , i think it should be public and there is no need to verify the sender. if we have neoscan(like etherscan)  later, it can be very convenient to query these public methods. so i think where need to verify where to add check of  `originator`  is better.

finally, @erikzhang , i think we also need  `Function signature description file` just like ethereum's ABI file ,which will generate when compile the smart contract, with this, the client will be very easy to support Various applications.

lllwvlvwlll:
@luodanwg 
1) I agree,  There are many mechanics by which people will want to deploy the contract so it should not be part of the standard.

2) As mentioned in the previous comment, I believe we should keep the delegated transfer functions because they allow for some potential user - application interaction that cannot easily exist otherwise. 

3) I agree with this as well.

luodanwg:
another suggestion. 
NEO is support runtime.notify func to notify the client some events. So i think we can also add the notify when Transfer or Approval is completed.

erikzhang:
Maybe we can mark some methods as "optional".

lllwvlvwlll:
Optional method and Runtime.notify() have been added to all 'authenticated' methods.

erikzhang:
What is the use of "NotifyOriginator"? Why don't we just add a "transferred" event?

luodanwg:
I have 3 suggestions
1.Good news, As i comfired with @erikzhang , NEO will also support ABI files later,  so the method should defined with much detailed parameters.  such as 
`private static bool Transfer(byte[] from,byte[] to,number value)`

2.Method totalSupply should return the value of token issued.  The hard capacity and soft capacity is the business needs(maybe have, maybe not have) ,so it should not include in our protocol. 

3.I think the Event NotifyOriginator should be `Transferred` event just like ERP-20 ,  so the sender and receiver all can catch it.
sample: `private static bool Transferred(byte[] from,byte[] to,number value)`

ps: We will also attach another NEP-5 template here later, could you also helps to check it , Thanks in advance. 😸

erikzhang:
@luodanwg agree

erikzhang:
If the transaction failed, VM will stop at a FAULT state, which can be detected easily.

Dexaran:
`approve` / `transferFrom` / `allowance` were included in ERC20 token standard to resolve EthereumVM related issues with "call stack depth attack". They should be considered deprecated and superfluous now. These functions doesn't introduce any functionality that could not be implemented without them. Call stack depth attack was resolved at the EIP150 (as you can see it happened a bit later than EIP 20 was created).

You should keep in mind that [ERC20](https://github.com/ethereum/EIPs/issues/20) token standard functions are leading to monetary losses. I wrote a number of articles about it (for example [this one](https://www.reddit.com/r/ethereum/comments/6h17aw/critical_problem_of_erc20_tokens/)).

The main problem of [ERC20](https://github.com/ethereum/EIPs/issues/20) token is that it has two ways of transferring funds:
1. `transfer`
2. `approve` + `transferFrom`

which assumes that first pattern is processed without handling by the receiver and the second pattern simulates transaction handling.

The purpose of `approve` + `transferFrom` was described in [solidity documentation](https://media.readthedocs.org/pdf/solidity/develop/solidity.pdf) at page 39: 

>Warning: There are some dangers in using send:  The transfer fails if the call stack depth is at 1024 (this can always be forced by the caller) and it also fails if the recipient runs out of gas.  So in order to make safe Ether transfers, always check the return value of send, use transfer or **even better: use a pattern where the recipient withdraws the money**

This was relevant at the moment of creating ERC20 token standard, but the stack depth attack was fixed through [EIP150](https://github.com/ethereum/EIPs/issues/150) . As the result of this `approve` + `transferFrom` must be considere **deprecated** now.

I think that there is no need to develop and implement the `approve / allowance` mechanism because:
1. `approve + transferFrom` is leading to monetary losses since it assumes another way to transfer tokens (And an alternative way to execute a transaction often does not implement handling)
2. vulnerability to re-approval attack is described [here](https://drive.google.com/file/d/0ByMtMw2hul0EN3NCaVFHSFdxRzA/view)
3. `approve` is superfluous since common pattern of transactions handling is fallback function execution. It is already described in [ERC223](https://github.com/ethereum/EIPs/issues/223).

I'd like to recommend to remove `decApprove` / `setApprove` / `getAllowance` from the token standard because of described problems of this approach.

anfn101:
We are trying to make NEP5 ERC223 compliant... and as the author of that standard, you do provide some good insight. We will take that into consideration.

luodanwg:
@lllwvlvwlll I hava attached our template at issue 6 , could you helps to review it ,thanks.
 https://github.com/neo-project/proposals/issues/6

erikzhang:
@Dexaran Strongly agree!

erikzhang:
What is ".nep-5.mediawiki.swp"?

anfn101:
I believe that's a swap file, created automatically as logging by some text editors. It can be removed without any issue. @lllwvlvwlll

nep-5.mediawiki
```diff
@@ -0,0 +1,97 @@
...
+  Created: 2017-08-10
+</pre>
+
+==Abstract==
+
+The NEP-5 Proposal outlines a token standard for the NEO blockchain that will provide systems with a generalized interaction mechanism for tokenized Smart Contracts.  This mechanic, along with the justification for each feature are defined.  A template and examples are also provided to enable the development community.
+
+==Motivation==
+
+As the NEO blockchain scales, Smart Contract deployment and invocation will become increasingly important.  Without a standard interaction method, systems will be required to maintain a unique API for each contract, regardless of their similarity to other contracts.  Tokenized contracts present themselves as a prime example of this need because their basic operating mechanism is the same.  A standard method for interacting with these tokens relieves the entire ecosystem from maintaining a definition for basic operations that a required by every Smart Contract that employs a token.
```
AshRolls: Grammar 

> A standard method for interacting with these tokens relieves the entire ecosystem from maintaining a definition for basic operations that **a** required by every Smart Contract that employs a token.

erikzhang: Can you edit it?

AshRolls: No problem, PR submitted to lllwvlvwll's repo.

lllwvlvwlll:
can we approve and Merge this?

erikzhang:
We still need more tests for its implementations.

saltyskip:
I wonder on usage of the wording "account" rather than "address" when referring to transferring funds in this document. Account implies a higher level Object than what is actually being used, and I think is a point of confusion for many developers when they begin to mix the concept of an address with the concept of an account. In blockchain projects I would typically think of an account as the combination of all relevant public and private keys (WIF, pubkey, privkey, address).

In summary think it would be more clear to refer to the byte array parameters in this document as addresses rather than accounts (which I argue is a more true representation of what they actually are).  Thoughts?

erikzhang:
Token Standard ABI:

```json
{
  "functions": [
    {
      "name": "totalSupply",
      "parameters": [],
      "returntype": "Integer"
    },
    {
      "name": "name",
      "parameters": [],
      "returntype": "String"
    },
    {
      "name": "symbol",
      "parameters": [],
      "returntype": "String"
    },
    {
      "name": "decimals",
      "parameters": [],
      "returntype": "Integer"
    },
    {
      "name": "balanceOf",
      "parameters": [
        {
          "name": "account",
          "type": "Hash160"
        }
      ],
      "returntype": "Integer"
    },
    {
      "name": "transfer",
      "parameters": [
        {
          "name": "from",
          "type": "Hash160"
        },
        {
          "name": "to",
          "type": "Hash160"
        },
        {
          "name": "amount",
          "type": "Integer"
        }
      ],
      "returntype": "Boolean"
    }
  ],
  "events": [
    {
      "name": "transfer",
      "parameters": [
        {
          "name": "from",
          "type": "Hash160"
        },
        {
          "name": "to",
          "type": "Hash160"
        },
        {
          "name": "amount",
          "type": "Integer"
        }
      ]
    }
  ]
}
```

neotracker:
It would be nice if we also had a standard set of functionality for ICOs (at least an ABI). https://github.com/neo-project/examples-csharp/blob/master/ICO_Template/ICO_Template.cs is a good start, but it'd be good if it could expose things like ico_start_time and ico_end_time. Not part of this proposal of course, but this seemed the most relevant so commenting here.

lllwvlvwlll:
@erikzhang we should also consider a 'listMethods' method which returns the definition for each supported method in the contract.

Ejhfast:
@Dexaran @erikzhang removing `approve` and `transferFrom` requires the SC to call `tokenFallback` on *receiver* contract, is that correct? given that NEO requires that all outside SC contract calls to be declared in advance for a contract, it does not seem possible for NEP-5 tokens to transfer to arbitrary contracts in NEO? (you would need to declare all possible contracts the NEP-5 would communicate with in advance?)

can you explain more? perhaps I am missing something

Dexaran:
Alright, I'll try to give a brief explanation.

### How it works with Ether. Fallback function and payable modifier.

As for Ethereum contracts and solidity, there is a fallback function that is executed whenever the call transaction does not provide a signature that matches any of the signatures of the functions implemented in this contract. Read [solidity docs](http://solidity.readthedocs.io/en/develop/contracts.html) for more detailed description. There is a special `payable` modifier that allows a function of a contract to receive funds (Ether). You can not deposit any Ether into contract that is not specially designed to receive Ether. The transaction will always fail if the function that is executed is not marked as `payable`.

If you want to create an Ethereum contract that will accept Ether deposits, you must explicitly declare a deposit function or a fallback function and use the `payable` modifier. Example:

```js

contract Receiver_of_Ether
{
    function() payable
    {
        // fallback function allows to deposit Ether to this contract.
        // payable modifier is required.
    }
}
```

This is specially done to prevent the accidentally sent Ethers being stuck inside contracts. If the contract is not explicitly allowed to receive funds then any attempt of depositing Ether will fail and the funds will not be transferred to the contract (the funds will remain at user's balance). 

### Rationale. Preventing user mistakes and losses of money.

We should keep in mind that not all users are technically advanced and some of them will try to send Ether into contracts by mistake. This is a reasonable protection from the most common user mistakes that could lead to disasterous consequences and millions of dollars lost.

### How it works with ERC223 tokens.

I followed this approach when developing the ERC223 standard. I think the contract **should** explicitly declare that it is specifically designed to receive tokens if it is going to receive tokens. Otherwise, it **must** reject token deposits by default. This is analogue of how it works with Ether.

As for Ether, contract must implement a fallback function with `payable` modifier to receive Ether.
 As for tokens, contract must implement a `tokenFallback` function to receive tokens. If the contract does not implement it, then any attempt to deposit (Ether or tokens) must fail.

```js

contract Receiver_of_Tokens
{
    function tokenFallback(address,uint256,bytes)
    {
        // tokenFallback function allows to deposit tokens to this contract.
        // no payable modifier, we do not allow user to deposit ETH, but only tokens.
    }
}
```

It must not be possible to transfer tokens to a contract that is not designed to receive them (that does not implement a special `tokenFallback` function or any other function that you agree to use as token transfer handler. The function name does not seem to be essential here).

### Mistakes of ERC-20 standard and why a contract MUST explicitly declare that it is designed to receive funds.

Unlike ERC-223 transfers or Ether transactions, [ERC-20](https://github.com/ethereum/EIPs/issues/20) transfer function does not require a receiver to implement a special function for it. A user can just send ERC-20 tokens to any address he wants. As a result, contracts that are not specifically designed to work with tokens (for example, ENS contracts and token contracts themselves) constantly receive tokens that are accidentally sent.

These tokens become stuck inside contracts without any possibility to recover them. In other words, users lose their tokens (money) this way.

[Event handling](https://en.wikipedia.org/wiki/Event_(computing)) is a common practice in programming. Transaction should be considered an event and it should be properly handled. ERC-20 lacks this transaction handling mechanism at all and this is a [known problem of ERC20 standard](https://medium.com/@dexaran820/erc20-token-standard-critical-problems-3c10fd48657b).

**It already results in lost money** (you can read about it [here](https://www.reddit.com/r/ethereum/comments/6e8y9o/the_ens_contract_becomes_token_holder_erc20/), [here](https://www.reddit.com/r/storj/comments/6t2czj/5000_are_lost_storj_erc20_tokens_are_vulnerable/) and [here](https://www.reddit.com/r/storj/comments/6ajjo3/attention_issues_of_the_upcoming_storj_migration/) ). The last time I checked it, there were about $400,000 lost in stuck ERC-20 tokens. It was a long time ago, and I believe that we are talking about millions of dollars lost at the moment.

Ejhfast:
Thanks @Drexan! This does not address the limitations of NEO I mentioned.
My concern above is that ERC223 cannot be implemented currently in NEO
given lack of dynamic SC calls

And if I am wrong and can be implemented, it at least has not been
described correctly in the current NEP-5 specification

Dexaran:
I'm not really familiar with NEO. I can only describe how it works on Ethereum and why.
As for NEO, I can highlight key points of the token standard:

1. You must disallow token deposits to contracts that are not explicitly designed to receive tokens.

2. Token deposits to contracts and deposits to addresses should behave differently.

3. You must implement a mechanism of transaction handling. This will allow to "do something on incoming token transfer".

4. Designing a couple of different functions to perform transfers to contracts and transfers to addresses is an antipattern. Someone will obviously choose the wrong way (a user will call `depositToAddress` when depositing into contract hence he will lose money).

Ejhfast:
thanks for taking the time @Dexaran! i do agree with your critique of ERC20

my point in raising this comment is that NEP5 is not following ERC223 in its current state, and possibly, *can't* follow it right now given differences between NEO and ETH. my understanding is that NEO smart contracts require that any calls between SCs to be declared statically within the contract that will do the calling. this has potential scalability benefits but may prevent a direct reimplementation of ERC223.

maybe @erikzhang can look at the above and comment further

Ejhfast:
@erikzhang an update on this:

@localhuman has confirmed with me that dynamic SC calls (calls between SCs that are not statically declared) are not allowed on the NEO VM. this means that ERC223 is currently not possible (a token needs to do a dynamic call on the receiver contract)

it is very important that NEP-5 tokens are powerful enough to support the behaviors specified by either ERC20 or ERC223. among the many other applications, decentralized exchanges will not work without these features. as it stands, unless a token implements the *optional* methods in this NEP-5, it is impossible to send it to another SC and have it do meaningful work on your behalf

we would recommend: 

(1) enabling dynamic SC calls: this would be a simple change in the VM, there is a lot of other low hanging fruit for performance before we get to sharding VM resources. there could also be something like a `dynamic` flag that indicates whether dynamic SC calls are used in a contract, to preserve scalability benefits for simpler contracts 

(2) more critically, make NEP-5 *fully compliant* with either ERC20 or ERC223. the tokens are not powerful enough for basic functionality right now 

in summary, (1) is important but (2) is absolutely necessary. it will be very bad if the ecosystem grows using the current version of NEP-5

canesin:
@luodanwg can you bring this issue to Zhengwen attention ? I think @Ejhfast  will do a PR for NEP5 Amendment with (2) making it fully compliant with ERC-20. Rgds.

erikzhang:
An important difference between Ethereum and NEO is that: in NEO, you can combine multiple invocations into one transaction, and ensure transaction consistency.
That means, you don't need a trigger to call `tokenFallback` automatically. You can call everything you want in `InvocationTransaction`.

Ejhfast:
@erikzhang if that is possible, it needs to be specified in the protocol, otherwise NEP5 tokens are broken, and will not be able to interact with SCs in interesting ways (lacking equivalent functionality to ERC223)

also, in terms of whether it is possible, `tokenFallback` should be called *conditionally* on a successful transfer, so I am not sure how this happens using the technique you propose

PR #7, opened 2017-08-21
NEP: Simulation of Consensus Nodes Assignment

lerider:
Simulation of Consensus Nodes Assignment

PR #8, opened 2017-08-21
NEP: Simulation of Consensus Nodes Assignment - Simulation file

lerider:


erikzhang:
You need to create a repo in your repositories and reference it in previous PR #7 .

PR #12, opened 2017-09-13
NEP: NeoContract ABI

erikzhang:
An application binary interface (ABI) is the interface between two program modules, one of which is often a library and/or operating system and the other one is usually an application created by a regular programmer.

This NEP describes the ABI standards for NEO smart contracts.


NeoContractABI.mediawiki
```diff
@@ -0,0 +1,137 @@
...
+  Title: NeoContract ABI
+  Author: Erik Zhang <erik@neo.org>
+  Type: Standard
+  Status: Draft
+  Created: 2017-9-13
+</pre>
+
+==Abstract==
+
+An application binary interface (ABI) is the interface between two program modules, one of which is often a library and/or operating system and the other one is usually an application created by a regular programmer.
```
totalvm: Application Binary Interface

revett:
Really nice 👌

NeoContractABI.mediawiki
```diff
@@ -0,0 +1,137 @@
...
+  "functions": [],
+  "events": []
+}
+</pre>
+
+<code>hash</code> is the script hash of the contract. It is encoded as a hexadecimal string in big-endian.
+
+<code>entrypoint</code> indicates that which is the entry point of the contract in <code>functions</code>.
+
+<code>functions</code> is an array of Function object which describes the details of each function in the contract.
```
totalvm: Function objects which describe

NeoContractABI.mediawiki
```diff
@@ -0,0 +1,137 @@
...
+
+<code>events</code> is an array of Event object which describes the details of each event in the contract.
```
totalvm: Event objects which describe

NeoContractABI.mediawiki
```diff
@@ -0,0 +1,137 @@
...
+
+===Function===
+
+Function object has the following structure:
+
+<pre>
+{
+  "name": "transfer",
+  "paramaters": [],
```
totalvm: parameters

NeoContractABI.mediawiki
```diff
@@ -0,0 +1,137 @@
...
+<code>returntype</code> indicates the return type of the function. It can be one of the following values: <code>Signature</code>, <code>Boolean</code>, <code>Integer</code>, <code>Hash160</code>, <code>Hash256</code>, <code>ByteArray</code>, <code>PublicKey</code>, <code>String</code>, <code>Array</code>, <code>InteropInterface</code>, <code>Void</code>.
+
+===Event===
+
+Event object has the following structure:
+
+<pre>
+{
+  "name": "refund",
+  "paramaters": []
```
totalvm: parameters

NeoContractABI.mediawiki
```diff
@@ -0,0 +1,137 @@
...
+{
+  "name": "transfer",
+  "paramaters": [],
+  "returntype": "Boolean"
+}
+</pre>
+
+<code>name</code> is the name of the function, which can be any valid identifier.
+
+<code>paramaters</code> is an array of Paramater object which describes the details of each paramater in the function.
```
totalvm: <code>parameters</code> is an array of Parameter objects which describe the details of each parameter in the function.

NeoContractABI.mediawiki
```diff
@@ -0,0 +1,137 @@
...
+<pre>
+{
+  "name": "refund",
+  "paramaters": []
+}
+</pre>
+
+<code>name</code> is the name of the event, which can be any valid identifier.
+
+<code>paramaters</code> is an array of Paramater object which describes the details of each paramater in the event.
```
totalvm: <code>parameters</code> is an array of Parameter objects which describe the details of each parameter in the event.

NeoContractABI.mediawiki
```diff
@@ -0,0 +1,137 @@
...
+
+===Paramater===
```
totalvm: Parameter

NeoContractABI.mediawiki
```diff
@@ -0,0 +1,137 @@
...
+
+Paramater object has the following structure:
```
totalvm: Parameter

NeoContractABI.mediawiki
```diff
@@ -0,0 +1,137 @@
...
+
+<pre>
+{
+  "name": "from",
+  "type": "Hash160"
+}
+</pre>
+
+<code>name</code> is the name of the paramater, which can be any valid identifier.
```
totalvm: parameter

NeoContractABI.mediawiki
```diff
@@ -0,0 +1,137 @@
...
+
+<code>type</code> indicates the type of the paramater. It can be one of the following values: <code>Signature</code>, <code>Boolean</code>, <code>Integer</code>, <code>Hash160</code>, <code>Hash256</code>, <code>ByteArray</code>, <code>PublicKey</code>, <code>String</code>, <code>Array</code>, <code>InteropInterface</code>.
```
totalvm: parameter

NeoContractABI.mediawiki
```diff
@@ -0,0 +1,137 @@
...
+
+===ParamaterType===
```
totalvm: ParameterType

NeoContractABI.mediawiki
```diff
@@ -0,0 +1,137 @@
...
+
+ParamaterType enum has the following values:
```
totalvm: ParameterType

NeoContractABI.mediawiki
```diff
@@ -0,0 +1,137 @@
...
+
+{|
+!name
+!description
+|-
+| Signature
+| A signature of transaction or block which is generated by user.
```
totalvm: A signature of a transaction or block which is generated by the user.

NeoContractABI.mediawiki
```diff
@@ -0,0 +1,137 @@
...
+
+This NEP describes the ABI standards for NEO smart contracts.
+
+==Motivation==
+
+NEO smart contract system is designed to be mutually invocable between contracts. To achieve this, we need a mechanism for exposing the interface of smart contracts. With NeoContract ABI, developers can easily create programs to invoke smart contracts or write clients that automatically access contract functionalities.
+
+==Rationale==
+
+We assume the Application Binary Interface (ABI) is strongly typed, known at compilation time and static. No introspection mechanism will be provided. We assert that all contracts will have the interface definitions of any contracts they call available at compile-time.
```
Sajjon: How about changing:

> We assert that all contracts will have the interface definitions of any contracts they call available at compile-time.

To:

> At runtime, we assert that all contracts will have the **same** interface definition. This is also true for any potential other contracts said contracts call.

NeoContractABI.mediawiki
```diff
@@ -0,0 +1,137 @@
+<pre>
+  NEP: <to be assigned>
+  Title: NeoContract ABI
+  Author: Erik Zhang <erik@neo.org>
+  Type: Standard
+  Status: Draft
+  Created: 2017-9-13
```
Sajjon: How about using `YYYY-MM-DD` format? :) Thus changing `2017-9-13` to `2017-09-13` if this date ever were to be parsed by some program?

NeoContractABI.mediawiki
Sajjon: Of course I meant "compile-time", I updated my comment...

neotracker:
The spec should probably specify the structure of the entry point parameters for Smart Contracts that are compliant with the spec. Currently the spec describes a set of functions and their expected parameters, but it's not clear from the ABI how one would actually invoke the entry point for a given function. Should the function name be the first parameter? The last? Seems like it should be the first parameter, but it would be good to call that out explicitly.

nep-3.mediawiki
```diff
@@ -0,0 +1,137 @@
...
+| A byte array.
+|-
+| PublicKey
+| An ECC public key which is encoded with compressed mode.
+|-
+| String
+| A string which is encoded in UTF-8.
+|-
+| Array
+| An array of objects.
```
notatestuser: Could you elaborate on what "object" means in this context? It doesn't seem to be defined as a standard type

igormcoelho:
@erikzhang I have a question regarding the scope of NEP-3, I don't know if people are already actually using it for automated processes, but we are (at Eco: eco.neocompiler.io). One small thing is causing a big issue to us (@vncoelho), perhaps it's just a mistake in C# ICO_Template (I will propose a fix for it), or perhaps we need to put more emphasis on a point here at NEP-3.  When we compile ICO_Template and generate the ABI, method names are automatically listed (following the same C# syntax), for example, "Deploy". I think the compiler and ABI is doing the right job, however, when I try to invoke the method "Deploy" (our platform does this automatically), it won't work, because the actual string operation is "deploy". So, the ABI becomes useless to us. In my opinion, in the scope of NEP-3 we should emphasize that:
- Method names in ABI should correspond to each string operation
So, a proper fix would be to have the C# ICO_template having functions with the same capitalization as the op string names (to allow the compiler to generate the ABI correctly). Do you agree?

RavenXce:
> Method names in ABI should match exactly the correspondent string operations in a smart contract
This is correct.

> So, a proper fix would be to have the C# ICO_template having functions with the same capitalization as the op string names (which are already consistent to NEP-5 spec), just to allow the compiler to correctly and automatically generate the ABI. Do you agree?
But I think the C# compiler should be improved instead (to use code annotations / comments, or have smarter derivation of operation name) instead of forcing functions to have same naming as operation string?

igormcoelho:
It already has some annotations, however, they also do not exist on ICO template... ok, I'll propose some changes (one or the other).

erikzhang:
Maybe we should make the compiler automatically turn the first letter of the method name into lowercase when generating the ABI?

igormcoelho:
@erikzhang I thought about that haha but I'm not a C# dev, but C++ and Java "standard" require methods to start lower case (and classes upper case). Isnt that the same for C#?

erikzhang:
> I thought about that haha I'm not a C# dev, but C++ and Java "standard" require methods to start lower case (and classes upper case). Isnt that the same for C#?

C# methods starts with upper case.

igormcoelho:
@RavenXce and @erikzhang, it seems the better option is to use DisplayName to adjust function names, so C# programmers can keep C# standard, while making everything 100% compatible with ABI signatures :)

PR #13, opened 2017-09-14
NEP: Wallet Standard

erikzhang:
This NEP describes a wallet standard that allows the wallet files to be shared between the various implementations of NEO.

### DIFF
--- a/WalletStandard.mediawiki
+++ b/WalletStandard.mediawiki
@@ -14,11 +14,11 @@ This NEP describes a wallet standard that allows the wallet files to be shared b
 
 ==Motivation==
 
-Currently, different client programs generate different wallet files. They have different file formats, different ways of storing, and different ways of encrypting. It is difficult for users to migrate between different client programs because the wallet files are not in the same format. Although migration can be achieved by exporting the private key, it is cumbersome for wallets with multiple private keys. We need an universal wallet standard format that allows users to migrate across all platforms safely and easily without having to change the wallet file or export the private keys.
+Currently, different client programs generate different wallet files. They have different file formats, different ways of storing, and different ways of encrypting. It is difficult for users to migrate between different client programs because the wallet files are not in the same format. Although migration can be achieved by exporting the private key, it is cumbersome for wallets with multiple private keys. We need an universal wallet format that allows users to migrate across all platforms safely and easily without having to change the wallet file or export the private keys.
 
 ==Rationale==
 
-The wallet standard should consider both security and cross-platform. For security, we require the implementations to use the NEP-2 mechanism to encrypt or decrypt the private keys. For Cross-platform, we use the JSON format to describe the wallet files, so that the contents of the wallet files can be easily recognized on each platform.
+The wallet standard should consider both security and cross-platform compatibility. For security, we require the implementations to use the NEP-2 mechanism to encrypt or decrypt the private keys. For cross-platform, we use the JSON format to describe the wallet files, so that the contents of the wallet files can be easily recognized on each platform.
 
 ==Specification==
 
--- a/WalletStandard.mediawiki
+++ b/WalletStandard.mediawiki
@@ -1,9 +1,9 @@
 <pre>
-  NEP: <to be assigned>
+  NEP: 6
   Title: Wallet Standard
   Author: Erik Zhang <erik@neo.org>
   Type: Standard
-  Status: Draft
+  Status: Accepted
   Created: 2017-09-14
   Requires: 2, 3
 </pre>
@@ -117,3 +117,5 @@ Contract object has the following structure:
 All old-format wallets should be able to be converted to this new JSON format easily. If these wallet files contain some extra data, they can be stored in the <code>extra</code> fileds.
 
 ==Implementation==
+
+*neon-js: https://github.com/CityOfZion/neon-js/blob/feature/nep-wallet/src/wallet/Wallet.js
--- a/WalletStandard.mediawiki
+++ b/WalletStandard.mediawiki
@@ -29,6 +29,7 @@ A wallet file in JSON format has the following basic structure:
 <pre>
 {
   "name": "MyWallet",
+  "version": "1.0",
   "scrypt": {},
   "accounts": [],
   "extra": null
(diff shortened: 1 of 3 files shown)


WalletStandard.mediawiki
```diff
@@ -0,0 +1,116 @@
...
+  Requires: 2, 3
+</pre>
+
+==Abstract==
+
+This NEP describes a wallet standard that allows the wallet files to be shared between the various implementations of NEO.
+
+==Motivation==
+
+Currently, different client programs generate different wallet files. They have different file formats, different ways of storing, and different ways of encrypting. It is difficult for users to migrate between different client programs because the wallet files are not in the same format. Although migration can be achieved by exporting the private key, it is cumbersome for wallets with multiple private keys. We need an universal wallet standard format that allows users to migrate across all platforms safely and easily without having to change the wallet file or export the private keys.
```
revett:
```
We need an universal wallet standard format that...
```

could change to the following, as the above reads slightly wrong:

```
We need an universal wallet format that...
```

WalletStandard.mediawiki
```diff
@@ -0,0 +1,116 @@
...
+
+==Rationale==
+
+The wallet standard should consider both security and cross-platform. For security, we require the implementations to use the NEP-2 mechanism to encrypt or decrypt the private keys. For Cross-platform, we use the JSON format to describe the wallet files, so that the contents of the wallet files can be easily recognized on each platform.
```
revett:
```
The wallet standard should consider both security and cross-platform.
```

could change to:

```
The wallet standard should consider both security and cross-platform compatibility.
```

revett:
```
For Cross-platform, we use...
```

could change to the following, as the capital `C` is not needed:

```
For cross-platform, we use...
```

WalletStandard.mediawiki
```diff
@@ -0,0 +1,116 @@
...
+  "name": "MyWallet",
+  "scrypt": {},
+  "accounts": [],
+  "extra": null
+}
+</pre>
+
+<code>name</code> is a label that the user has made to the wallet file.
+
+<code>scrypt</code> is a ScryptParameters object which describe the parameters of SCrypt algorithm used for encrypting and decrypting the private keys in the wallet.
```
revett:
```
...parameters of SCrypt algorithm...
```

could be changed to the following, as the capital `C` is not needed:

```
...parameters of Scrypt algorithm...
```

WalletStandard.mediawiki
```diff
@@ -0,0 +1,116 @@
...
+<code>extra</code> is an object that is defined by the implementor of the client for storing extra data. This field can be <code>null</code>;
+
+===Contract===
+
+Contract object has the following structure:
+
+<pre>
+{
+  "script": "21036dc4bf8f0405dcf5d12a38487b359cb4bd693357a387d74fc438ffc7757948b0ac",
+  "parameters": [],
```
localhuman: Should the returntype also be listed here?

saltyskip:
If the locked field is set to true, then does the struct still contain the value of the private key? This would seem unnecessary

ixje:
According to the **rationale** section of this document, NEP-2 is _required_. NEP-2 describes the `scrypt` parameters to be used. As such the `ScryptParameters` object in the wallet format seems to be superfluous.

erikzhang:
Clients can specify different parameters than NEP-2 for SCrypt in `ScryptParameters`.

ixje:
@erikzhang I understand that `ScryptParameters` can give flexibility in how to process the keys. If  `ScryptParameters` are used that deviate from those specified in NEP-2 then that seems to violate the following statement, no?

> For security, we **require** the implementations to use the NEP-2 mechanism to encrypt or decrypt the private keys

From the [Account.key](https://github.com/neo-project/proposals/blob/33cfa67c92df8ba226da9c27de1aa569c03e3a65/WalletStandard.mediawiki#account) description

> `key` is the private key of the account in the NEP-2 format

I derive that the purpose of the WalletFormatStandard is to only have NEP-2 compliant keys in its output (again for security and cross platform compatibility reasons). Under those circumstances another client also only needs to process NEP-2 compliant keys when importing (which makes sense to me). This brings me back to wondering why is the `ScryptParameters` key needed? The client that currently uses different scrypt parameters will just have to use the correct ones before exporting to the proposed WalletFormatStandard.

Perhaps you can sketch a scenario where the `ScryptParameters` add value. I currently can only think of such a scenario if they do not comply with the remainder of the proposed WalletFormat in the first place. By having variable `scrypt()` parameters you're basically forcing all future clients to also support variable scrypt parameters while they could just have them fixed according to the NEP-2 standard.

erikzhang:
There are 3 reasons:
1. Some devices, such as mobile phones or web pages, cannot quickly encrypt/decrypt NEP2 keys.
1. If you have a large number of private keys in your wallet, even a good performance device will take a lot of time to decrypt your wallet.
1. If the strength of the NEP2 is not enough in the future and needs to be upgraded, we do not need to update the wallet standard itself, but only modify the `ScryptParameters`.

ixje:
Thanks for the elaboration and I see your point. I'll give it some more thought in this context, as some of the current wording seems more strict that the actual intention.

erikzhang:
> Additionally it seems like all change should always be returned back to the the sending address, rather than to a "default" address. Is this a technical limitation?

The main reason for sending to the default address is that the transaction may have more than one input address. So the best thing to do now is to send change first to the default address in the input addresses, and if not, to the global default address.

snowypowers:
What is the state of `Account.contract` if the Account is a normal/watchonly account?
  - `{}` (empty object?)
  - `null` (null reference?)
  - `{script: '', parameters:[], deployed: false}` (empty Contract object?)

I believe this can be made clearer like `Account.key` (deemed `null` for watchonly/contract addresses)

erikzhang:
If an account is watchonly, then `Account.contract` should be `null`.

canesin:
@erikzhang we have the NEON team implementing it now, they will provide feedback next week if they find any enhancements or issues are raised. You can schedule the final merge for Friday 22.

erikzhang:
That's great!

erikzhang:
@canesin Any feedback from NEON team?

slipo:
@erikzhang I've been assigned to do this task for the neon-wallet team. I should be done with the initial implementation in some days. In any case, so far the proposal makes sense to me.

slipo:
Hi @erikzhang, what do you think about marking fields as mandatory or optional? I'm having an issue using neon-js with neon-wallet because neon-wallet doesn't have the public addresses. It only has the encrypted keys stored. Therefore, neon-js throws an exception when I try to use it to create the wallet with accounts lacking the addresses.

To be clear, I'm not arguing that the address field should be optional or not. But perhaps the spec should specify.

erikzhang:
The `address` field is not optional. You can derive the address from private key.

slipo:
Yeah, makes sense. It causes a small problem because we only have the encrypted key so we can't upgrade the wallets fully until the user logs into all their accounts.

Anyways, the neon-wallet NEP-6 implementation has been merged and I have no other comments at the moment. lgtm.

erikzhang:
So, I can mark this NEP as `final`, right?

PR #14, opened 2017-10-05
Simplify descriptions around key material

ixje:
also more closely matches the actual implementation
https://github.com/neo-project/neo/blob/master/neo/Wallets/KeyPair.cs#L75


ixje:
I just noticed the following in the existing text:

> without consideration for initialization vectors or **block chaining**

I now realise the reason why it is written in the original text is such, because this way using CBC mode doesn't matter (because we should be doing 16 bytes per encryption). I'm afraid us developers are likely to skip the ** Specification** section and go directly to **Encryption steps**, then thinking step 3 + 4 of that section can be combined into a single AES call. 

Perhaps the above quote could be emphasized more (it's really easy to miss)? assuming you don't want to explicitly force EBC or CBC mode.

erikzhang:
I think it should be pushed to https://github.com/neo-project/proposals/tree/nep-2

PR #19, opened 2017-11-10
NEP Dynamic Invoke

localhuman:
This NEP Proposal outlines a mechanism whereby a Smart Contract is provided the ability to invoke other Smart Contracts not known until runtime, rather than being limited to invoking only Smart Contracts that are defined at compile time. In order to retain the ability for Smart Contracts to interface with a future Dynamic Sharding process, included is a proposed specification to be used in the creation of Smart Contracts to denote whether a Smart Contract needs the Dynamic Contract Invocation feature or not.


erikzhang:
Does dynamic invocation cost more fees?

localhuman:
It is my opinion that using the `DYNAMICCALL` should cost more than `APPCALL` or `TAILCALL`.  I also believe that publishing an SC with the `HasDynamicInvoke` should cost more than publishing one without that feature.

erikzhang:
So should SC with `HasStorage` cost more than SC without `HasStorage`?

localhuman:
I think that would make sense.

erikzhang:
What about this:
1. Every contract that be deployed costs 100 gas, this is the basic cost.
1. If it has a `HasStorage` flag, the cost is increased by 400 gas.
1. If it has a `HasDynamicInvoke` flag, the cost is increased by 500 gas.

localhuman:
That sounds about right. I will update the proposal with those figures.

localhuman:
Updated the fee structure, and removed the section of code demonstrating that the cost of a `DYNAMICCALL` would cost more than an `APPCALL`.

erikzhang:
perfect

localhuman:
I've updated the NEP number in the proposal to 4 🎉 
Is it ok to begin implementation?

erikzhang:
Yes, let's do it!

erikzhang:
@localhuman Since we don't have an opcode `DYNAMICCALL`, can you edit the NEP to correct it?

localhuman:
Yes, will do!

PR #16, opened 2017-10-17
NEP: Prevent voting attacks

lerider:

PR #20, opened 2017-11-21
NEP No Vote, No Claim

lerider:
This NEP propose that addresses should only be able to claim GAS if the address has voted for consensus nodes.


nep-2.mediawiki
```diff
@@ -3,7 +3,11 @@
   Title: Passphrase-protected private key
   Author: Erik Zhang <erik@neo.org>
   Type: Standard
+<<<<<<< HEAD
+  Status: Accepted
+=======
   Status: Final
+>>>>>>> fb19052d7b9a7c1b10efd0bee7c2ebe10e26baa7
```
mhuggins: Artifacts of a bad rebase above: ☝️

nep-NoVoteNoClaim.mediawiki
```diff
@@ -0,0 +1,41 @@
...
+  Title: No vote, No claim
+  Author: Malcolm Lerider
+  Type: Standard
+  Status: Draft
+  Created: 2017-11-02
+</pre>
+
+==Abstract==
+
+This NEP suggests that gas claims should be rejected unless the address has voted for consensus votes.
```
ixje: for consensus nodes. ?

nep-NoVoteNoClaim.mediawiki
```diff
@@ -0,0 +1,41 @@
...
+
+An address only has to vote once. For following gas claims, the first vote is still valid.
+
+NEO in an address which has not voted will still receive GAS as usual, but it will stay as unavailable GAS. The claim to turn unavailable GAS into available GAS will only pass if the address has voted for consensus nodes.
+
+==Rationale==
+
+Voting should be encouraged but not enforced. It will require certain steps that a new user will need some time to figure out. Moreover, there will be passive NEO holders who does not keep themselves updated on all protocol changes. An incentive to vote without affecting these mentioned parties is needed.
+
+Alternative solution is to only allow NEO holders to receive GAS after voting. This would be an enforcement of voting and thus not optimal. Firstly, it would affect the parties mentioned above negatively. Secondly, NEO holders may not be satisfied with the nominees and should have the option to abstain vote.
```
ixje: Maybe "to abstain from voting." reads easier?

localhuman:
What would stop wallet makers from automating the process of voting so that the wallet can make claims?

nep-NoVoteNoClaim.mediawiki
lerider: Yes, typo.

nep-2.mediawiki
lerider: Thanks

lerider:
Automating the process is not necessarily bad if it is automated to vote randomly for 1 nominee. The problem lies in calculating weighted average for all ballots, no vote = no ballot. If all wallets automated it, then it would be high voting participation, and the influence with someone who holds 20% would also be close to 20%. Without votes, someone with 20% could get 50% influence if there is for example only 40% voting participation.

There is a problem if it is automated to vote for all (or many) nominees, in which case the following NEP is required:
https://github.com/neo-project/proposals/pull/16

jlqe:
My proposal is to implement voting as a pre-requisite for receiving NEO system fees (DeployContract 500 GAS etc). NEO holders have to exercise "work" through Proof of Voting, by looking through the CN list and voting for the best ones (ie CN with lowest fees, highest uptime). In return, the NEO "voters" are included in the distribution of system fees.

This has many benefits:
1) Large holders who hold NEO in Cold Storage will not be shortchanged, as they will continually receive GAS through decay algorithm. However, since they did not engage in work through voting, they are not eligible to receive system fees.

2) NEO holders will have incentive to vote for the most qualified CN. Stronger NEO network -> More usage -> Increased System fees -> Distributed equally among those who exercise "work of voting"

3) There will be an additional "abstain" scenario, if NEO holders find that none of CN meets their specifications. Since the purpose of the NEP is to increase participation, NEO holders will still have to participate to cast their vote of "abstinence"

coranos:
Only 50 people hold 67% of the NEO, of over 100,000 addresses.
Of those, only 26 have claimed gas in the last 60 days.
I don't think this proposal will meaningfully change the voting process.

lerider:
I like jlqe's suggestion if it's technically not too complex

coranos:
I see several addresses with thousands of unclaimed gas and zero NEO.

How will they be able to claim the gas, as they have no NEO to vote?

I also see several addresses with thousands of NEO that never claim gas. If they do not vote, will elections fail due to low voter turnout? Or will their vote simply not matter until they decide to vote?

canesin:
I think @jlqe proposition is better, but we could remove the abstain and add the possibility of delegating votes for a current running CN appointed address. The setting would be:

1 - Each NEO can vote in as many CN they wish (as is today, to solve the single outcome issue). So 1 NEO = number of candidates votes. Once votes are cast, they stay there until the adress votes in another CN.
2 - To earn sys fees you need to vote, if you don't vote you can still earn from the block GAS generation, but not from system usage.
3 - Each CN should appoint a address to vote in his behalth, and address can delegate the voting to CN they trust, so they can move funds to cold storage and still earn sys fee.

lerider:
@canesin Agree with all three points. Think this NEP should contain point 1 and 2. Point 3 may be a NEP on its own.

lerider:
Ready for new review.

lerider:
@coranos  @canesin  @jlqe 

Please see and review changes

erikzhang:
We need one PR one proposal.

mwherman2000:
What data are people going to base their CN votes on?

nep-NoVoteNoSysFees.mediawiki
```diff
@@ -0,0 +1,39 @@
...
+  Type: Standard
+  Status: Draft
+  Created: 2017-11-02
+</pre>
+
+==Abstract==
+
+This NEP suggests an encouragement for NEO holders to vote for consensus nodes. 
+
+GAS distribution is divided into two parts: 1. Generation of new GAS by NEO tokens and 2. Recycled system fees. In the current implementation, NEO holders will automatically receive both these distributions as unclaimed GAS. This NEP suggests that only generation of new GAS should be automatically distributed to NEO. To be eligible to receive recycled system fees, the address holding the NEO must have voted for consensus nodes. How the address has voted does not affect how much of the recycled system fees they get; it is distributed proportionally to all NEO that are held in an address that has voted.
```
vncoelho: @lerider, @vang1ong7ang mentioned that we may still faill into the prisionner dilema if we keep distributing the "inflation".

> This NEP suggests that only generation of new GAS should be automatically distributed to NEO

In this sense, if the recycled system fees are dropped to a low level, voter will have no additional incentives (willing to vote).

PR #22, opened 2017-12-26
NEP: Stack Isolation for NeoVM

erikzhang:
This NEP proposes that stack isolation of the NeoVM evaluation stack should be carried out to ensure the security of dynamic invocations and provide support for future new features.

### DIFF
--- a/README.mediawiki
+++ b/README.mediawiki
@@ -61,7 +61,7 @@ First review [[nep-1.mediawiki|NEP-1]]. Then clone the repository and add your N
 | Stack Isolation for NeoVM
 | Erik Zhang
 | Standard
-| Accepted
+| Final
 |-
 | [https://github.com/neo-project/proposals/pull/25 9]
 | URI Scheme
--- a/README.mediawiki
+++ b/README.mediawiki
@@ -37,7 +37,7 @@ First review [[nep-1.mediawiki|NEP-1]]. Then clone the repository and add your N
 | Dynamic Contract Invocation
 | localhuman, unignorant
 | Standard
-| Final
+| [[nep-8.mediawiki|Replaced]]
 |-
 | [[nep-5.mediawiki|5]]
 | Token Standard
@@ -57,7 +57,7 @@ First review [[nep-1.mediawiki|NEP-1]]. Then clone the repository and add your N
 | Standard
 | Accepted
 |-
-| [https://github.com/neo-project/proposals/pull/22 8]
+| [[nep-8.mediawiki|8]]
 | Stack Isolation for NeoVM
 | Erik Zhang
 | Standard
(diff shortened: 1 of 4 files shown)


shargon:
`CALL` and `APPCALL` should copy all values but working on isolated way too, or this opCodes should not 
 be modified ?
Nodes shoud reject any new contract with old OpCodes on deploy? (how deal with this)
Are defined the new OpCode values?

shargon:
0xB**X** have a hole
What do you think?

```
CALL_I = 0xB0,
CALL_E = 0xB1,
CALL_ED = 0xB2,
CALL_ET = 0xB3,
CALL_EDT = 0xB4
```

erikzhang:
0xBX is OK.

shargon:
In witch branch are `DYNAMICCALL` implemented? @localhuman

localhuman:
In master.  We got rid of `DYNAMICCALL` opcode  and use `APPCALL` instead.

shargon:
Then if there are not `DYNAMICCALL` OpCode why need `CALL_ED` and `CALL_EDT`. Is not the same?

localhuman:
Does this relate to an issue of an `APPCALL` overwriting the `altstack` of the method it is being called within?

erikzhang:
@localhuman Since we don't have `DYNAMICCALL`, can you edit NEP-4 to correct it?

hal0x2328:
This change is urgently needed in order to safely do dynamic calls to untrusted contracts, and the AltStack needs to be isolated as well, as that is where contract global/local variables are stored. Very often `StorageContext` is saved in a named variable at the beginning of a contract, which could lead to any number of cross-contract security issues.

valdor27:
Is someone other than Erik even working on NEO. No one seems to pay any heed to this PR

erikzhang:
> Is someone other than Erik even working on NEO. No one seems to pay any heed to this PR.

Be patient, I'm working on it.

erikzhang:
You can check the progress of the development for NEP-8 in this branch: https://github.com/neo-project/neo-vm/tree/feature/nep-8

dicarlo2:
> On the other hand, some features, such as exception handling, also require stack isolation to be implemented.

Will exception handling be addressed in NEP-8 or another NEP?

erikzhang:
> Will exception handling be addressed in NEP-8 or another NEP?

It's not in NEP-8.

dicarlo2:
@erikzhang what's the status of this NEP? When will it be live?

erikzhang:
The implementation has been completed. But it has not been deployed to the mainnet.

dicarlo2:
@erikzhang do you have an estimate of when this will be deployed? I'm really looking forward to using this feature.

erikzhang:
If all goes well, it will be deployed to the testnet next week.

dicarlo2:
@erikzhang Great news! We've already updated neo-one to support stack isolation, and actually require it for the typescript compiler, so the sooner the better :)

ixje:
FYI; `neo-python` and `neo-boa` now also have support for NEP8

PR #15, opened 2017-10-16
NEP: Triggers for NeoContract

erikzhang:
Trigger is a mechanism for triggering execution for smart contracts. This NEP defines 4 kinds of triggers, they are <code>Verification</code>, <code>VerificationR</code>, <code>Application</code> and <code>ApplicationR</code>.

PR #25, opened 2018-01-15
NEP-7 URI Standard for Native Asset Transfer

saltyskip:


nep-7.mediawiki
```diff
@@ -0,0 +1,117 @@
...
+
+Currently, there is no standard URI for NEO clients to consume. Bitcoin implements a URI standard so that clicking links or scanning QR codes can easily implement a bitcoin payment. Similarly the transfer of native assets on the NEO chain should be just as simple.
+
+In the future this URI scheme should enable generic smart contract invocation. This would allow for transfer of NEP-5 tokens, but these extensions should be added on once the more basic URI standards are agreed upon
+
+==Specification==
+
+===Native Asset Transfer URI===
+
+A Native asset transfer would have the following URI. It describes the reciever address, assetID, and additional attributes that will be sent with the transaction. The attributes are presented in raw hex data.
```
snowypowers: Native asset or Native Asset or native asset?

receiver spelling error

nep-7.mediawiki
```diff
@@ -0,0 +1,117 @@
...
+
+<code>NEO:<address>[?assetID=<assetID>][?amount=<amount>][?attributes=<key>=<value>]</code>
```
snowypowers: I would prefer that the key/value pair be separated by another symbol. I think its good to follow querystring conventions, not sure if the decimal point in the amount will screw things up

nep-7.mediawiki
```diff
@@ -0,0 +1,117 @@
...
+
+Begin transaction of unspecified amount of NEO
+
+<code>NEO:AeNkbJdiMx49kBStQdDih7BzfDwyTNVRfb?assetID=c56f33fc6ecfcd0c225c4ab356fee59390af8560be0e930faebe74a6daff7c9b</code>
+
+Begin transaction to specified address of 1 NEO
+
+<code>NEO:AeNkbJdiMx49kBStQdDih7BzfDwyTNVRfb?assetID=c56f33fc6ecfcd0c225c4ab356fee59390af8560be0e930faebe74a6daff7c9b&amount=1.0</code>
+
+Begin transaction to specified address of 1 NEO with transaction attribute Remark = "Hello"
```
snowypowers: Can we have another example with 2 attributes? It is currently not clear if i can attach 2 or more attributes and how to do it.

One way we can do it is to just attach the entire attribute serialized string in. So a tx with 3 Hello remarks will be:
```
attributes=f00548656c6c6ff00548656c6c6ff00548656c6c6f
```

Whether we want to include the array length byte would be up to the standard.

nep-7.mediawiki
```diff
@@ -0,0 +1,117 @@
...
+|-
+| 0x02-0x03
+|	ECDH02-ECDH03
+|	Public key for ECDH key exchange
+|-
+| 0x20 
+|	Script
+|	Additional validation of transactions
+|-
+| 0x30
```
snowypowers: This would be pretty cool and useful to implement a vote for me button

nep-7.mediawiki
```diff
@@ -0,0 +1,117 @@
+<pre>
+  NEP: 7.0
```
erikzhang: The number `7` has been alloced for [NEP: Triggers for NeoContract](https://github.com/neo-project/proposals/pull/15).
After a proposal being accepted, I will alloc a number for it.

nep-7.mediawiki
apisit: Will do.
Here is the example of attaching more than one attributes.  Implemented using standard URL package in Go.

https://play.golang.org/p/USCC87orRMS

nep-7.mediawiki
apisit: I believe this follows query string convention.
the decimal point in it is fine. Please see s

erikzhang:
As a URI standard for transfer, do we really need complete support for transaction attributes? Or do we just support some of the attributes associated with the transfer?
Like this:
```
neo:AQc5mtFayAdoCK13BW1cGAzAHyo9SoUWe7?asset=602c79718b16e442de58778e148d0b1084e3b2dffd5de6b7b16cee7969282de7&amount=0.1&description=Hello
```

apisit:
Transaction attributes are optional.

I believe putting a list of supported attribute keys in the proposal can benefit developers who are reading it. It's just reflecting the NEO document right now.  

An application might choose to support only one attribute key. e.g. `0x90` for `description` 
A list of supported keys is there to help with validation when an application chooses to implement it.

I totally agree that the more human-readable URI, the better. 

I was thinking about the case which an application can let user generate a QR code and pass one's public key for ECDH in `0x02` field.

e.g.
`NEO:AQc5mtFayAdoCK13BW1cGAzAHyo9SoUWe7?assetID= 602c79718b16e442de58778e148d0b1084e3b2dffd5de6b7b16cee7969282de7&amount=0.1&attributes=0x02=02ed53ad58c838435d4dd7a4b25c1eba01384c814ca53a539405434807afbb04b4`


what if we have a map table of transaction attribute and URI key? 

e.g. 

| NEO transaction attribute | URI Key |
|-------|------|
| 0x02 | publicKey |
| 0x09 | description |
| 0xf0 | remark1 |

_note: not a complete list._

then it would be something like

`NEO:AQc5mtFayAdoCK13BW1cGAzAHyo9SoUWe7?assetID= 602c79718b16e442de58778e148d0b1084e3b2dffd5de6b7b16cee7969282de7&amount=0.1&attributes=publicKey=02ed53ad58c838435d4dd7a4b25c1eba01384c814ca53a539405434807afbb04b4`

erikzhang:
If we have a map table of transaction attribute and URI key, can we use the keys in URI directly?
e.g.
```
neo:AQc5mtFayAdoCK13BW1cGAzAHyo9SoUWe7?assetID= 602c79718b16e442de58778e148d0b1084e3b2dffd5de6b7b16cee7969282de7&amount=0.1&publicKey=02ed53ad58c838435d4dd7a4b25c1eba01384c814ca53a539405434807afbb04b4&description=Hello
```
By the way, the scheme name of URI is usually lowercase.

saltyskip:
A thorough reference implementation can be found here

https://github.com/O3Labs/NEP9-go

saltyskip:
@erikzhang What Needs to be done to push this to Final?

erikzhang:
I think we can support NEP-5 assets in this standard as well.

```
neo:<address>[?asset=<AssetID>|<ScriptHash>][&amount=<amount>][&<TransactionAttributeKey>=<value>]
```

If we want to transfer native asset, we put the AssetID to the `asset` field, which is 32-bytes length hash;

If we want to transfer NEP-5 asset, we put the ScriptHash of the contract to the `asset` field, which is 20-bytes length hash.

apisit:
I have the implementation of NEP-5 asset transfer and smart contract invocation. I'll update the proposal tomorrow to get some feedback.

saltyskip:
Won't this interfere with generic smart contract execution URI's in the future? If we do this then we'll have a specific format for a small subset of invoke transactions (NEP-5 token transfer) and another format for everything else?

Or is NEP-5 token transfer URI important enough such that it deserves it's own URI scheme?

erikzhang:
If we have the generic smart contract execution URI's, we do not need a small subset of NEP-5 token transfer.

saltyskip:
Ok we've added generic smart contract execution URI into the proposal, please review when you have time

nep-9.mediawiki
```diff
@@ -0,0 +1,161 @@
...
+<code>neo:AQc5mtFayAdoCK13BW1cGAzAHyo9SoUWe7?assetID= 602c79718b16e442de58778e148d0b1084e3b2dffd5de6b7b16cee7969282de7&amount=0.1&ecdh02=02ed53ad58c838435d4dd7a4b25c1eba01384c814ca53a539405434807afbb04b4&description=Hello</code>
+
+
+This should be sufficient to facilitate transfer of native NEO assets safely.
+---
+
+===Smart Contract Invocation URI===
+To invoke the smart contract, we need to specify a script hash and the operation of the smart contract that is being invoked as well as providing required parameters. This way an application can tailor the screen to match the invoking operation with pre-filled information. e.g. transfer NEP-5 token, voting, etc. what left for a user to do is to authorize/sign the transaction. It would have the following URI. 
+
+<code>neo:<ScriptHash>?operation=<operation>[&<parameterName>=<parameterValue>][&<TransactionAttributeKey>=<value>]</code>
```
snowypowers: This would require the `paremeterName` to not collide with `TransactionAttributeKey` (unlikely but just raising a thought)

nep-9.mediawiki
```diff
@@ -0,0 +1,161 @@
...
+
+===NEP-5 Token Transfer URI===
+
+A NEP-5 token transfer is a smart contract invocation and would have the following URI. 
+
+<code>neo:<ScriptHash>?operation=transfer&from=<address>&to=<address>&amount=<amount>[?<TransactionAttributeKey>=<value>]</code>
```
snowypowers: any leeway for optionals? For example, I could have a donate some ONT tokens to me link with the `from` address empty. I could even leave `amount` empty for the user to decide how much to donate. Might make it messy though...

nep-9.mediawiki
apisit: unlikely but yes possible, however, I think it's just up to the parser to validate it. there are 38 of them. they could be reserved words.

nep-9.mediawiki
apisit: I believe `from` and `amount` fields could be optional. it's pretty much like doing a native transfer. I'll update it.

saltyskip:
I like @snowypowers insight about this URI being generally scoped for "value transfer" I think this is a good way of thinking about it that this time. We can thus support NEP-5 and Native Assets at this time, and then gradually expand as other templates reach maturity such as non fungible tokens. 

I think it makes sense to keep URI's in general to be scoped to well established contract types only.

belane:
@saltyskip @apisit What do you think to include `fee` as optional parameter, for future usages?

```
neo:<address>[?asset=<AssetID>|<ScriptHash>][&amount=<amount>][&fee=<fee>][&<TransactionAttributeKey>=<value>]
```

saltyskip:
Fee as in a network fee? System Fee? Or a fee to be sent to the owner of the contract as a means of payment? 

In regards to network fee, I believe it should be determined entirely by the user/wallet software. There is no good way to know what would be an appropriate network fee at the time of URI generation. If it is a static URI it is completely impossible, if it is a generated URI than one could query the network status, and make some sort of rough calculation, but this seems unwieldy.

If it is a system fee, then given we are only doing NEP-5 contracts, and well established templates I'm not sure it makes sense. I expect basic templated operations like NEP-5 transfers to remain free for a long time.

The same line of thinking for a fee that is sent to the contract owner as a means of payment. As a side note I believe adopting standards for fee payments for smart contract owners will reduce the amount of Utility Coins that are produced in favor of using GAS for decentralized apps, but that is another topic

saltyskip:
I assume you are referring to network fees (used to determine transaction priority)

In this case appropriate Fees are dynamically determined based on network traffic, however URI's are generally statically generated. There is no way to accurately determine a suggested amount to send as a network fee statically as that would require knowledge of the state of the network. 

I think the client consuming the URI (which should have all available network info at the time) to be the sole determiner of an appropriate network fee

saltyskip:
I have added a proposed Supported Smart Contract URI's. @snowypowers please let me know your thoughts

nep-9.mediawiki
```diff
@@ -0,0 +1,189 @@
...
+3. Well documented and easy to consume by a client
+4. A set of examples where the URI is useful
+
+If it meets these criteria than can be considered to enter the list of supported URI schema.
+
+===NEP-5 Token Transfer===
+
+A NEP-5 token transfer is a smart contract invocation and would have the following URI. 
+
+<code>neo:<ScriptHash>?operation=transfer&from=<address>&to=<address>&amount=<amount>[?<TransactionAttributeKey>=<value>]</code>
```
snowypowers: It doesn't make sense for a URI to contain the `from` field as anyone clicking on it will want to fill in their own address.

Can we specific if these are optional (`from` and `amount` should be optional)

snowypowers: Another note, what endian should the scripthash be in? (I am asking if this should be specified)

nep-9.mediawiki
```diff
@@ -0,0 +1,189 @@
...
+|-
+| GAS 
+| 602c79718b16e442de58778e148d0b1084e3b2dffd5de6b7b16cee7969282de7 
+|}
+
+====Examples====
+
+Begin transaction to specified address
+
+<code>neo:AeNkbJdiMx49kBStQdDih7BzfDwyTNVRfb</code>
```
snowypowers: Are URIs case sensitive? It would be troubling if the case is ignored for addresses.

igormcoelho:
I wonder if the complete AssetID hash is really necessary... I mean, if it could be determined by the first initial hash values (such as 602c79), like github does for commit hashes, perhaps that would save some valuable space (for qrcodes). The ambiguity can be possibly become a problem here? just wondering...

erikzhang:
>I wonder if the complete AssetID hash is really necessary... I mean, if it could be determined by the first initial hash values (such as 602c79), like github does for commit hashes, perhaps that would save some valuable space (maybe for qrcodes). The ambiguity can possibly become a problem here?

It is very easy to make two hash values with same first few bytes.

igormcoelho:
> It is very easy to make two different hash values with same first few bytes.
But the asset must be registered in the main net, required to pay all fees... and if this happens, people could simply start using more bytes in uri. But I get your point, for security purposes it's safer to use the complete id.

erikzhang:
So what if the checksum itself be replaced?

igormcoelho:
> So what if the checksum itself be replaced?

Then we will need a checksum of the checksum :D hahah just kidding. I mean Erik, the current URI has no checksum, so data can be manipulated anyway (but I agree that's not in the scope of the NEP). I'm just saying that perhaps those middle bytes of the assetid could be replaced by something that represents all transfer information (including the whole assetid), and an extra checksum, with fewer space. But theoretically speaking, I agree with you that there may exist another checksum that is coeherent with the manipulated assetid (not the original one, missing pieces), and also coherent with the rest of the data.. but I don't imagine how to generate it, really. The checksum can be optional anyway, as everything else I'm saying :D If anyone is interested I can create some practical examples to put here, but like I said, just a thought. ;)

mwherman2000:
Is this a valid query parameter string? i.e. multiple equal signs? 
```
attributes=0x02=02ed53ad58c838435d4dd7a4b25c1eba01384c814ca53a539405434807afbb04b4
```

apisit:
@mwherman2000 You might saw the old one. It's already updated to use Transaction attribute name as a query key.

I have updated the proposal to include detail on endianness and optional fields for NEP5.

We have a simple implementation in JS of embedding a QR code generated with this NEP9 spec [here](https://github.com/O3Labs/nep9.js)

igormcoelho:
This script for generating the QR Code is really nice, congratulations!

erikzhang:
Can we handle the transfer of global assets and the transfer of NEP assets in a unified way?

apisit:
yeah that looks better @erikzhang 
I'll update it.

erikzhang:
Two suggestions:

1. Now the article structure is a bit messy, can you rearrange it?
2. For the `asset` field, can we support the alias for neo and gas?

erikzhang:
Now we need to update the implementations.

apisit:
Updated the implementations.

https://github.com/O3Labs/NEP9-go
https://github.com/O3Labs/nep9.js

erikzhang:
One last question, should we mark it as `Final` or `Active`?

nickfujita:
I've updated the sample QR code creation page to conform to the latest updates:
- update `assetID` to `asset`
- update for `neo` and `gas` aliases for `asset`
- updates for nep5 token transfer compatability

This can be used as an example for QR code creation in accordance with NEP9.

https://o3.network/nep9/
[source](https://github.com/O3Labs/o3.network/blob/master/nep9-draft/index.html)

apisit:
@erikzhang I voted it to be `Final`

saltyskip:
👍 for final

nickfujita:
+1 to finalize! 💯

igormcoelho:
Congratulations, that's nice! Seems final to me as well :) +1

PR #27, opened 2018-01-30
NEP: Improve Pricing Structure

RavenXce:
In order to solve some underlying issues with the current gas pricing structure, I've proposed adjusting the gas ratio and dropping the free gas cap, in addition to changing how `Storage.Put` cost is derived as mentioned in https://github.com/neo-project/neo/issues/159,.

A consensus mechanism for dynamic gas ratio would be nice, but the implementation for such an endeavour is unclear at the moment.

Other individual op costs can also be further simplified or tweaked based on what we now know, as compared to when NEO was launched.

### DIFF
--- a/nep-10.mediawiki
+++ b/nep-10.mediawiki
@@ -1,0 +1,80 @@
+<pre>
+  NEP: 10
+  Title: Pricing Structure
+  Author: Ivan Poon <ravenxce@gmail.com>
+  Type: Informational
+  Status: Draft
+  Created: 2018-01-30
+</pre>
+
+==Abstract==
+
+This NEP describes an improvement to the current GAS pricing scheme for invoking transactions on the NEO blockchain.
+
+==Motivation==
+
+Currently, due to how <code>Storage.Put</code> operations are [https://github.com/neo-project/neo/blob/48128f1a4a3c8a0b7c4f260145d0098d4dd5ea5d/neo/SmartContract/ApplicationEngine.cs#L395 calculated], 
+there is a minimum of 1 GAS charged regardless of how small the stored payload is.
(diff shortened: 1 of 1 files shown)


nickfujita:
Just wanted to provide some additional details on the matter. Under the current fee model, NEP5 token transfers take about 2.4~ gas per transfer (1 CheckWitness, 2 Get commands to pull balance of each account, 2 Put commands to update balances of each account; less if the from account balance becomes 0, then the total gas is 1.5~). So we can safely batch 3 transfer commands into a single transaction.

At the current limit of 500 transactions per block, we are limited to a max of 1500 NEP5 transfers per block. Depending on the block times, this leads to a max transfers per second of 75-60 (20-25 second block times).

If we were to update the fee model as suggested, retaining the `gas_free` amount, updating the min Put cost to 0.1 gas, and assuming that a simple Put of the account token balance falls within this amount. We can possibly expect a the cost per NEP5 transfer decreased to 0.6~. Which will allow batches of roughly 16 transfers per transaction. Resulting in a new max of 8000 NEP5 transfers per block, a 533% increase, and a new transfers per second of 400-320 (20-25 second block times).

This will help to reduce the load during airdrops and bulk KYC registrations. For KYC registrations, the increase would be from 200-160 to 825-660 registrations per second.

This figures are all assuming the transactions are all within the `gas_fee`, and that it is kept in place at 10 gas per transaction. However, if it is removed, these numbers become irrelevant since the the cost for making such transactions won't change whether they are batched in much higher numbers or not at all.

erikzhang:
In NEO 3.0, prices will be significantly reduced. We will give a plan soon.

erikzhang:
Pricing structure will be modified in NEO 3.0. I will close this PR.

PR #31, opened 2018-03-19
NEP: P2P protocol upgrade

shargon:
We believe is necessary a P2P protocol optimization to avoid unnecessary data exchange, reducing the computational consumption, in order to improve and speed up communication between the nodes.

### DIFF
--- a/NEP-P2P_Protocol_change.mediawiki
+++ b/NEP-P2P_Protocol_change.mediawiki
@@ -34,7 +34,7 @@ TODO
 ==Implementation==
 
 The current header includes a magic numbers of 4 bytes in all messages. 
-This proposal try to change this header to 20 bytes only during the initial negotiation of the communication.
+This proposal try to change this header to 20 bytes only during the initial negotiation of the communication avoiding message overload.
 
  +--------------------------+       +--------------------------------------------------------+
  |   Protocol negotiation   |  ...  |                    Message Protocol                    |  
@@ -45,7 +45,7 @@ This proposal try to change this header to 20 bytes only during the initial nego
 
 ===Protocol negotiation (20 bytes)===
 
-This is the first message sent, and is checked for continue or disconnect the client.
+This is the first message sent, and UUID (Protocol Magic Number) is evaluated to establish the connection or disconnect the client.
 
  +----------------------------+
  |    Protocol negotiation    |
@@ -59,6 +59,7 @@ This is the first message sent, and is checked for continue or disconnect the cl
 * UUID - 16 bytes
 
 We think that is better have a unique id (16 bytes) on the negotiation than an integer (4 bytes) on each message.
+UUID should be defined on protocol.json as NEO protocol ID.
 
 * Version Message
 
--- a/NEP-P2P_Protocol_change.mediawiki
+++ b/NEP-P2P_Protocol_change.mediawiki
@@ -57,11 +57,13 @@ This is the first message sended, and is checked for continue or disconnect the 
  +------+ +-------------------+ 
  
 * UUID - 16 bytes
- We think that is better to be unique id that 4 bytes as currently doing
- 
+
+We think that is better to be unique id that 4 bytes as currently doing
+
 * Version Message
- This message is like Message Protocol with the VersionPayload as Payload
- 
+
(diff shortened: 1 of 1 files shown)


erikzhang:
Have you tested how much performance it can improve? Does it solve the bottleneck of NEO network communication?

shargon:
Is not implemented (neither tested) yet, but i'm sure that the compression system could help, a lot of, with heavy blocks

AshRolls:
@erikzhang do you accept this NEP in principle? We can start implementing if so.

erikzhang:
I don't think this improvement will lead to significant performance improvements, so it won't be accepted in the short term.

shargon:
The main goal is add compression to the p2p protocol

erikzhang:
> The main goal is add compression to the p2p protocol

@shargon  We can add compression flag to `VersionPayload.Services`. Then only if the connected two peers apply the `compression` flag both, they can communicate with each other with compressed message.

erikzhang:
If we need to support compression in `Message`, just add flags to `Message`. And change the protocol version.

NEP-P2P_Protocol_change.mediawiki
```diff
@@ -0,0 +1,129 @@
...
+  Title: P2P Protocol change
+  Author: Shargon and Belane
+  Type: Standard
+  Status: Draft
+  Created: 2018-03-07
+</pre>
+
+==Abstract==
+
+We believe that a P2P protocol optimization is necessary in order to avoid unnecessary data exchange between nodes.
```
ixje:
```suggestion
A P2P protocol optimization is deemed necessary in order to avoid unnecessary data exchange between nodes. 
```
I think a more generic (vs personal "we believe") writing style suits NEPs better as they are quite serious documents (imo).

NEP-P2P_Protocol_change.mediawiki
```diff
@@ -0,0 +1,129 @@
...
+In this sense, reducing the computational consumption and, consequently, improving and speeding up the communication between nodes.
+
```
ixje:
```suggestion
This proposal describes various improvements.
```
In general an abstract should be standalone readable and trigger the user to want to read more/go in depth. I believe this small addition gives that last bit of information that can trigger.

NEP-P2P_Protocol_change.mediawiki
```diff
@@ -0,0 +1,129 @@
...
+==Motivation==
+
+Improve the P2P protocol to support higher loads, as well as improving the communication speed and prioritizing consensus messages. 
+In addition, we think that it is necessary to implement a compression mechanism and save all bytes that should be possible.
```
ixje:
```suggestion
```
I believe the 
> Improve the P2P protocol to support higher loads, as well as improving the communication speed and prioritizing consensus messages. 

is actually a good concise and all encompassing description. 

The "compression" line specifically highlights one optimization that leads to above description. Perhaps it fits better in the rationale section?

NEP-P2P_Protocol_change.mediawiki
```diff
@@ -0,0 +1,129 @@
...
+TODO
+
+==Test Cases==
+
+TODO
+
+==Implementation==
+
+The current header includes a magic number of 4 bytes in all messages. 
+With this current proposal, this header is sent only during the initial negotiation of the communication, avoiding message overload.
```
ixje:
```suggestion
With this proposal, this header is sent only during the initial negotiation of the communication, reducing overhead in all remaining communication messages.
```
"current proposal" suggests _this_ one will change. At best it will be amended or superseded so I think it is better to refer to it as "This proposal".

I had to read "avoiding message **overload**" a couple of times before I understood the intention. I think this suggestion is more self explanatory/clear.

NEP-P2P_Protocol_change.mediawiki
shargon: TCP has his own CRC protocol

shargon:
@ixje TCP have his own CRC and ensure the integrity of the packet

erikzhang:
Implemented in NEO 3.0: https://github.com/neo-project/neo/pull/710

PR #37, opened 2018-04-10

lightszero:
非同质化代提案，中文版


erikzhang:

erikzhang:
@lightszero I have some questions for you to answer.

1. As a Non-fungible Token Standard, why we need `totalSupply`? Since each token is not the same, there should be no need to count the number of tokens?

1. In the case of a transfer, if we know the `nftid`, we can get the owner of it. So there's no need to provide the `from` parameter for `transfer`?

1. What is the purpose of `getTXInfo`?

PR #38, opened 2018-04-11
NEP: Token Standard

lightszero:


erikzhang:
Did you forget to change the authors list?

PR #39, opened 2018-04-16
NEP: Token Sales Standard

snowypowers:
This NEP proposes a standard for implementing certain features for a token sale. With this NEP, wallets will be able to present the user with relevant information during the token purchase process, allowing for a smoother experience.

### DIFF
--- a/nep-sales.mediawiki
+++ b/nep-sales.mediawiki
@@ -58,14 +58,6 @@ This standard defines two method types:
 
 * Remarks: ''kycStatus'' Checks the whitelist status for the provided address. Returns True if registered.
 
-====kycLimit====
-
-* Syntax: <code>public static int kycLimit(byte[] address)</code>
-
-* Remarks: ''kycLimit'' returns the maximum amount of tokens that the provided address can exchange for. This method should ignore the global remaining tokens and focus on the personal limit. Sales without a personal limit should return the largest possible integer.
-
-* Example: For an address registered to exchange up to 1000 tokens, in a sale where 1 NEO is worth 100 tokens, 1 GAS is worth 10 tokens, this method should return 1000 at the beginning. After contributing 1 NEO and 10 GAS, this method should return 800. 
-
 ====saleStatus====
 
 * Syntax: <code>public static bool saleStatus()</code>
--- a/nep-sales.mediawiki
+++ b/nep-sales.mediawiki
@@ -72,12 +72,6 @@ This standard defines two method types:
 
 * Remarks: ''saleStatus'' returns the current state of the sale. Returns true if sale is active.
 
-====saleRemaining====
-
-* Syntax: <code>public static int saleRemaining()</code>
-
-* Remarks: "saleRemaining" returns the amount of tokens left for sale. This is the remaining amount of token left available for exchange. Sales without a limit should return the largest possible integer.
-
 ====saleDetails ''(optional)''====
 
 * Syntax: <code>public static string saleDetails()</code>
--- a/nep-sales.mediawiki
+++ b/nep-sales.mediawiki
@@ -1,7 +1,7 @@
 <pre>
   NEP: <to be assigned>
-  Title: Crowdsale Standard
-  Author: Thomas Saunders, Yak Jun Xiang
+  Title: Token Sales Standard
+  Author: Yak Jun Xiang (snowypowers@gmail.com)
   Type: Standard
   Status: Draft
   Created: 2018-03-16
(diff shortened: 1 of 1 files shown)


nep-sales.mediawiki
```diff
@@ -0,0 +1,123 @@
...
+
+* Remarks: ''mintToken'' attempts to create tokens for the specified address. If successful, this action should increase the totalSupply of the tokens. This method must be implemented together with the ''token_mint'' event.
+
+====kycStatus====
+
+* Syntax: <code>public static bool kycStatus(byte[] address)</code>
+
+* Remarks: ''kycStatus'' Checks the whitelist status for the provided address. Returns True if registered.
+
+====kycLimit====
```
nickfujita: While this is a useful method in some specific sales cases, it might not be appropriate as a standard method. 

- Depending on the token sale, there are different limits; with some sales having multiple limited rounds. 
- Doesn't fit if there are sales that have no limits, and returning the `largest possible integer` seems a bit hacky.
- Doesn't address sales where there are minimum contribution requirements.

Side note: if this method were to be kept, a different name may be more appropriate. `kycLimit` may seems to suggest there is a limit pertaining to the kyc process or number of kyc addresses. This is really more a part of the sales methods.

It may just be better to remove this method, and have contract owners provide users with information about round limits as a part of `saleDetails`. `saleDetails` can also be updated to take an optional address parameter to allow contracts to provide user specific details about their contributions to date.

nep-sales.mediawiki
```diff
@@ -0,0 +1,123 @@
...
+
+* Syntax: <code>public static bool saleStatus()</code>
+
+* Remarks: ''saleStatus'' returns the current state of the sale. Returns true if sale is active.
+
+====saleRemaining====
+
+* Syntax: <code>public static int saleRemaining()</code>
+
+* Remarks: "saleRemaining" returns the amount of tokens left for sale. This is the remaining amount of token left available for exchange. Sales without a limit should return the largest possible integer.
```
brianlenz: I think `saleRemaining` should either be optional or perhaps better, not included in the NEP. Token sales can be complex and many have different phases of the sale, so what this represents will be ambiguous in many cases. I think it's better to keep the interface as simple as possible and only include methods and events that add value almost universally to all token sales.

nep-sales.mediawiki
```diff
@@ -0,0 +1,123 @@
...
+
+* '''(Required)''' : methods that are compulsory in order to be NEP-compliant. Methods are required unless marked otherwise. These methods are mostly user-facing and thus must be standardised for a smooth experience.
+
+* '''(Optional)''' : methods that are optionally implemented. These methods are not all user-facing and contains certains parameters not present in all sales. However, developer tools are encouraged to follow this specification for the best experience.
+
+===Methods===
+
+====mintTokens====
+
+* Syntax: <code>public static string mintTokens(byte[] address)</code>
```
brianlenz: What is the `string` return type for? It would be cool if it was possible to return a `string` message when a mint fails, but I'm not sure that's feasible given how raw transactions are submitted, and also in the context of `TriggerType.Verification`. It's worth considering if there is a way to return a human-readable message, and if not, changing this to just return `bool`.

nep-sales.mediawiki
```diff
@@ -0,0 +1,123 @@
...
+
+====saleRemaining====
+
+* Syntax: <code>public static int saleRemaining()</code>
+
+* Remarks: "saleRemaining" returns the amount of tokens left for sale. This is the remaining amount of token left available for exchange. Sales without a limit should return the largest possible integer.
+
+====saleDetails ''(optional)''====
+
+* Syntax: <code>public static string saleDetails()</code>
```
brianlenz: It might be cool to add an optional `byte[] address` argument that could be used to generate a personalized status message for a given address. This message could indicate any per-address sale details (e.g. if different users get different token ratios or have different limits). This would give ICO implementors a bit more flexibility and power in delivering a more useful and accurate message in the wallet interface. Worth consideration, anyway.

snowypowers: Ok, by doing this, we can remove the limit / sale remaining by letting the contract owner compile their own custom string that may target the specific address if needed.

nep-sales.mediawiki
```diff
@@ -0,0 +1,123 @@
...
+
+* Syntax: <code>public static bool kycDeregister(byte[] address)</code>
+
+* Remarks: ''kycDeregister'' removes the registration of the provided address and returns True if successful. This method can only be called by the contract owner. This method should be implemented with the event ''kyc_deregister''.
+
+===Events===
+
+====mintToken====
+
+* Syntax: <code> public static event Action<byte[], int> token_mint</code>
```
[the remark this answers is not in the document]
snowypowers: I do think this is a valid point. @lllwvlvwlll @anfn101 What do you think?

nep-sales.mediawiki
snowypowers: i will remove this in favor of making `saleDetails` be the main source of information since it will be difficult in ensuring this information stays relevant for all token sales.

nep-sales.mediawiki
snowypowers: I will remove this in favor of using `saleDetails` be the main function to convey any meta information for the contract.

nep-sales.mediawiki
snowypowers: I'm not well versed in this area so I need to ask if the return value appears anywhere else other than through RPC `invoke`? I think that return values are a little wasted currently... I do like if we can return something more relevant than just a `bool`.

erikzhang:
We don't need any kyc methods if NeoID has been implemented I think.

snowypowers:
@erikzhang Even with NeoID, there will still need to be a standard in order for wallets to display the `IsAddressWhitelisted` status. The `kycRegister` / `kycDeregister` methods are optional and can be deprecated once NeoID reaches production.

I do think that in the short term we should focus this on KYC and have a review once NeoID looks to go live.

erikzhang:
Can we remove `mintTokens` because [NEP-7](https://github.com/neo-project/proposals/blob/597f0dc53a80bbedb14974c8168f8df288c1d82e/nep-7.mediawiki) will be implemented.

snowypowers:
I have removed `mintToken` from this NEP. Now, the methods left are purely focused on the crowdsale side of things.

erikzhang:
Perhaps the standard can standardize not only the methods, but also the triggers for token sale.

snowypowers:
At this stage, it is difficult for us to standardize on something that cannot be implemented and tested. Pushing it early would also cause confusion between contracts before / after NEP7. 

My suggestion would be to have an amendment (version 1.1) to include the relevant triggers after NEP-7 goes to production.

On a side note, I do think that the token creation functionality should be part of NEP-5 instead of this as suggested by Brian.

erikzhang:
Actually NEP-7 is ready to merge into neo master branch.
https://github.com/neo-project/neo/pull/172

snowypowers:
@erikzhang Yes I think I can work triggers into the standard. Off the top of my head, here's a rough draft:

> In order for a transaction to be a valid transaction for minting tokens, it will have to attach a TransactionAttribute of `Description` with the text `mintTokens`. VerificationR trigger will check the attribute and reject transactions without a purpose.

The attribute check is to make sure that the contract is still able to receive funds for other purposes beyond ICO. 

> Upon accepting the transfer, ApplicationR is invoked in the contract. It should check for the existance of TransactionAttribute of `Description` with the text `mintTokens` to ensure this transaction is used for ICO. The contract should then perform the necessary actions to mint tokens and raise the event `tokens_created` on success.

Similarly, we have to check for the purpose here because the contract may have uses for asset transfers beyond ICO. The event raised here can be fleshed out in this NEP but **it should be transferred to NEP5**.

I do believe that `Verification` trigger functionality is out of the scope of this NEP.

If this is OK, I would like to propose a separate PR to amend NEP5 to **include relevant token events such as token_created and token_destroyed**

erikzhang:
I think it seems more like a KYC standard. But not all token sales need KYC.

snowypowers:
This isn't a specification purely for KYC but it does have a lot of elements for KYC simply due to the fact that KYC has more to it.

With the merger of NEP-8, it looks like we should keep specifications to a singular purpose. This specification looks to split:

1. Sales Standard - Interface for common wallets to retrieve information for sales in order to streamline the sales process. The methods `saleStatus` and `saleDetails` are included in this spec. Requires NEP-5.

2. KYC Standard - Interface for wallets to integrate the KYC process with common wallets. The methods `kycStatus`, `kycRegister` and `kycDeregister` are included in this spec. Requires NEP-5 and the Sales Standard.

erikzhang:
Agreed.

snowypowers:
Updated:

- Title to `Token Sales Standard`
- Removed Thomas as author (he made the KYC standard but the Sales standard under me)
- Tweak motivation to suit the standard more
- Removed KYC methods
- Removed Required tag (there should not be any optional methods anymore)
- Removed all events

erikzhang:
There's a problem. In the method `saleDetails`, it returns a string that briefly describes the sale. And this string usually contains some numbers.

The problem is that there is currently no good way to convert numbers to strings in NeoVM. Some APIs need to be added.

erikzhang:
Maybe `saleDetails` should return a url?

Issue #40, opened 2018-04-18
New NEP: Composite Smart Contracts


mwherman2000:
Joe and I are looking for feedback and will post this draft shortly to #dev-general. 

https://github.com/mwherman2000/proposals/blob/master/NEP-8.mediawiki [updated]

[It was short - 1 operation/method - so it was easy to write it up to get actual feedback.]

This NEP is very timely and relevant. Looking forward to everyone's feedback.

This based on the work Joe Stewart has done with Non-Fungbile Token extensions to NEP-5 and the work I've done to support NEP-5 extensions for Requisitions.

Ejhfast:
This seems like a good idea to me. One reasonable assumption this proposal makes is that the standard will change or evolve in the future. (If the standard never changes, there would be no need for this method.) Maybe the rationale could include additional motivation for why it is likely the standard will fork or change in the future, justifying the existence of this method?

mwherman2000:
Ethan, when you are referring to "the standard", can you clarify what you mean/are referring to? ...e,.g the proposed NEP, the subject NEPs, etc?

Ejhfast:
I am referring to the NEP-5 standard, the standard that this proposal is aiming to improve.

mwherman2000:
The proposed NEP applies to all NEPs not just NEP-5.

saltyskip:
👍 For this, Can the return value be a JSON string instead? This also would allow for additional metadata to be put here in future about the contract.

e.g
```
{
	"standards": ["NEP-5", "NEP-1234"]
}
```

erikzhang:
>For this, Can the return value be a JSON string instead? This also would allow for additional metadata to be put here in future about the contract.

Good idea!

mwherman2000:
The proposed amendment is to return the following string
```
"{ \"standards\": [\"NEP-5\", \"NEP-1234\"] }"
```
instead of
```
"NEP-5", "NEP-1234"
```
Why choose a more complex string format? What are the use cases you can imagine and the additional value?  I think it's important to keep the `supportedStandards` protocol simple, practical and easy to implement.

Cons
1. We don't have JSON support in the VM or a supporting library to create *or parse* JSON strings. How would this work, for example, in the case of dynamically invoking one smart contract method from another?  
2. It goes against the pattern already established for `symbol`, `totalSupply`, `name`, etc.
3. JSON is already available as a serialization format between off-chain clients and `neo-cli`.

[...]

PR #41, opened 2018-04-19

hal0x2328:
Initial NFT proposal from Splyse/HashPuppies team


erikzhang:

saltyskip:
Have you considered a divisible non fungible token? Easier to explain with an example.

Let's see we want to put a house on the blockchain. There could be one NFT that represents this house, however due to the immense value it holds, houses are split up into pieces in some sort of real estate trust. So each person might own 10% of each house. This allows for high value objects to be tokenized more effectively. 

The whole is non fungible even if the the individual pieces of aren't. This also works really well for tokenized art. 

If an art piece is worth 50 million dollars, it is hard for own person to own as one non fungible token. However if it is tokenized into 100 fungible parts than it becomes much easier for collective ownership. 

I bring this up because NEO seems to be an excellent place to put high value (singular )non fungible items as a single contract, and then tokenizing the individual non fungible entity into smaller (fungible or non fungible) parts. This leads into the process of digitizing real world assets

hal0x2328:
@erikzhang By virtue of the fact that they together own one NFT, doesn't that mean they each technically own a part of one NFT? The multi-sig method could serve most use cases except where you want to sell your share without the consent of the other owners and cooperation of the token contract (which is what I think you are trying to illustrate with your comment). Yes, that would require a different model where parts of the token have to exist as a unique entity with a single owner. 

Perhaps a master meta-token type could be created that is just composed of a list of other tokens and the relative percentages of ownership they represent. But, now you have to consider divisibility of the sub-tokens - what if I want to sell my 10% share to two people, one of whom pays me 80% and the other and 20% of the share, giving them a resulting 8% and 2% stake in the master token? If the token share owner can't create new child tokens at will through the act of transferring them this scheme doesn't work.

Really interesting to think about, would love to add thoughts to the new proposal from @saltyskip when it is added.

erikzhang:
@hal0x2328 I have some questions for you to answer.

1. As a Non-fungible Token Standard, why we need `circulation` and `balanceOf`? Since each token is not the same, there should be no need to count the number of tokens in an account?

1. In the case of a transfer, if we know the `tokenid`, we can get the owner of it. So there's no need to provide the `from` parameter for `transfer`?

1. As a Non-fungible Token Standard, why we need `decimals`? Does this mean that each token is still divisible as @saltyskip has said? If so, then the `balanceOf` method should accept tow parameters: `owner` and `tokenid`.

1. Do we really need `RWData` and `URIData`? I guess a lot of tokens may just need `ROData`. Maybe we can simplify the standard?

erikzhang:
In addition, there is a description in the [ERC721](https://github.com/ethereum/EIPs/issues/721#issuecomment-343246872):

> NOTE: Current limitations in Solidity mean that there is no efficient way to return a complete list of an address's NFTs with a single function call. Callers should not assume this method is implemented efficiently (from a gas standpoint) and should strenuously avoid calling this method "on-chain" (i.e. from any non-constant contract function, or from any constant contract function that is likely to be called on-chain).

But in Neo, we have great ways to return to the complete list of NFTs. So we really don't need `tokenOfOwnerByIndex`.

hal0x2328:
> As a Non-fungible Token Standard, why we need circulation and balanceOf? Since each token is not the same, there should be no need to count the number of tokens in an account?

`circulation` is a synonym for `totalSupply`. This was done to maintain compatibility with the NEX ICO template in case external code relies on that operation since it exists in other NEP-5 implementations.

`balanceOf` would be used in conjunction with `tokenOfOwnerById` - we need to know how many items exist to iterate over - see the reasoning for including `tokenOfOwnerById` below and see if it makes sense.

> In the case of a transfer, if we know the tokenid, we can get the owner of it. So there's no need to provide the from parameter for transfer?

This parameter could be eliminated, true.

> As a Non-fungible Token Standard, why we need decimals? Does this mean that each token is still divisible as @saltyskip has said? If so, then the balanceOf method should accept tow parameters: owner and tokenid.

This is just to maintain compatibility with existing wallets, explorers, etc that expect that method to exist as it does with NEP-5, and might even be a way they use to differentiate (at the time, the `supportedStandards` proposal was not yet in-the-works). However, since it always returns zero, eliminating would effectively have the same result as long as the NFT implementation returned a False value at the end of execution where no operations match. I've seen some developers returning -1 from contract errors, so keeping this function and forcing Decimals to always return 0 seemed reasonable.

> Do we really need RWData and URIData? I guess a lot of tokens may just need ROData. Maybe we can simplify the standard?

We were trying to envision the most common use cases and develop a uniform way to satisfy all of them. In our game, non-fungible assets will not only have "genetics" but will also have transient properties that should be attached to the asset and carry over when the token is transferred. We figure once gaming or other NFT assets become more complex this would be a common requirement, and querying these properties should be done in a uniform way. And URIData would also likely be a pretty globally-needed feature too, we expect that NFT assets are going to need some sort of representation of their uniqueness in image form for human eyes.

> But in Neo, we have great ways to return the complete list of NFTs. So we really don't need tokenOfOwnerByIndex.

It seemed to us you still would want a way for your off-chain code to request one token at a time from a subset in a predictable order from the contract instead of retrieving the entire list every time (which eventually might grow quite large), and maintaining the Ethereum method for doing so might make it easier for NFT-using dApps to be ported to Neo.

erikzhang:
> circulation is a synonym for totalSupply. This was done to maintain compatibility with the NEX ICO template in case external code relies on that operation since it exists in other NEP-5 implementations.

We are not going to make NFT standard be compatible with NEP-5. Also, adding methods for compatibility with the NEX ICO template is not the right choice, and we should make this standard more focused on the needs of NFT itself. NEX can modify their templates for NFT standard.

> However, since it always returns zero, eliminating would effectively have the same result as long as the NFT implementation returned a False value at the end of execution where no operations match. I've seen some developers returning -1 from contract errors, so keeping this function and forcing Decimals to always return 0 seemed reasonable.

If an error occurs in the contract, the exception should be thrown by the `THROW` opcode directly. If `decimals` always returns zero, then it is really not necessary to exist.

> And URIData would also likely be a pretty globally-needed feature too, we expect that NFT assets are going to need some sort of representation of their uniqueness in image form for human eyes.

Agree with that. But I don't think RWData would be a common requirement. And perhaps RWData is more suitable for storage offchain.

> It seemed to us you still would want a way for your off-chain code to request one token at a time from a subset in a predictable order from the contract instead of retrieving the entire list every time (which eventually might grow quite large)

Perhaps we can use the Iterator APIs to return the token iterator instead of the token list? And, if it's off-chain code, we don't have to care about the size of the data.

> and maintaining the Ethereum method for doing so might make it easier for NFT-using Ethereum dApps to be ported to Neo.

I don't think we can port Ethereum dApps easily since there are many difference between NeoVM and EVM.

erikzhang:
I've been thinking about @saltyskip proposed divisible non-fungible token. Here are some of my simple ideas:

## Divisibility Extension For Non-fungible Token

### Methods

1. decimals()

### Methods (if decimals() > 0)

1. balanceOf(owner, tokenid)
1. transferPartially(tokenid, to, amount)

### Events (if decimals() > 0)

1. transfer(tokenid, to, amount)

erikzhang:
@hal0x2328 @saltyskip @lightszero @RunLyo 
I amended this standard to include optional divisibility and remove some non-essential methods.
Please review it.

saltyskip:
Aren't we moving away from optional methods due to recent proposal of composite smart contracts? If that is the case then we should move the divisibility into a separate proposal in my opinion. 

I'm a fan of this because it allows us to establish the simplest possible base layer for the non fungible tokens, and then move into more complex ideas via the composite smart contract

erikzhang:
> Aren't we moving away from optional methods due to recent proposal of composite smart contracts? If that is the case then we should move the divisibility into a separate proposal in my opinion.

I think this can be an exception because we can easily check `decimals() > 0`to know whether those divisibility methods are available.

saltyskip:
It seems like we must pick one or the other. 

Either 
A) Proposals never have optional methods, and instead each one is represented as a composite smart contract. 

For example NEP-5 would only consist of required methods. IF it had optional methods then supported methods would like 
```
[5, 5.1]
```

This would show that the optional methods are indeed present in the contract

B) Contracts have optional methods. However in my opinion this defeats the purpose of having optional methods in the first place. 

@mwherman2000 maybe you have some thoughts on this matter

erikzhang:
I think contracts can have optional methods, so long as we can detect their existence in a certain way.

mwherman2000:
@erikzhang I thought we just agreed that NEPs  will no longer have optional operations/methods?

Discussion: https://github.com/neo-project/proposals/issues/40#issuecomment-388562807

Your agreement: https://github.com/neo-project/proposals/issues/40#issuecomment-388699187

There shouldn't be any "backdoors" such as testing for the number of decimals.

Where do we stand on this?

mwherman2000:
RE: 5.1

I don't think we should start using decimals in the NEP numbering scheme. I've started a specific discussion here: <https://github.com/neo-project/proposals/issues/50>

erikzhang:
> I thought we just agreed that NEPs will no longer have optional operations/methods?

Any result requires a reason. What are the reasons why we do not allow optional methods? It is because it's difficult for us to know whether these methods exist by NEP numbers. For NFT, we have a very simple way to determine. In this case, splitting it into two NEPs is a very inefficient thing and creates unnecessary troubles in the future when it is used.

mwherman2000:
RE: "It is because it's difficult for us to know whether these methods exist by NEP numbers."

@erikzhang What is difficult about it?

jhwinter:
I'm new to the NEO community, so I'm probably pretty ignorant on a good bit of what you guys are talking about. However, reading through these comments and the posted links, I'd say that composite smart contracts make more sense than optional operations/methods. Having special use cases or exceptions where we allow optional methods would defeat the purpose of having standards as well as make it less obvious how to interact with smart contracts. 
If I query a smart contract and it tells me it supports standards `[10, 5, 1234]`, then I have some base knowledge of what data I can ask from it. In my eyes, that makes things more consistent and robust. 
I like @saltyskip's ideas for divisible NFT's. However, I agree with his earlier comment that divisible NFT should be in its own separate proposal and that this proposal should focus on establishing the base NFT implementation.

saltyskip:
How can we move this proposal forward? What outstanding questions do we still need to answer?

erikzhang:
The problem now is that we don't know if this standard works. More projects are needed to use this standard to verify that it is perfect.

saltyskip:
Hello, I've started working on some implementations regarding NFT's. First off great work! 👏 👏 One question I have is about a tokens read only properties. I wonder if the admin should be able to modify these like they are able to modify the URI.

I can see the potential use case where a token "levels up" for instance you initially have a silver membership and then an admin levels you up to a gold membership upon some task completion. 

If we are concerned about immutability, maybe there needs to be a readonly properties section and a write properties section so that the admin has modification access to the write portion of the NFT

erikzhang:
I think that each application has different business logic, they can define whether they need read-only data, and how to read and write such data.

saltyskip:
So is it preferred that all token information is located at the URI instead of inside of some storage area in the contract?

hal0x2328:
> So is it preferred that all token information is located at the URI instead of inside of some storage area in the contract?

In our contract we are using the `URI` field for an image preview link, and that URI location can be updated by the contract owner (for instance if we need to migrate the image storage to another provider). 

We're using the read-only `properties` field for storing all the read-only data associated with the token (in our case the DNA of a virtual puppy) which should never change over the life of the token. 

We are considering adding a read-write field that is custom to our application to have more flexibility in terms of things like "leveling up". This didn't make it into the final NEP per Erik's feedback at the time but there's no reason dApps can't add it as needed, it's not likely their requirements for a storage format would be compatible with each other anyway.

saltyskip:
It would be very cumbersome for wallet to be able to access NFT if there is no agreed upon place where read/write properties are stored as a part of this standard. Additionally it would be good to have some basic format for the listed properties. 

Is it a json object? 

I like @hal0x2328 approach of having a storage area in the contract for read properties whereas the URI points elsewhere. Can we amend the proposal to have the read and write storage areas for the NFT attributes?

hal0x2328:
> It would be very cumbersome for wallet to be able to access NFT if there is no agreed upon place where read/write properties are stored as a part of this standard.

I don't imagine wallets would need read or display this information, given the wide variety of use cases for NFT. This should be a storage area for for the dApp's code to read and write, and also for people who want to build specialized explorer add-ons because they can bring value to the dApp's users. 

A wallet shouldn't want to display every parameter of a HashPuppy's state. A list of token IDs with an associated preview image should be all they are worried about including, otherwise it doesn't scale for the wallet developers once there are dozens of different NFTs in the ecosystem all representing different types of things.

saltyskip:
I don't expect it to be a fully fleshed out description, but more like some basic metadata details

Here is the metadata structure for erc-721
```
{
    "title": "Asset Metadata",
    "type": "object",
    "properties": {
        "name": {
            "type": "string",
            "description": "Identifies the asset to which this NFT represents"
        },
        "description": {
            "type": "string",
            "description": "Describes the asset to which this NFT represents"
        },
        "image": {
            "type": "string",
            "description": "A URI pointing to a resource with mime type image/* representing the asset to which this NFT represents. Consider making any images at a width between 320 and 1080 pixels and aspect ratio between 1.91:1 and 4:5 inclusive."
        }
    }
}
```

Here is the metadata for OEP-4

https://github.com/ontio/OEPs/blob/master/OEPS/OEP-5.mediawiki#tokenmetadata

I think it would be good to have some meta data construct for the related for each specific NFT. These properties could be optional or not

hal0x2328:
Yeah, I like the metadata idea to describe what information is available. Another issue to tackle is - some people are storing/returning data in the serialized byte format used by Neo internally, and some are using JSON strings. So the metadata needs to know something about how the data will be encapsulated so the parser can accommodate it.

Personally I don't like the idea of the smart contract having to deal with JSON formatting when there are no built-in interop methods to serialize/deserialize objects to and from JSON, so I'm returning serialized byte arrays.

saltyskip:
I think Serialized Byte Array will be ok, but....there is currently no way to perform the serialization/deserialziation process in something like Neon-js.

I've filed the issue here

https://github.com/CityOfZion/neon-js/issues/393

Without this it not feasible for average client to access smart contract methods easily

saltyskip:
@erikzhang Can you confirm that preferred method of storage is serialized NEO dictionaries and arrays as opposed to JSON strings?

I think it is important distinction for developers who are consuming data from smart contracts

erikzhang:
Which method will return the metadata?

saltyskip:
In my ideal case it would be a new method, but perhaps @hal0x2328 has some thoughts

lightszero:
write metadata description is another job,i think you can write  a new standard for that,do not need to change smartcontract.

hal0x2328:
Yes, I'd think a new method would be best, something like `tokenMetadata`. And make it generic enough so that it works for NEP-5, NEP-11 or newer token types that get introduced after the standard is accepted.

nep-11.mediawiki
```diff
@@ -0,0 +1,176 @@
...
+  Created: 2018-04-18
+</pre>
+
+==Abstract==
+
+This NEP defines a standard non-fungible token system for the NEO Smart Economy.
+
+==Motivation==
+
+NFTs are required to track, exchange and enforce ownership of digital assets. A non-fungible token (NFT) can be thought of like a property deed - each one is unique and carries some non-mutable information (e.g. the physical address of the property). Other information, such as the owner of the property, can be changed. Also, we provide a built-in optional divisibility within each non-fungible asset. This allows for high value objects to be tokenized more effectively.
```
vncoelho: This optional divisibility were a nice inclusion. Good insight.

nep-11.mediawiki
```diff
@@ -0,0 +1,176 @@
...
+</pre>
+
+Returns a short string symbol of the token managed in this contract. e.g. <code>"MNFT"</code>. This symbol SHOULD be short (3-8 characters is recommended), with no whitespace characters or new-lines and SHOULD be limited to the uppercase latin alphabet (i.e. the 26 letters used in English). 
+
+This method MUST always return the same value every time it is invoked. 
+
+====totalSupply====
+
+<pre>
+public static BigInteger totalSupply()
```
vncoelho: For example, if we are talking about all cars of a state in Brazil, then, `totalSupply` would return that amount that can increase or decrease based on an entity that is registering that assets, right?

nep-11.mediawiki
```diff
@@ -0,0 +1,176 @@
...
+</pre>
+
+Returns the total token supply deployed in the system.
+
+====decimals====
+
+<pre>
+public static byte decimals()
```
vncoelho: In case that one car is divisible and another one is not, how would we track each corresponding part?

nep-11.mediawiki
```diff
@@ -0,0 +1,176 @@
...
+If the token managed in this contract is indivisible, the function SHOULD return <code>0</code>.
+
+If the function returns a non-zero value, all of the ''OPTIONAL'' methods in this standard MUST be implemented.
+
+This method MUST always return the same value every time it is invoked. 
+
+====tokens====
+
+<pre>
+public static enumerator tokens()
```
vncoelho: In the case I am mentioning totalSupply and tokens would walk together? Or totalSupply would be a maximum physical limit of number of registered vehicles?

nep-11.mediawiki
```diff
@@ -0,0 +1,176 @@
...
+
+If the transfer is not processed, the function SHOULD return <code>false</code>.
+
+====ownerOf====
+
+<pre>
+public static enumerator ownerOf(byte[] tokenid)
+</pre>
+
+Returns an <code>enumerator</code> that contains all the co-owners that own the specified token.
```
vncoelho: Sounds precise.

saltyskip:
If it is reccomended that tokenMetaData be part of a seperate proposal can we at least adjust properties so that it has some basic reccommended structure?

eg.
```
Properties {
   //Reccommended Properties
   name: String?
   description: String?
   image: String?
   //Contract Specific Properties
   property_one: String?
   property_two: String?
}
```

I think URI should store something other than a simple imageurl, this can instead be encoded into the properties. The uri should point to a fuller representation of the object which may be contract specific. It could be a json object that is much bigger than the one stored in the contract which could be http, ipfs, or neofs in the future. 

My main goal is to reduce some of the ambiguity around Properties and URI. Right now the expected data in both the properties and URI is very ambiguous. If we have some stricter rules around properties, then the URI can be more ambiguous

nep-11.mediawiki
erikzhang: A car is divisible, not that it really cuts the car. Dividable cars mean they are shared by many people. So why one car is divisible and another one is not?

nep-11.mediawiki
erikzhang: `totalSupply` is not the maximum limit.

erikzhang:
I have two different suggestions:

1. We keep the `tokenURI()` method and require the URI to return a json file containing the metadata.
2. Remove the `tokenURI` method and add the `tokenMetaData()` method. Fields such as `imageUrl` are included in the returned data structure.

saltyskip:
I need some time to create some application to try and consume NFT contract(eg NFT explorer) as it is written to think about best course of action, I think we will have better idea of how it will function with this infrastructure in place, and then we can tweak methods as needed. Will post something once I have it up

nep-11.mediawiki
vncoelho: Let's change example.

That NFT can represent the right that the owner has with someone.
This right can be divisible.

But, the same NFT can have both cases depending on the demand and offer of the market.

vncoelho: I mean, shares of a NFT.

ProDog:
With the development of blockchain games, many game assets need to be published and traded. I think the NFT standard needs to be determined as soon as possible.

saltyskip:
@ProDog of course it is good that we move fast on a proposal, but I think there needs to be time for developers to build and interact with NFT contracts before something can be viewed as final.

The best course of action is to let developers to build and deploy them on testnet and see what shortcomings if any there are in the proposal, then we can more readily move towards a finalized standard

saltyskip:
Ive completed a V1 of an NFT explorer + Minting contract here

https://nft.o3.app/

It includes the testenet hashpuppy contract (e7b2046b2412c4c7f1531ce144a73d47c3b272fe)

as well as my own contract based on @hal0x2328's excellent work on https://github.com/Splyse/neo-nft-template/blob/master/nft_template.py

ProDog:
Here is my contract：https://github.com/ZoroChain/Zoro-Contracts/blob/master/NFT/NFTContract/NFTContract.cs，It was developed based on what you discussed earlier

saltyskip:
Is it deployed on testnet, what is the script hash? You should be able to experiment with it on my above explorer. The issue is that my testnet nodes are down

steven1227:
@saltyskip Hi andrei, it is great effort that you have tried! I am wondering how is the outcome of your NFT contract ? especially for the Uri method as I see there are some options that can be done for this method.

saltyskip:
@steven1227 as mentioned previously I have deployed the contracts and explorer which are viewable here 

https://nft.o3.app/

Let me know if you have any issues navigating the website

erikzhang:
@saltyskip Where is the source code? What data does `tokenMetaData()` return?

saltyskip:
Deployed Version (omitted admins)

https://gist.github.com/saltyskip/00951f3964c7b511bed685b8e8d54c9f

I have not implemented the metadata field at this time and simply stored image URI in URI field.

I've also open sourced the NFT explorer code here

https://github.com/O3Labs/nft-explorer 

Before a formal specification it would be good if someone other than myself attempted to create/mint an NFT contract that is compatible with this explorer

erikzhang:
@hal0x2328 @saltyskip I amended the `tokenURI()` method:

https://github.com/neo-project/proposals/blob/8dfa5e76c559844c15e555303fcbccd3f87fd8d3/nep-11.mediawiki#tokenuri

Please check it.

erikzhang:
@saltyskip I find that, in your implementation, it doesn't implement the methods `tokens()` and `tokenURI()`. Also, it doesn't implement the method `tokensOfOwner()` correctly.

See the following links:

https://github.com/neo-project/proposals/blob/8dfa5e76c559844c15e555303fcbccd3f87fd8d3/nep-11.mediawiki#tokens
https://github.com/neo-project/proposals/blob/8dfa5e76c559844c15e555303fcbccd3f87fd8d3/nep-11.mediawiki#tokenURI
https://github.com/neo-project/proposals/blob/8dfa5e76c559844c15e555303fcbccd3f87fd8d3/nep-11.mediawiki#tokensOfOwner

kindex:
I offer to use tokenURI in two variants: 
- it can return URL to http/https JSON file
- it can return JSON directly as described in https://en.wikipedia.org/wiki/Data_URI_scheme . For example, data:text/json:,{image:'https://blockchaincuties.com/rest/svga/17/1.svg'}. This format is compatible with URI.

erikzhang:
Currently it's hard for a contract to return a JSON object. However, we can be compatible with the Data URI standard, and there may be JSON support in smart contracts in the future.

steven1227:
> * it can return URL to http/https JSON file

I agree with using the first one :it can return URL to http/https JSON file

WyattMufson:
What is the purpose of the `tokens` method? Wouldn't implementing it mean that once you are past 1000 NFTs minted in total you'll start paying more significant gas fees to mint additional NFTs? At best it should be optional.

I don't think that implementing `decimals` adds any benefit, it should be removed.

I'm not sure I understand why `balanceOf` takes two arguments. 

`tokensOfOwner` should be renamed to `tokensOf` to match the naming scheme with `ownerOf` and `balanceOf`.

I would also suggest adding `approve` and `allowance`.

For reference I think the updates to the OEP5 standard are better: https://github.com/WyattMufson/OEPs/blob/oep5/OEPS/OEP-5.mediawiki

steven1227:
> Have you considered a divisible non fungible token? Easier to explain with an example.
> 
> Let's see we want to put a house on the blockchain. There could be one NFT that represents this house, however due to the immense value it holds, houses are split up into pieces in some sort of real estate trust. So each person might own 10% of each house. This allows for high value objects to be tokenized more effectively.
> 
> The whole is non fungible even if the the individual pieces of aren't. This also works really well for tokenized art.
> 
> If an art piece is worth 50 million dollars, it is hard for own person to own as one non fungible token. However if it is tokenized into 100 fungible parts than it becomes much easier for collective ownership.
> 
> I bring this up because NEO seems to be an excellent place to put high value (singular )non fungible items as a single contract, and then tokenizing the individual non fungible entity into smaller (fungible or non fungible) parts. This leads into the process of digitizing real world assets

@WyattMufson I think the current NFT standard is following this idea.

WyattMufson:
@steven1227 I'm not aware of any NFTs deployed that follow this standard that use the decimals field. If it is going to just return 0 like in [Andrei's example](https://gist.github.com/saltyskip/00951f3964c7b511bed685b8e8d54c9f), it should be omitted.

An NFT with fractional ownership is going to behave differently than one without. At the very least NEO requires something equivalent of [ERC721](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-721.md)

hal0x2328:
I don't think removing the `decimals` operation is a good idea, even if we're not talking about fractional NFTs. In my opinion, a contract should be explicit and the end VM state should provide the invoker with assurance that the operation was performed as intended. 

 Since `zero` and `false` are the same thing on Neo (an empty byte array), I don't think it's a good idea to return `false` on an unknown operation because it becomes even more of an ambiguous result. Rather, the contract should throw an exception to inform the invoker that they have an error in their code, they are trying to invoke a non-existent function.

Any token will be queried by a number of dApps and wallets for its decimals value. So if I didn't include an explicit `decimals` operation in my contract, wallets or dApps that tried to read information about the token's properties through an invocation would end up in a `FAULT` state which will likely lead to unexpected results.

(I am aware that there are no guidelines nor best practices for how a non-existent operation invocation should return, i.e. `false` vs `FAULT` so perhaps that is something else that should be a NEP or at least well-defined and consistent throughout the documentation.)

WyattMufson:
@hal0x2328 I agree that unknown operations should `FAULT` vs `false`. I agree that we should have an official NEP - "the end VM state should provide the invoker with assurance that the operation was performed as intended".

I guess this comes down to if wallets will be querying all assets the same way. For example my [NEO/ONT iOS SDK](https://github.com/Ryucoin/neovm-utils) has different asset interfaces, so that developers don't have to worry about calling a non-existent methods (they can still call custom method if they know it exists).

I'm still not sure about if `tokens` should remain, that will put a cap on how many total NFTs can be minted (depending on the gas fees and other limits set by the VM). `tokensOfOwner` should be renamed to `tokensOf`. I also think we should add `approve` and `allowance`.

hal0x2328:
I vote for leaving the `tokensOfOwner` name as-is since it is already a familiar part of NFT standards on other blockchains.

`approve` and `allowance` would only be useful if there is also a `transferFrom`operation but all of those operations were rejected as part of the NEP-5 standard, so really not sure if we need to re-hash that discussion again. Not to mention there are already two incompatible `transferFrom` methods in contracts on the Neo MainNet.

WyattMufson:
Besides Ethereum, which doesn't seem to really implement `tokensOfOwner` which other blockchains use that? And why is that a consideration?

`approve` and `allowance` don't require a `transferFrom` method, just like how `transfer` doesn't require an argument for `from`. You just call `transfer(toAcct, tokenId)` and it checks the caller is either the owner or approved address.

If the NEP5 standard rejecting those methods means that all future standards won't have them then sure, we can leave it off. This all seems like taking NEP5 and trying to tweak it for NFTs instead of trying to make a standard for NFTs.

hal0x2328:
> Besides Ethereum, which doesn't seem to really implement `tokensOfOwner` which other blockchains use that? And why is that a consideration?

It's only a consideration because it's easier to attract developers from other blockchains if they see familiar APIs. As far as which chains, I was talking about any EVM-based chains but it does look like it's been removed from the NFT standard on Ethereum at some point which means the others will likely follow. So I guess it doesn't really matter anyway, `tokensOf` is fine.


> `approve` and `allowance` don't require a `transferFrom` method, just like how `transfer` doesn't require an argument for `from`. You just call `transfer(toAcct, tokenId)` and it checks the caller is either the owner or approved address.
> 
> If the NEP5 standard rejecting those methods means that all future standards won't have them then sure, we can leave it off. This all seems like taking NEP5 and trying to tweak it for NFTs instead of trying to make a standard for NFTs.

Yes, I think you're right, we've been building the standard according to what we *think* is going to be familiar and easiest to implement for DEXs who implemented the NEP-5 standards already. That's why I suggested that `transferFrom` should be implemented, since that's what they are going to look for by default. But yes, it is superfluous in an NFT like the `t_from` argument. Maybe we should just write the standard as simple as possible and force the DEXs to go along even if they have to rethink how their code interacts with the contract.

So, moving forward with the rest of your proposed additions - how do you see `allowance` functioning in the context of an NFT? Should it return the total number of tokens approved for transfer, given a spender? Or should it return the approved spender, given a token_id? Or, possibly an enumerator for all tokens approved for a particular owner and spender pair? 

The name `allowance` itself kind of loses its meaning when in the context of NFTs, but it's another holdover from NEP-5/other fungible token standards - should it also be changed?

WyattMufson:
> Yes, I think you're right, we've been building the standard according to what we think is going to be familiar and easiest to implement for DEXs who implemented the NEP-5 standards already.

My reasoning for participating with NEP11 and OEP5 is specifically for interoperability for a cross-chain DEX.

> Maybe we should just write the standard as simple as possible and force the DEXs to go along even if they have to rethink how their code interacts with the contract.

I agree with this reasoning.

>  how do you see `allowance` functioning in the context of an NFT?

So ERC721 doesn't have `allowance`, but does have `getApproved` - which is what `allowance` does in OEP5 (in the proposed enhancements). It just returns the approved spender for the given tokenId.

WyattMufson:
Also what are your thoughts on the `tokens` method? Wouldn't the gas price to mint new tokens increase for each new token minted? Keeping track seems like it could get expensive.

hal0x2328:
In my implementation token IDs are sequential numbers, so there's actually no need for `tokens` nor an issue with increasing GAS costs on mint, to get the next ID I just use the current total supply number and then increment and store the new total.

WyattMufson:
Yeah that’s what we do too

WyattMufson:
@hal0x2328 so then to recap:

- Remove `tokens`
- Rename `tokensOfOwner` to `tokensOf`
- Leave `decimals` (if not implemented return 0)
- Add `approve` (takes address and tokenId)
- Add `allowance` or `getApproved` (takes tokenId, return address)
- `transfer` can be called by the owner or by the approved address
- Maybe add `clearApproved`

> (I am aware that there are no guidelines nor best practices for how a non-existent operation invocation should return, i.e. `false` vs `FAULT` so perhaps that is something else that should be a NEP or at least well-defined and consistent throughout the documentation.)

I also think it'd make sense to open a separate NEP for this. All the current tutorials on NEO `return false` while for Ontology it's `raise Exception`.

hal0x2328:
Did we ever establish why `balanceOf` was changed to take a `tokenid` argument? I don't understand the reason for that either.

WyattMufson:
I think it was based off the idea that decimals could be non-zero. If we decide that this NFT standard is for non-divisible NFTs, then `balanceOf` should just take an address and `decimals` should be removed. Which should probably be the case since `transfer` does not take an amount.

WyattMufson:
I've updated the proposed amendment to OEP5 to use `getApproved` instead of `allowance`.

WyattMufson:
Also what was the reasoning for eliminating optional methods? For example `tokenURI` is optional on ERC721 and OEP5 - and we aren't using it for the Ryu NFTs.

hal0x2328:
More clarification is needed around the use of enumerators - specifically what data is returned when it says "an enumerator that returns all tokens". Is it intended to return token IDs only? Or the ID plus the actual token properties? (I know `properties` isn't a part of the standard but an NFT without some sort of unique read-only data associated with it is pretty useless, and storing that information off-chain is not trustless).

I assume the intent was for the results of `Storage.Find` to be passed as the return value from the contract. This is going to be a set of key/value pairs that can be iterated over. So in my contract, I have a set of serialized properties that represents a token, e.g.
```key -> value
token/01 -> {owner: scripthash, approved: scripthash, properties: encoded_data, uri: https://...}
``` 
(I use prefixes on all storage values as a general practice to avoid storage injection vulnerabilities and make sure that the scope of `Storage.Find` is always correctly limited).

So when I do `Storage.Find(ctx, 'token/')` and return that enumerator from `tokens`, the enumerator is returning each literal storage key and serialized value. But, what if another contract stores token IDs using another prefix? dApps will need to know the prefix/ID format for each contract in order to make sense of them.

But it gets even more problematic when we talk about the `tokensOf` operation - I can't easily enumerate the `token/` keys by owner since that isn't stored there. Instead I have another set of keys under the prefix `owners/` where there is a mapping from owner scripthash to token ID, but not the actual token data (since it doesn't make sense to duplicate all the fields into another serialized object). So the enumerator here would be returning just token ID as the value, but the standard still just refers to tokens owned by a user, so it's not consistent with the output of the `tokens` operation.

WyattMufson:
I think returning an array of just tokenIds is fine. And if you need to lookup its properties on-chain, you'd do that with the tokenId. There just needs to be a way for a user to know which tokens they have.

WyattMufson:
So here are the changes that @hal0x2328 and I have discussed here and offline:

- Remove `tokens`
- Rename `tokensOfOwner` to `tokensOf`
- `tokensOf` returns a serialized NVM array of tokenIds
- Leave `decimals` (if not implemented return 0)
- Add `approve` (takes address and tokenId)
- Add `getApproved` (takes tokenId, return address)
- `transfer` can be called by the owner or by the approved address
- Add `clearApproved` to remove the approved address
- Add `properties` that returns a serialized NVM object with the NFT's properties (such as name, etc)

- Optionally we could clarify and make this a non-divisible NFT standard and make it in-line with ERC721 and OEP5. That would involve removing `decimals` and the parameter tokenId from `balanceOf`.

Let's get this moving since this standard is used by literally no one and there are no plans for any adoption as is.

@erikzhang @saltyskip @steven1227 @mwherman2000

WyattMufson:
Also interested in what @EdgeDLT has to say about this.

erikzhang:
> tokensOf returns a serialized NVM array of tokenIds

Why not return an enumerator? The whole array could be very large.

> Add approve (takes address and tokenId)
> Add getApproved (takes tokenId, return address)

Can you describe their application scenarios?

> Add properties that returns a serialized NVM object with the NFT's properties (such as name, etc)

Currently we have `name()` and `tokenURI()`. What's the difference them and `properties()`?

hal0x2328:
> > Add approve (takes address and tokenId)
> > Add getApproved (takes tokenId, return address)
> 
> Can you describe their application scenarios?

These operation would allow external contracts to conduct a transfer at a later time on behalf of the user, such as a DEX or in our case an auction contract. 

> > Add properties that returns a serialized NVM object with the NFT's properties (such as name, etc)
> 
> Currently we have `name()` and `tokenURI()`. What's the difference them and `properties()`?

`name` only returns the name of the NFT token contract itself, `properties` could return the assigned name of the individual token or any other unique static data, such as the physical address of property in the real world, or a DNA string, etc. A NFT that just represents a number index with no other immutable properties defined in the contract is not very useful except in very limited scenarios.

WyattMufson:
> Why not return an enumerator? The whole array could be very large.

What are the downsides? What are the limitations in NVM with returning arrays? Is there a limit to how big an array can be? It could return the array paginated. There just needs to be a way for users to see all of their tokens - if there isn't, then it would be as if they don't have the tokens.

hal0x2328:
> > Why not return an enumerator? The whole array could be very large.
> 
> What are the downsides? What are the limitations in NVM with returning arrays? Is there a limit to how big an array can be? It could return the array paginated. There just needs to be a way for users to see all of their tokens - if there isn't, then it would be as if they don't have the tokens.

The max array size is 1024. 

Returning an enumerator gives the invoker the flexibility to choose how to return the data from the enumerator depending on how they format their invocation script, instead of hard-coding pagination into the contract.

Its a good idea and nice feature and simplifies the contract code quite a bit, but it's not well-supported by SDKs yet (For example I've added an issue to neon-js to add support for this: https://github.com/CityOfZion/neon-js/issues/437 ) so you have to write your own invocation script to handle iterating over the enumerator object.

WyattMufson:
@hal0x2328 so then is the DynamicList structure a good idea? That bypasses the 1024 item limit and can have it return up to 7,333,888 items (1024 x (1024 x 7 - 6))

WyattMufson:
> Returning an enumerator gives the invoker the flexibility to choose how to return the data from the enumerator depending on how they format their invocation script, instead of hard-coding pagination into the contract.

I'm not familiar with how the enumerator works

hal0x2328:
> I'm not familiar with how the enumerator works

It's just returning a VM-internal enumerator object (e.g. the result of `Storage.Find` to the invoker so that they can call `next` repeatedly and filter which results they want instead of getting all of the data at once. See the issue I posted to neon-js, it explains it more fully with a Python example.

You could use some serialized object like your DynamicList structure to return a larger array or just return the direct result of `Runtime.Serialize`, but you either have to write your own pagination routine into the contract which is rigid and complicates the contract code, or you end up returning the entire set of results to the invoker, which is not very efficient.

Putting the iterator outside of the contract code lets the invoker decide how to ask for and get just the data they want from the entire set - much more efficient and far simpler on the contract side.

steven1227:
@saltyskip As I saw you are early member who give the idea about _divisible non fungible token_, can you also join current discussion for the improvement of this NEP ?  Maybe we can combine those two ideas together.

erikzhang:
> These operations would allow external contracts to conduct a transfer at a later time on behalf of the user, such as a DEX or in our case an auction contract.

If my understanding is correct, for a DEX, deposit through `approve()` and trade via `transferFrom()`.

But we can `transfer` the tokens to the DEX contract to deposit. What's the difference?

> properties could return the assigned name of the individual token or any other unique static data, such as the physical address of property in the real world, or a DNA string, etc.

So, can't it be implemented by `tokenURI()`?

WyattMufson:
> If my understanding is correct, for a DEX, deposit through `approve()` and trade via `transferFrom()`.
> But we can `transfer` the tokens to the DEX contract to deposit. What's the difference?

You don't need a `transferFrom` function, `transfer` will work for both the owner and the approved address. 

The difference is that the user can maintain control of their asset, and should they change their mind on the approval, they can call `clearApproved`.

> So, can't it be implemented by `tokenURI()`?

`properties` would store data on-chain while `tokenURI` points to some data/JSON that is stored off-chain.

saltyskip:
@steven1227 

> Leave decimals (if not implemented return 0)

I think this would be fine solution for divisible NFT. I also like Wyatt's suggestion that we have properties for on chain data, and URI for off chain blockchain data. This is a good solution for separating readonly and readwrite properties.

@WyattMufson should a permission system for token transfer of third party be part of another standard? ie (clearapproved, getapproved)

This seems like it would be an extension that could be applicable both to NFT + NEP5, rather than core to an NFT itself

WyattMufson:
@saltyskip I think that a separate NEP for `approve`, `clearApprove` and `getApproved` would be good. While the syntax will change slightly depending on fungible vs. non-fungible (`transferFrom` would be needed for fungible) - it would be helpful for all asset contracts. I can help write a proposal with you for that. It'd be similar to what I proposed with #92.

MorganDream:
The return type of `Iterator` is very unfriendly from view of SDK.
Most SDK query RPC service of seed nodes, the response is in format of json, iterator is not compatible with json.

> Putting the iterator outside of the contract code lets the invoker decide how to ask for and get just the data they want from the entire set - much more efficient and far simpler on the contract side.
This is not a friendly solution for SDK, either. Constructing a script is not what developer wants to do. If most invokes require such `postScript`, why not customize it on contract side?

> The max array size is 1024.
If http requests for a large size of array, the response body would use pagination.
So I suggest the return type should be like:
`size: totol_number,
page: page_number,
items: tokens_in_array`
And then serialize this object and return it.

erikzhang:
> The difference is that the user can maintain control of their asset, and should they change their mind on the approval, they can call clearApproved.

For an honest DEX, you can change your mind by withdrawing. For a malicious DEX, even if you want to call `clearApproved()`, it's useless because it can call `transfer()` to transfer tokens immediately after your approve. So for me, there is no essential difference between the two solutions. So I would prefer to choose a simpler solution.

> properties would store data on-chain while tokenURI points to some data/JSON that is stored off-chain.

You can store data on-chain, and returns a URI in "data URI scheme". [RFC 2397](https://tools.ietf.org/html/rfc2397)

>  If most invokes require such postScript, why not customize it on contract side?

I think the main reason is the cost. Performing complex operations in a contract requires a lot of fees, so it is more reasonable to perform these operations locally.

MorganDream:
> I think the main reason is the cost. Performing complex operations in a contract requires a lot of fees, so it is more reasonable to perform these operations locally.

I agree. But it seems difficult to deal with it locally when you get the response from RPC invoke, the iterator would be lost in RPC response, you cannot get nothing from it.
@hal0x2328 has proposed a post script method to deal with it, but this post script will also be run in VM and cost extra fees, too.

erikzhang:
> @hal0x2328 has proposed a post script method to deal with it, but this post script will also be run in VM and cost extra fees, too.

It can be run in the requested RPC node only.

hal0x2328:
> For an honest DEX, you can change your mind by withdrawing. For a malicious DEX, even if you want to call `clearApproved()`, it's useless because it can call `transfer()` to transfer tokens immediately after your approve. 

How about a DEX that simply ceases operation suddenly (not maliciously) and provides no method for the DEX wallet owner to sign the token ownership back over to the user?

> You can store data on-chain, and returns a URI in "data URI scheme". [RFC 2397](https://tools.ietf.org/html/rfc2397)

The current draft for `tokenURI` says "The URI MUST point to a JSON file that conforms to the "NEO NFT Metadata JSON Schema". To clarify, you are you saying it's OK to embed the NEO NFT Metadata JSON in RFC-2397 format directly in the `tokenURI` on-change storage key, and just add optional fields to that metadata to store specific read-only properties of the token?

This could work, but users must be careful about assumptions of how `tokenURI` storage is maintained. In the current draft, I would expect that the contract owner could update this data after deployment in case the server hosting the URI moved to a different domain, etc. 

But if we are talking about storing immutable properties data in a `tokenURI` base64 blob, a contract owner must not be allowed to _ever_ change this data once it is created, otherwise the value of the NFT could be manipulated by the contract owner by changing its properties and what it represents.

WyattMufson:
> How about a DEX that simply ceases operation suddenly (not maliciously) and provides no method for the DEX wallet owner to sign the token ownership back over to the user?

Also there's the case where the NFT contract itself bars interaction with a DEX contract (for example, with #92).

RavenXce:
> How about a DEX that simply ceases operation suddenly (not maliciously) and provides no method for the DEX wallet owner to sign the token ownership back over to the user?

How is this a DEX then? A proper DEX can just build into proper withdrawal methods / escape hatches for byzantine behaviour, intentional or not. It's pointless to only cover one type of byzantine behaviour when both can be covered with a simpler interface. After all if a user can't be sure that a DEX provides methods to return ownership, he also can't be sure that it won't have methods to steal funds. So he should just be sure of the former, making `approve` useless.

--Minor rant--

Honestly `approve` methods are too error-prone by both SC authors and users. For example from the suggestion I have no idea what the expected result or event should be for repeated `approve` calls, or if the token can be transferred away during approval, or how would the approved SC know if the approval has been rescinded (to say "cancel" the auction - do we really want to delegate observing this to the dApp / clients etc)? More importantly users have no idea what `approve` means anyway. So it's way simpler to just move tokens.

As someone who built DEXs on multiple blockchains, my general conclusion is that the operation is a failed attempt to cover a deficiency in the ERC-20 spec which resulted in a bigger mistake. Basically the whole spec was made too hastily.

If `approve` can be avoided it should.

hal0x2328:
> So it's way simpler to just move tokens.

I'll defer to your experience here but we don't have a NEP defining how a DEX should operate with regard to withdrawal guarantees. 

If we have no `approve` method/behavior defined in any NEP, and we are offloading the responsibility on the DEX to provide a failsafe "escape hatch" for users, then we should have a verifiable standard that must be followed in order to call yourself a DEX on Neo.

RavenXce:
> we are offloading the responsibility on the DEX to provide a failsafe "escape hatch" for users

We are already offloading the responsibility of all other correct & safe operation to the DEX anyway.

>  then we should have a verifiable standard that must be followed in order to call yourself a DEX on Neo.

Should we have a standard for how auctions should be carried out too? Or should we debate whether Neo is allowed to call itself decentralized because it is running on only N nodes? Specs are meant for contract interoperability, not labelling for the sake of labels.

To elaborate, a common standard for tokens and NFTs is useful, whereas labeling dApps isn't - who takes the blame if a DEX on first glance has the correct operations but on closer inspection / audit doesn't fulfill whatever "DEX" is supposed to mean? We all know what a DEX is is already highly contended so it's silly to drag a token spec into that.

Of course I would agree the common assumption is for a DEX is that for it to actually be non-custodian, and there should be withdrawal methods regardless of any external byzantine behavior (short of the blockchain consensus itself). But I think that is really besides the point here.

At the end of the day, withdrawal guarantees is one of the things that a dApp has to decide / proclaim, just like it's up to the token to decide any additional KYC, admin controls, etc. it wants to build in. The user or the community still has to verify if that is true regardless of "verifiable standard".

hal0x2328:
My only goal here is to protect users from losing funds through poor practices. If DEX now has no meaning then we should just stop using the term altogether.

RavenXce:
Sure, I have similar goals from other perspectives. I would love to support the spec and therefore am eager to quickly shut down what I feel is a can of worms. No ill will intended.

WyattMufson:
I think we could remove `approve`, `getApproved` and `clearApproved` from this standard and debate adding them to a future NEP elsewhere as @saltyskip had mentioned.

WyattMufson:
@erikzhang @steven1227 we've made the changes to the proposal

steven1227:
@erikzhang I agree with @hal0x2328 that there should be two filed: `tokenURI` and `property`.

Then reason is from @hal0x2328 : 

First, the data URI type must be base64-encoded according to the RFC. That means the smart contract or any iterator script will be unable to evaluate the content of the token properties since there is no Interop method to decode base64, and it would be very expensive GAS-wise to use a custom base64-decode function running inside the VM.

Second, the http and https URI schemes would need to be mutable (for instance, if the token owner needed to move the database to a new domain or migrate to NeoFS when it is released, etc), but the data scheme would need to be immutable if it is used to store properties.

So if immutability of a tokenURI is conditional and depends on what type of data is stored in it, it could lead to contracts written where the owner could rewrite the data URI at any time after the token purchase, completely changing what it represents and its value. Essentially it destroys non-fungibility if the update logic isn't written correctly, so the contract developer and users have to verify the logic to ensure this will never happen. I think it is better to remove the ambiguity and just say that tokenURI is mutable and properties is immutable.

I fully agree with that and there might be case the smart contract need to use the properties of NFT in order to do next decision.

for the enumerator in `tokensOf`, I like this but I am not sure the convenience for this enumerator. @saltyskip do you have any comment on this ?

saltyskip:
Seperation of tokenURI + property LGTM, I agree that enumerator is currently not so convenient for developer currently.

But the issue for supporting Enumeratiors in neon-js is at least being tracked here

https://github.com/CityOfZion/neon-js/issues/437

So if we can handle them in Neon-js at least I think it is fine, as that is what most developers are using as sdk for NEO blockchain

steven1227:
@saltyskip thanks Andrei. It looks like the supporting of Enumerator is mentioned by neon-js several days ago, but not sure what is the process now. @hal0x2328 @snowypowers 
And not only neon-js, but also need other SDK to support it.

erikzhang:
Without tokens, how do we traverse all NFTs?

erikzhang:
> there is no Interop method to decode base64

If it is necessary, we can add these Interops.

> So if immutability of a tokenURI is conditional and depends on what type of data is stored in it, it could lead to contracts written where the owner could rewrite the data URI at any time after the token purchase, completely changing what it represents and its value.

Even with `properties`, the contract owner can modify everything in the contract if he wish. He can add an administrator account when developing the contract and then use it to modify the storages of the contract.

WyattMufson:
> Without tokens, how do we traverse all NFTs?

Why is this needed? Why would this be required for all NFT contracts?

> Even with `properties`, the contract owner can modify everything in the contract if he wish. He can add an administrator account when developing the contract and then use it to modify the storages of the contract.

The same could be said about `name`, `symbol`, `totalSupply` and every other part of storage in a contract. I'm not sure what your point is here. With `tokenURI`, by definition, some of the data is stored off-chain (like images) which can be easily changed (by design). `properties` would be meant to be immutable data specific to the use case of the contract, similar to how `name` and `symbol` don't change but on a per-token basis. Obviously a malicious actor could modify it, just as they could interfere with all other aspects of the contract. This is a standard for an interface for interoperability - it sets an expectation for how a contract is meant to operate.

WyattMufson:
> So if we can handle them in Neon-js at least I think it is fine, as that is what most developers are using as sdk for NEO blockchain

@saltyskip I can also look into adding support to [O3Labs/neo-utils](https://github.com/O3Labs/neo-utils) and [Ryucoin/neovm-utils](https://github.com/Ryucoin/neovm-utils) for Golang and Swift support.

saltyskip:
> @saltyskip I can also look into adding support to [O3Labs/neo-utils](https://github.com/O3Labs/neo-utils) and [Ryucoin/neovm-utils](https://github.com/Ryucoin/neovm-utils) for Golang and Swift support.

I think the most important thing right now is to have a deployed contract on testnet with the latest functionality, and then we can adjust NFT explorer to meet latest standards before we add support in those SDKS

steven1227:
@saltyskip @hal0x2328 just checked with @snowypowers that he is not working on this support currently. Can anyone here do that? I think this is the most important utility for NFT in wallet and exchange.

steven1227:
> > Without tokens, how do we traverse all NFTs?
> 
> Why is this needed? Why would this be required for all NFT contracts?
> 
I believe this is needed because most wallets and explorers will list all/part of the NFTs for users to review.

hal0x2328:
> > > Without tokens, how do we traverse all NFTs?
> > 
> > 
> > Why is this needed? Why would this be required for all NFT contracts?
> 
> I believe `tokens` is still needed because most wallets and explorers will list all/part of the NFTs for users to review.

I'm not opposed to having a `tokens` function since it is literally 2 lines of code to implement. But I think explorers should use the minting `transfer` event to track and index new NFTs instead of running each page view through an RPC request, and a wallet should use `tokensOf` to just get the tokens for the addresses in the wallet.

WyattMufson:
@steven1227 I agree with @hal0x2328 on `tokens` since it is minimal work for the developers it could be included. But is it really needed, given that we can track ownership with transfer events, similar to how NEP-5 tokens are indexed?

steven1227:
@WyattMufson OK. Agree with your guys. Then looks like we get agreement on this standard?

WyattMufson:
@steven1227 @erikzhang so then what would the next step be? Merge this in and have someone deploy an example onto the testnet as @saltyskip had mentioned?

nep-11.mediawiki
```diff
@@ -0,0 +1,256 @@
...
+</pre>
+
+Returns a distinct Uniform Resource Identifier (URI) for a given asset.
+URIs are defined in [https://tools.ietf.org/html/rfc3986 RFC 3986].
+This URI contains optionally mutable properties for the given NFT.
+
+The parameter <code>tokenid</code> SHOULD be a valid NFT.
+If not, this method SHOULD <code>throw</code> an exception.
+
+The URI MUST point to a JSON file that conforms to the "NEO NFT Metadata JSON Schema":
```
erikzhang: Both `tokenURI()` and `properties()` return the "NEO NFT Metadata JSON Schema"?

nep-11.mediawiki
```diff
@@ -0,0 +1,256 @@
...
+    },
+    "image": {
+      "type": "string",
+      "description": "A URI pointing to a resource with mime type image/* representing the asset to which this NFT represents. Consider making any images at a width between 320 and 1080 pixels and aspect ratio between 1.91:1 and 4:5 inclusive."
+    }
+  }
+}
+</pre>
+
+The schema must contain a <code>title</code>, <code>type</code> and <code>properties</code>.
```
erikzhang: I don't understand this line. I think the returned JSON file must contains `name`, `description`, and `image`.

nep-11.mediawiki
```diff
@@ -0,0 +1,256 @@
...
+
+Any custom fields can be included in the JSON file.
+Each field contains a <code>type</code> and <code>description</code>.
```
erikzhang: And what's this?

nep-11.mediawiki
WyattMufson: The overall json has fields `title`, `type` and `properties`. `properties` has `name`, `description` and `image`.

nep-11.mediawiki
WyattMufson:
```
"name": {
    "type": "string",
    "description": "Identifies the asset to which this NFT represents"
}
```

You can see that `name` has `type` and `description`

nep-11.mediawiki
WyattMufson: `tokenURI()` can return an http/https or data that points to a JSON following the schema. `properties()` returns an NVM object serialized in the same format.

nep-11.mediawiki
erikzhang: It's the **schema** of the JSON file, not the JSON file itself.

nep-11.mediawiki
erikzhang: And they have the same fields? This makes me feel very strange.

WyattMufson: Why? Won't it make it easier to parse for the client if the two are in the same format?

nep-11.mediawiki
WyattMufson: I'm not sure I understand

WyattMufson: Should I change the word *schema* to *JSON*?

erikzhang: You can learn JSON schema at here: http://json-schema.org/

nep-11.mediawiki
erikzhang: Both them have the same fields, which one is correct? For example, I return different name from `tokenURI()` and `properties()`. This is very confusing.

nep-11.mediawiki
WyattMufson: Ahhh, thank you for that - understood. That makes a lot more sense.

nep-11.mediawiki
WyattMufson: Was a misunderstanding

WyattMufson:
@erikzhang misunderstood what the JSON schema meant - can fix that. The goal is really just to have two jsons for `tokenURI()` and `properties()`. `tokenURI()` would return a JSON that includes off-chain data, such as image. `properties()` would return a JSON with immutable on-chain data, such as the name.

WyattMufson:
@erikzhang do we need `name`, `description` and `image` to be required in `tokenURI()`? Would name make more sense in `properties()`?

WyattMufson:
And is `description` needed?

erikzhang:
I think it's hard to specify in a standard which fields are mutable and which ones are immutable. These are all things related to the application logic. And it is also unreliable to use two methods to distinguish whether the data are immutable. In this case, I still think that we only need one method is enough.

WyattMufson:
I think the issue arises from the fact that `tokenURI` can point to a JSON stored off-chain. Some things will need to be stored on-chain (name, DNA, statistics, etc).

WyattMufson:
I also think the specific fields will be dictated by application logic - so it'd be hard to require any specific property.

erikzhang:
Even if data is stored on-chain, it still can be mutable. So we can keep `properties()` only, and remove `tokenURI()`. Then we return both mutable and immutable data from the same method.

WyattMufson:
@erikzhang I think that'd work. I'll make the relevant edits when I get a chance

WyattMufson:
I've opened a PR on the Splyse branch with the update: https://github.com/Splyse/proposals/pull/3

nep-11.mediawiki
```diff
@@ -180,32 +179,13 @@ The returned tokens should be in the following format:
 It must have the fields <code>id</code>, <code>uri</code> and <code>properties</code>.
```
erikzhang: This line should be modified too. And can we just return `id`?

WyattMufson: Oh you're right, I missed that one. Should it not be `id` and `properties`?

erikzhang: It's easier for a contract to return an enumerator of `id`.

RavenXce:
Hi, besides that final typo, are we approaching finalization of spec? Can we give an ETA for final spec?

Also is there a reference implementation and are there live NFTs already?

nep-11.mediawiki
WyattMufson: If we return just `id` won't the client then have to get the `properties` in a separate network call? Would it not be more efficient to get the full NFT?

steven1227:
@RavenXce I believe we are approaching the final step. There is no reference implementation on this updated version. There is no live NFTs but I know some projects are implementing it, I know a NFT-based game is almost accomplished and I will involve him here.

nep-11.mediawiki
WyattMufson: Opened a PR @hal0x2328 https://github.com/Splyse/proposals/pull/4

WyattMufson:
@RavenXce I think once my last PR to the splyse branch is merged in it'll be ready. Ryu will have an implementation of this out on the testnet in 1-2 weeks after this is finalized.

WyattMufson:
@erikzhang @steven1227 should be good to go

nep-11.mediawiki
WyattMufson: Fixed

erikzhang: There is currently no way to return an enumerator of a complex data structure. So the best way is just returning the `id`.

nep-11.mediawiki
```diff
@@ -0,0 +1,231 @@
...
+The parameter <code>tokenid</code> SHOULD be a valid NFT.
+If not, this method SHOULD <code>throw</code> an exception.
+
+====Metadata JSON Schema====
+
+<pre>
+{
+  "title": "Asset Metadata",
+  "type": "object",
+  "properties": {
```
erikzhang: Since we have removed the `tokenURI()` from methods, we should add a new similar field into the "Metadata JSON Schema". This new field could be optional.

nep-11.mediawiki
hal0x2328: > There is currently no way to return an enumerator of a complex data structure. So the best way is just returning the `id`.

I've been writing my contracts so that the enumerator returns the objects directly from storage where they are saved as a serialized bytearray, then I use the neon-js `StackItem.deserialize` method to parse them. This way I only need to do a single RPC request instead of dozens or more to get the same data.

erikzhang: In this way, the way the object is stored needs to be specified. However, the storage contents of different contracts may not be consistent. So I think the easiest way is to divide it into two steps, the first step returns the enumerator of id, and the second step uses id to get the details.

nep-11.mediawiki
WyattMufson: Why?

nep-11.mediawiki
WyattMufson: Don't we specify the format in the proposal?

nep-11.mediawiki
erikzhang: For example, each NFT can have a web page that can display some details or pictures.

WyattMufson: But do we need to specify a specific field? Isn't there the optional `image` field? Or `tokenURI` would kind of be a primary URI for that token?

nep-11.mediawiki
erikzhang: Because each contract may have special data structures and storage, our standards should only specify interfaces, and should not specify implementations unless it is necessary.

nep-11.mediawiki
erikzhang: We've removed `tokenURI()`. So I'm asking to add it as a field of "Metadata JSON Schema", which will be returned by `properties()`.

nep-11.mediawiki
WyattMufson: I'm pretty indifferent. @hal0x2328?

nep-11.mediawiki
WyattMufson: So one of the items in `properties` in the schema will be `tokenURI`?

erikzhang: Yes.

WyattMufson:
@erikzhang @hal0x2328 I've opened another PR addressing the changes: https://github.com/Splyse/proposals/pull/5

I agree with Erik that the standard shouldn't be enforcing a specific implementation and only the interface. We could certainly define an extension NEP with a new method to return a specific enumerator but that's probably outside of the scope of this specific NEP.

WyattMufson:
@erikzhang @steven1227 that should do it hopefully

erikzhang:
It is perfect now. We only need a few implementations to turn it into a `Final` state.

corollari:
Would it be possible to reconsider the JSON schema? There's a few questions that may arise in the current version (and have actually arised in the ERC721 standard):
- What is the purpose of `title`? Can it be different from `"Asset Metadata"`?
- What about the several `"type"` properties? Is it possible to change these to something else, and, if so, what possible values can it hold?
- Does the schema just describe the general properties and values of a JSON object or does it actually describe a strict format? Eg: should the spaces inside the format be kept the way they are in the specification? 

Given that messages in this format will have to be stored inside the blockchain it may be worth looking into removing unnecessary information or specifying it further.

WyattMufson:
@corollari Erik and I had this conversation before. You're just misunderstanding what it's saying. That's just the [schema](http://json-schema.org/), not an example JSON.

An example implementation would be:
```
{
   "name": "Slime",
   "description": "A slime",
   "image": "{some image URI}",
   "tokenURI": "{some URI}",
}
```

or even simply:
```
{
   "name": "Slime 2",
   "description": "A slime"
}
```

corollari:
@WyattMufson I see, my bad then, should have read the whole discussion before commenting.

WyattMufson:
@corollari no worries, I made the same mistake!

corollari:
@WyattMufson I submitted a PR adding these examples and specifying several details so that the description is easier to follow.

WyattMufson:
@hal0x2328 check out that PR

corollari:
Also, it may be interesting to consider changing the property name "tokenURI" to "external_url" in order to follow the non-official standard that is in use by some ethereum projects [1]. I don't think that this is an important change but given that tokenURI is an arbitrary name, changing it may reduce developer friction for developers coming from ethereum.

[1] https://docs.opensea.io/docs/2-adding-metadata

WyattMufson:
I think the goal of having the property be `tokenURI` is to keep it consistent with the method from Ethereum, but to have it just be one of the properties. @erikzhang?

WyattMufson:
Since `tokenURI` isn't necessarily a URL. It could be an image, data, url, etc

corollari:
I know, I just wanted to point out that non-official standard and have a discussion about the trade-offs. But I have no strong feelings about it so if you think that changing it is not a good idea then let's keep it the way it is.

WyattMufson:
I don't have any particularly strong feelings. I think @erikzhang might, so I'd wait on him

erikzhang:
Both `tokenURI` or `external_url` are fine.

WyattMufson:
@hal0x2328 made this example implementation: https://gist.github.com/hal0x2328/3237fd9f61132cea6b6dd43899947753

erikzhang:
We also need an implementation with decimals greater than zero.

WyattMufson:
The [Ryu NFT Collectibles](https://github.com/Ryucoin/RyuNFTs-NEO) contracts are live on: 
- the MainNet at [c938e29961a8d05fcf2a9847e95aebae9003c8be](https://explorer.o3.network/contracts/c938e29961a8d05fcf2a9847e95aebae9003c8be)
- the TestNet at 1e5b05152dc1a6b4ff121bf5b47bccd380b42350

Tommo-L:
Ready to merge?   I know it's not perfect but something is better than nothing. And recently, we want to make NNS as a NFT, we hope we can complete the NFT standard.

Tommo-L:
Since we already reference counter in neovm, and we have `balanceOf` method which return the total amount of owner's NFTs.
```cs
// Returns the total amount of NFTs owned by the specified address.
public static BigInteger balanceOf(byte[] owner)
```
I think we can change `tokensOf` to `tokensPageOf`, which is invoked by RPC in most cases.

```cs
public static byte[][] tokensPageOf(byte[] owner, int pageNo, int pageSize) // limit the size by a fixed value or reference counter
```

Tommo-L:
> We also need an implementation with `decimals` greater than zero.

Example? We can help write an NFT example in neo-devpack-dotnet.

steven1227:
> Since we already reference counter in neovm, and we have `balanceOf` method which return the total amount of owner's NFTs.
> 
> ```cs
> // Returns the total amount of NFTs owned by the specified address.
> public static BigInteger balanceOf(byte[] owner)
> ```
> 
> I think we can change `tokensOf` to `tokensPageOf`, which is invoked by RPC in most cases.
> 
> ```cs
> public static byte[][] tokensPageOf(byte[] owner, int pageNo, int pageSize) // limit the size by a fixed value or reference counter
> ```

I agree with that. The tokensPageOf method is better than the enumerators. It decreases the complexity from the sdk/client side.

corollari:
A counter-point to the idea of changing tokensOf is that, by applying that change, backwards-compatibility is broken with the current smart contracts that have already been deployed.

gsmachado:
what's missing to be merged?

shargon:
@gsmachado It has conflicts, and I think an implementation should be linked.

vncoelho (commented):
Perhaps we should also modify here in order to explain the use of `onPayment`, @erikzhang @shargon @igormcoelho.

A similar strategy as #126 should be included here for managing transfer `to` deployed contracts.

erikzhang:
I think this will be replaced by #122.

erikzhang:
I will close this since I can't modify the branch. The new PR is in #130.

Issue #33, opened 2018-03-22
NEP5: In C#.NEO, you can't have a method and an event with the same name (e.g. `transfer`) as required in the specification


mwherman2000:
The NEP5 spec (https://github.com/neo-project/proposals/blob/master/nep-5.mediawiki) states that there are to be a Required method and a Required event. Both called `transfer`.  This isn't allowed in C#.NEO.

Is this an error/bug in the specification? Is there a specification that is more current/up-to-date?

erikzhang:
Here is an example:

https://github.com/neo-project/examples-csharp/blob/master/ICO_Template/ICO_Template.cs#L28

mwherman2000:
The next line in the example is what I've referring to: https://github.com/neo-project/examples-csharp/blob/master/ICO_Template/ICO_Template.cs#L29 doesn't conform to the specification or visa versa.

The event is called `transfer` in the specification and `Transferred` in the example.

RavenXce:
The method name is not important and isn't specified in the spec as it is compiled away.  Only the event name matters.

mwherman2000:
NEP5 is really an *interoperability* specification between NEP5 smart contracts and wallets ...that's why it needs to reliable.

Issue #32, opened 2018-03-22
NEP5: Do the required methods have to be declared `public` per the current specification?


mwherman2000:
The NEP5 spec (https://github.com/neo-project/proposals/blob/master/nep-5.mediawiki) states that there is a Required and Optional set of methods that need to be implemented and that they need to be declared public.  Is this spec up-to-date?  For example, the referenced example https://github.com/lllwvlvwlll/Woolong/blob/master/Woolong/Woolong.cs, declares these methods as private.  Is there a specification that is more current/up-to-date?

PR #44, opened 2018-05-17
NEP-5 Amendment

erikzhang:
This amendment of NEP-5 standardizes the implementation details of each method.

### DIFF
--- a/nep-5.mediawiki
+++ b/nep-5.mediawiki
@@ -98,7 +98,7 @@ If the transfer is not processed, the function SHOULD return <code>false</code>.
 ====transfer====
 
 <pre>
-public static event Action<byte[], byte[], BigInteger> transfer
+public static event transfer(byte[] from, byte[] to, BigInteger amount)
 </pre>
 
 MUST trigger when tokens are transferred, including zero value transfers.
--- a/nep-5.mediawiki
+++ b/nep-5.mediawiki
@@ -69,7 +69,7 @@ Returns the token balance of the <code>account</code>.
 
 The parameter <code>account</code> SHOULD be a 20-byte address. If not, this method SHOULD <code>throw</code> an exception.
 
-If the <code>account</code> is an unused address, this method SHOULD return <code>0</code>.
+If the <code>account</code> is an unused address, this method MUST return <code>0</code>.
 
 ====transfer====
 
@@ -81,9 +81,9 @@ Transfers an <code>amount</code> of tokens from the <code>from</code> account to
 
 The parameters <code>from</code> and <code>to</code> SHOULD be 20-byte addresses. If not, this method SHOULD <code>throw</code> an exception.
 
-The parameter <code>amount</code> SHOULD be greater than or equal to <code>0</code>. If not, this method SHOULD <code>throw</code> an exception.
+The parameter <code>amount</code> MUST be greater than or equal to <code>0</code>. If not, this method SHOULD <code>throw</code> an exception.
 
-The function SHOULD return <code>false</code> if the <code>from</code> account balance does not have enough tokens to spend.
+The function MUST return <code>false</code> if the <code>from</code> account balance does not have enough tokens to spend.
 
 If the method succeeds, it MUST fire the <code>transfer</code> event, and MUST return <code>true</code>, even if the <code>amount</code> is <code>0</code>, or <code>from</code> and <code>to</code> are the same address.
 
@@ -103,7 +103,9 @@ public static event transfer(byte[] from, byte[] to, BigInteger amount)
 
 MUST trigger when tokens are transferred, including zero value transfers.
 
-A token contract which creates new tokens SHOULD trigger a <code>transfer</code> event with the <code>from</code> address set to <code>null</code> when tokens are created.
+A token contract which creates new tokens MUST trigger a <code>transfer</code> event with the <code>from</code> address set to <code>null</code> when tokens are created.
+
+A token contract which burns tokens MUST trigger a <code>transfer</code> event with the <code>to</code> address set to <code>null</code> when tokens are burned.
 
 ==Implementation==
 
--- a/nep-5.mediawiki
+++ b/nep-5.mediawiki
@@ -85,7 +85,7 @@ The parameter <code>amount</code> SHOULD be greater than or equal to <code>0</co
 
 The function SHOULD return <code>false</code> if the <code>from</code> account balance does not have enough tokens to spend.
 
-If the method succeeds, it MUST fire the <code>transfer</code> event, and MUST return <code>true</code>, even if the <code>amount</code> is <code>0</code>.
+If the method succeeds, it MUST fire the <code>transfer</code> event, and MUST return <code>true</code>, even if the <code>amount</code> is <code>0</code>, or <code>from</code> and <code>to</code> are the same address.
 
 The function SHOULD check whether the <code>from</code> address equals the caller contract hash. If so, the transfer SHOULD be processed; If not, the function SHOULD use the SYSCALL <code>Neo.Runtime.CheckWitness</code> to verify the transfer.
 
--- a/nep-5.mediawiki
+++ b/nep-5.mediawiki
@@ -45,7 +45,7 @@ This method MUST always return the same value every time it is invoked.
 public static string symbol()
 </pre>
 
-Returns a short string symbol of the token managed in this contract. e.g. <code>"MYT"</code>. This symbol '''SHOULD''' be short (3-8 characters is recommended), with no whitespace characters or new-lines and '''SHOULD''' be limited to the uppercase latin alphabet (i.e. the 26 letters used in English).
+Returns a short string symbol of the token managed in this contract. e.g. <code>"MYT"</code>. This symbol SHOULD be short (3-8 characters is recommended), with no whitespace characters or new-lines and SHOULD be limited to the uppercase latin alphabet (i.e. the 26 letters used in English).
 
 This method MUST always return the same value every time it is invoked.
 
--- a/nep-5.mediawiki
+++ b/nep-5.mediawiki
@@ -19,55 +19,91 @@ As the NEO blockchain scales, Smart Contract deployment and invocation will beco
 
 In the method definitions below, we provide both the definitions of the functions as they are defined in the contract as well as the invoke parameters.
 
-This standard defines two method types:
-
-* '''(Required)''' : methods that are present on all NEP5 tokens.
-
 ===Methods===
 
 ====totalSupply====
 
-* Syntax: <code>public static BigInteger totalSupply()</code>
+<pre>
+public static BigInteger totalSupply()
+</pre>
 
-* Remarks: "totalSupply" returns the total token supply deployed in the system.
+Returns the total token supply deployed in the system.
 
 ====name====
 
-* Syntax: <code>public static string name()</code>
+<pre>
+public static string name()
+</pre>
 
-* Remarks: "name" returns the token name.
+Returns the name of the token. e.g. <code>"MyToken"</code>.
+
+This method MUST always return the same value every time it is invoked.
 
 ====symbol====
 
-* Syntax: <code>public static string symbol()</code>
+<pre>
+public static string symbol()
+</pre>
 
-* Remarks: "symbol" returns the token symbol.
+Returns a short string symbol of the token managed in this contract. e.g. <code>"MYT"</code>. This symbol '''SHOULD''' be short (3-8 characters is recommended), with no whitespace characters or new-lines and '''SHOULD''' be limited to the uppercase latin alphabet (i.e. the 26 letters used in English).
+
+This method MUST always return the same value every time it is invoked.
 
 ====decimals====
 
-* Syntax: <code>public static byte decimals()</code>
+<pre>
+public static byte decimals()
+</pre>
 
-* Remarks: "decimals" returns the number of decimals used by the token.
+Returns the number of decimals used by the token - e.g. <code>8</code>, means to divide the token amount by <code>100,000,000</code> to get its user representation.
+
+This method MUST always return the same value every time it is invoked.
 
 ====balanceOf====
 
-* Syntax: <code>public static BigInteger balanceOf(byte[] account)</code>
+<pre>
+public static BigInteger balanceOf(byte[] account)
+</pre>
 
-* Remarks: "balanceOf" returns the token balance of the '''account'''.
+Returns the token balance of the <code>account</code>.
+
+The parameter <code>account</code> SHOULD be a 20-byte address. If not, this method SHOULD <code>throw</code> an exception.
+
+If the <code>account</code> is an unused address, this method SHOULD return <code>0</code>.
 
 ====transfer====
 
-* Syntax: <code>public static bool transfer(byte[] from, byte[] to, BigInteger amount)</code>
+<pre>
+public static bool transfer(byte[] from, byte[] to, BigInteger amount)
+</pre>
 
-* Remarks: "transfer" will transfer an '''amount''' of tokens from the '''from''' account to the '''to''' account. Contract '''MUST''' check the <code>payable</code> flag of the receiver contracts to decide whether it should transfer the tokens to those contracts.
+Transfers an <code>amount</code> of tokens from the <code>from</code> account to the <code>to</code> account.
+
+The parameters <code>from</code> and <code>to</code> SHOULD be 20-byte addresses. If not, this method SHOULD <code>throw</code> an exception.
+
+The parameter <code>amount</code> SHOULD be greater than or equal to <code>0</code>. If not, this method SHOULD <code>throw</code> an exception.
+
+The function SHOULD return <code>false</code> if the <code>from</code> account balance does not have enough tokens to spend.
+
+If the method succeeds, it MUST fire the <code>transfer</code> event, and MUST return <code>true</code>, even if the <code>amount</code> is <code>0</code>.
+
+The function SHOULD check whether the <code>from</code> address equals the caller contract hash. If so, the transfer SHOULD be processed; If not, the function SHOULD use the SYSCALL <code>Neo.Runtime.CheckWitness</code> to verify the transfer.
+
+If the <code>to</code> address is a deployed contract, the function SHOULD check the <code>payable</code> flag of this contract to decide whether it should transfer the tokens to this contract.
+
+If the transfer is not processed, the function SHOULD return <code>false</code>.
 
 ===Events===
 
 ====transfer====
 
-* Syntax: <code>public static event Action<byte[], byte[], BigInteger> transfer</code>
+<pre>
+public static event Action<byte[], byte[], BigInteger> transfer
+</pre>
 
-* Remarks: The "transfer" event is raised after a successful execution of the "transfer" method.
+MUST trigger when tokens are transferred, including zero value transfers.
+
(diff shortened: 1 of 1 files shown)


shargon:
It's perfect for me

mwherman2000:
There are several issues related to the completeness of the NEP-5 specification from an implementation and testing/quality assurance perspective.  For example, <http://docs.neo.org/en-us/exchange/v2.7.3.html#dealing-with-nep-5-assets-transactions> is a more complete, more definitive, more authoritative specification for NEP-5 than than the NEP-5 specification itself.   This content needs to be either (a) moved/copied into NEP-5, or (b) there needs to be a new NEP-5 that defines the basic architecture of a NEO smart contract (e.g. the Main method, if-then-else dispatching, individual callable methods).

You can find a more detailed discussion and motivation for the above here:
<https://github.com/neo-project/proposals/issues/33>
and here
<https://github.com/neo-project/proposals/issues/32>

The specification needs to be corrected/updated. It's not strong enough to be implemented reliably on the NEO platform as it is written. It needs to be a specification ...not a high-level guide.

mwherman2000:
The different formats and lengths of an NEO Account address are applicable across the entire NEO architecture and should be contained in their own NEP specification - and referenced by NEP-5.  They shouldn't be defined in NEP-5. The new NEP should also define specific named constants for these lengths.

image

erikzhang:
@mwherman2000 NEP is Neo enhancement proposal, not the yellow paper.

NEO yellow paper is on the way.

mwherman2000:
@erikzhang What yellow paper? Do you have a link to a draft?

These are highly relevant topics as far as the NEP-5 discussion is concerned and the discussion can't assume anything about a document that no one has seen.  Right?

erikzhang:
@mwherman2000 I mean, NEO's basic design will be defined in the upcoming yellow paper, not in the NEO Enhancement Proposals. NEP is for new proposals.

mwherman2000:
@erikzhang when can we expect to see it? Do you have a draft version?

erikzhang:
I don't have a draft yet. Maybe I should create a new repository for it and anyone can contribute to it.

DaveOnBlocks:
> The function SHOULD check whether the from address equals the caller contract hash. If so, the transfer SHOULD be processed; If not, the function SHOULD use the SYSCALL Neo.Runtime.CheckWitness to verify the transfer.

Is this implying that sending a transfer(address1,address1,amount) should succeed?

DaveOnBlocks:
> If the to address is a deployed contract, the function SHOULD check the payable flag of this contract to decide whether it should transfer the tokens to this contract.

How does one check the payable flag?

erikzhang:
> The function SHOULD check whether the from address equals the caller contract hash. If so, the transfer SHOULD be processed; If not, the function SHOULD use the SYSCALL Neo.Runtime.CheckWitness to verify the transfer.

>Is this implying that sending a transfer(address1,address1,amount) should succeed?

It means that if contract A calls the transfer method of contract B, and parameter `from` is the scripthash of contract A, then the transfer should be processed.

erikzhang:
> How does one check the payable flag?

Use the API `Neo.Contract.IsPayable`

https://github.com/neo-project/neo/blob/b0b0646fb86b350e57f9010ca88ca373144d1353/neo/SmartContract/StateReader.cs#L127

nep-5.mediawiki
```diff
@@ -19,55 +19,91 @@ As the NEO blockchain scales, Smart Contract deployment and invocation will beco
...
 
-* Syntax: <code>public static event Action<byte[], byte[], BigInteger> transfer</code>
+<pre>
+public static event transfer(byte[] from, byte[] to, BigInteger amount)
+</pre>
+
+MUST trigger when tokens are transferred, including zero value transfers.
 
-* Remarks: The "transfer" event is raised after a successful execution of the "transfer" method.
+A token contract which creates new tokens SHOULD trigger a <code>transfer</code> event with the <code>from</code> address set to <code>null</code> when tokens are created.
```
nickfujita: +1000 on this!

This was not in the original NEP5, but block explorers like neotracker or neoscan depend on this event to be called when tokens are minted. So there have been cases where this notification was not emitted at the time of minting, and resulted in confusion for users who would see their tokens in their wallets, but not on the explorer. Until they make a transfer.

Hopefully this will prevent that from happening in the future.

That being said, would it be possible to update this from a `SHOULD` to a `MUST`?

nickfujita: After further consideration, I am starting to wonder if it makes more sense to remove this clause from the `transfer` event. The ask is to have the `transfer` event triggered from a method that is not a part of the NEP5 standard. Whether it be though a token sale, direct allocation method, or some dynamic token creation event, these all result in an action more akin to that of a `mintToken` or `mint` event, rather than a `transfer` that is inferred to be a `mint` when the `from"`address is `null`.

The desire for a separate event for token creation (and even token burn) have been expressed a few times in other posts.

https://github.com/neo-project/proposals/pull/39

https://github.com/CityOfZion/neo-python/issues/457

I would like to reaffirm the proposal stated in the latter thread, with more urgency on the `mint` or `mintToken` event over the `burn` event.

#### mint or mintToken
```
public static event mint(byte[] to, BigInteger amount)
public static event mintToken(byte[] to, BigInteger amount)
```
MUST trigger when tokens are minted.


#### burn or burnToken
```
public static event burn(byte[] from, BigInteger amount)
public static event burnToken(byte[] from, BigInteger amount)
```
MUST trigger when tokens are burned.

These events will help to increase transparency into contract mint and burn events amongst users, and possibly improve confidence that tokens are not being minted unknowingly.

erikzhang: Too many events. The standard should be as simple as possible.

nickfujita: While I agree that the standard should be simple to understand and implement, shouldn't it also be descriptively accurate?

Rather than adding the SHOULD clause to the `transfer` event, it seems more accurate to create a separate event called `mint`. I understand that the implication that tokens are being transferred from `null` because they are being created, but this implication assumes this prior knowledge from a reader rather than being clear and self explanatory.

Could we please reconsider just the addition of the `mint` event to clean up the notifications being emitted by contracts?

DaveOnBlocks: I don't quite see the need for a mint/burn event myself either. Seeing funds created/destroyed by coming from/to null shows that the tokens either came from nowhere or were sent to nowhere. 

+1 on changing the SHOULD to a MUST though. If tokens are being created/destroyed it is good to be able to pick this up without checking TotalSupply constantly and inferring the change. Having the change notification arise out of a certain block is a much cleaner way to show it.

canesin: https://www.ietf.org/rfc/rfc2119.txt

erikzhang:
Let's vote?

mwherman2000:
NEP-5 now talks about **methods** everywhere ...and not "operationX".  

Does this imply that only an NEP-5 SC's detailed methods are the only way to invoke their functionality?  ...i.e. has calling `Main` been deprecated?  When should an app call `Main` and when should it call one of the detailed methods?

shargon:
`The parameter amount MUST be greater than or equal to 0`
to
`The parameter amount MUST be greater than 0` 
?

Why we need a TX with 0 amount?

erikzhang:
> Why we need a TX with 0 amount?

The `transfer` method may be invoked by another contract, and the number may be calculated in that contract, possibly equal to zero. To ensure the success of the entire contract, we should allow a transfer of `0` amount.

RavenXce:
The changes are good but I'm just wondering how it's possible to set `Neo.Contract.IsPayable` on existing contracts.

erikzhang:
> The changes are good but I'm just wondering how it's possible to set Neo.Contract.IsPayable on existing contracts.

I will create new APIs for doing that.

dicarlo2:
lgtm - I particularly like the clarity around transfer events that create and burn tokens.

mwherman2000:
@erikzhang Can you answer these questions? https://github.com/neo-project/proposals/pull/44#issuecomment-395970612

erikzhang:
@mwherman2000 I'm sorry I didn't understand your questions.

mwherman2000:
@erikzhang  There are 2 primary ways that the functionality in a "detail method" in an SC (i.e. any public method other than `Main()`) can be invoked:
1. As an operation string + an array of arguments passed in an invocation of `Main()`, or
2. By executing an `invokefunction` to invoke one of the detail methods directly - bypassing `Main()` (reference:  http://docs.neo.org/en-us/exchange/v2.7.4.html#dealing-with-nep-5-assets-transactions)

The new wording in the NEP-5 Amendment removes any references to approach #1 ...and now only refers to method calls (approach #2).

Is this what was intended? ...or does the wording need to be expanded to cover both approach #1 and approach #2?

erikzhang:
I think you have misunderstood. It is not possible to bypass the main method to invoke the other methods in a contract directly.

mwherman2000:
The NEO documentation "Dealing with NEP-5 Assets Transactions" says you can bypass `Main()`, doesn't it?  

Reference: http://docs.neo.org/en-us/exchange/v2.7.4.html#dealing-with-nep-5-assets-transactions

image

mwherman2000:
The code you linked to shows what? ...that all calls to non-`Main()` SC methods are always invoked by calling `Main()` with the operation string + array of objects as arguments?  Please be patient with me.

erikzhang:
> The code you linked to shows what? ...that all calls to non-Main() SC methods are always invoked by calling Main() with the operation string + array of objects as arguments? Please be patient with me.

Yes, that's it.

mwherman2000:
RE: Erik: "It is not possible to bypass the main method to invoke the other methods in a contract directly."

If the only way to invoke the non-`Main()` methods is by calling Main() with the operation string + array of objects as arguments, then I've got the following questions about the wording in the NEP-5 Amendment:

1. Why does the Amendment use method signatures to document the NEP-5 operations? ...this is very confusing when in reality it doesn't matter what the non-`Main()` method signatures are as long as `Main()` does the right thing when it receives a valid operation and array of object arguments, correct?

2. Why does the Amendment declare the non-`Main()` methods to be `public` when they should be `private` ...if `Main()` is the only method that calls them?  ...why are these methods needed at all?

3. If "it is not possible to bypass the main method to invoke the other methods in a contract directly"., then an NEP-5 compliant SC can be implemented entirely within the `Main()` method with no non-`Main()` methods, correct?  That is, there is no requirement to have any non-`Main()` methods at all in an NEP-5 compliant SC, correct?

4. The simplest specification could be a table, n'est pas?

| operation    | args[0]        | args[1]       | args[2]        | args[3]       | return type |
| ------------- |-------------| -------------|-------------|-------------|-------------|
| `"totalSupply"` |  |  |  |  | `BigInteger` |
| `"name"` |  |  |  |  | `string` |
| `"symbol"` |  |  |  |  | `string` |
| `"decimals"` |  |  |  |  | `byte` |
| `"balanceOf"` | `byte[]` account|  |  |  | `BigInteger` |
| `"transfer"` | `byte[]` from |   `byte[]` to | `BigInteger` amount |  | `bool` |

5. `Action`s  will still need to be defined using "method" notation.  i.e. I don't think they fit into the tabular format very well.

erikzhang:
> If "it is not possible to bypass the main method to invoke the other methods in a contract directly"., then an NEP-5 compliant SC can be implemented entirely within the Main() method with no non-Main() methods, correct? That is, there is no requirement to have any non-Main() methods at all in an NEP-5 compliant SC, correct?

Yes. Theoretically, we can light the bulb by simply controlling the electrons. But we're still going to use the switch, right? Although using the switch, we still control the electrons at the physical level, but this is two different levels of things.

mwherman2000:
@erikzhang , can you *enlighten* me? :-)  To implement an NEP-5 compliant SC, do the non-`Main()` methods need to be implemented or not?  If they do, what is the non-electron justification?

erikzhang:
> To implement an NEP-5 compliant SC, do the non-Main() methods need to be implemented or not?

I don't want to be overly entangled in words. A method of a contract is not specifically a method in some programming languages, but something abstract. You can just think of it as an operation.

mwherman2000:
NEP-5 needs to be usable and useful as an implementation, testing *and* an off-chain/on-chain interoperability specification. @erikzhang, it can't be vague. It can't leave Interface details open to speculation or subjective interpretation.

mwherman2000:
@erikzhang  Would you like me to re-edit NEP-5 based on the above?

shargon:
You only have one entry point. Is `main`. Then you can do whatever you want

mwherman2000:
@shargon Agreed. But this needs to be made absolutely clear in the NEP-5 specification and the Amendment as well as the current specification are misleading.  They both use C# syntax to describe the agruments to `Main()`.  This makes no sense, is confusing and misleading.  Why does the Amendment specifiy an NEP-5 SC use C# syntax to specify the SC has a method

`public static bool transfer(byte[] from, byte[] to, BigInteger amount)`

when it is more simply and unambiguously stated that one of the allowed sets of arguments that can be passed to `public static object Main(string operation, params object[] args)` is:

| operation    | args[0]        | args[1]       | args[2]        | args[3]       | return type |
| ------------- |-------------| -------------|-------------|-------------|-------------|
| `"transfer"` | `byte[]` from |   `byte[]` to | `BigInteger` amount |  | `bool` |

Let's break the pattern of creating a House of Cards.

kemargrant:
At this time there needs to be an allocation of resources to explicitly document the neo smart contract platform. It may be a bounty, stake holder driven, or a person without neo council. 

From the outside looking in Solidity is much more straightforward. I am not a c# dev so I may have some bias, as well as the fact ethereum is very well documented.

Even with well documented code there are still room for mistakes, the latest being the icx contract. The more room left for interpretation is the more rope you have given to the developer.

erikzhang:
> Would you like me to re-edit NEP-5 based on the above?

If we decide to change the wording, then all the NEP will have to be modified so that their style is consistent. But I prefer to use the existing wording.

erikzhang:
> From the outside looking in Solidity is much more straightforward. I am not a c# dev so I may have some bias, as well as the fact ethereum is very well documented.

Neo and NeoContract are not designed for a particular language. Their design goal is to use any programming language for smart contract development. At the same time Neo and NeoVM also have a variety of different language implementations.
So their documents must be an abstract expression. But I agree that we should use as precise a language as possible to describe all the standards.

RavenXce:
FWIW, ethereum also uses solidity to describe a large part of their specs, even though solidity is not the only language that compiles for the evm.

See the recently finalized NFT standard for example: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-721.md

mwherman2000:
@RavenXce, the ETH examples are largely irrelevant because Solidity/ETH SCs don't have the same SC architecture/structure that NEO smart contracts have where a NEO SC invocation is a call to one and only one SC entrypoint method `Main()`.  The sole essence of the NEO smart contract procotol is to pass `operation` and `args` parameters to `Main()` and `Main()` alone.  NEP specifications need to be clear and compatible  and totally unambiguous about this.

Any other functions you define in your C# contract can't be called and aren't call.  So list these methods in the NEP-5 (or any NEP) specification is misleading.

canesin:
Sounds good to me, adding SHOULD recommendations for burning and minting tokens on the `transfer` event is the only request I would make (since it doesn't change the standard practically I don't thing that is a big issue.

Is it possible to just use `null` ? So if is minting `from` is `null` and if burning `to` is `null` in the event and observers can just treat as such ?

mwherman2000:
`null` is not a valid/usable/reliable construct in C#.NEO ...if that is the specific `null` concept you're referring to @canesin.

Reference: https://youtu.be/Nj4-m2o94VE?t=33m32s ...listen to the next *2* slides ...until timecode 35:39  ...2 minutes.

mwherman2000:
RE: Given that, @mwherman2000, what you're suggesting is that the spec should be written in terms of the state of the stack at the beginning of the smart contract script execution.

@dicarlo2 To be perfectly clear, I've never stated, suggested or endorsed defining NEP-5's invocation pararmeters in terms of the NEO VM stack, correct?  *never*.

However, if you reread this thread, we have had two experts, @erikzhang and @shargon, attest that, in fact, there is and needs to be the `Main()` method in a C# smart contract and all invocations of a NEO C# smart contract call the `Main()` method.  It's the only prescribed/supported way to invoke a SC's internal functionality.  Here's the evidence:

- Erik: https://github.com/neo-project/proposals/pull/44#issuecomment-399314648
- Erik: https://github.com/neo-project/proposals/pull/44#issuecomment-399319297
- Shargon:  https://github.com/neo-project/proposals/pull/44#issuecomment-399434394

@dicarlo2 Where/who is the source of your information?  Do you have to unwind/rewind the discussion thread?  Please clarify your sources of your information?  If have to revisit these first principles, let start again.

@dicarlo2 Perhaps you can double check you information offline with @erikzhang  and @shargon.  We need a 3-way confirmation/agreement here.

mwherman2000:
@Ejhfast Exactly because of these situations.  I know the answers and I'm only asking the questions to lead the discussion and get written answers that we can reference in discussions like the one we're having now.

@Ejhfast and @dicarlo2: Are @erikzhang  and @shargon correct or not correct?  See the 3 links quoted above.

[Note: I've edited my previous comment several times to make it as clear and precise as possible.]

Ejhfast:
Yes: Erik, Shargon, and Alex are all correct. There is no inconsistency in
what they are saying. The current C# compiler requires a Main() method.
This does not mean that future compilers or the VM require that. Alex even
linked to a compiler that does not require a Main() method right now

mwherman2000:
@Ejhfast and @dicarlo2: Maybe we're reshaping this discussion topic to: How to represent multiple language bindings in an NEP.

I need on the C# `Main()` binding as the authoritative approach because C# is the reference implementation language for the NEO platform.  If a non-C# language has a different binding, that's a different situation. Perhaps we need to discuss that separately.

However, the core issue is the current convention of using C# method syntax is misleading:

`public static bool transfer(byte[] from, byte[] to, BigInteger amount)`

...and is contributing to the fragility and failures of the NEO platform.

Ejhfast:
You are trying to create a lot of discussion around something that is not a
big deal. At NEX we are building very sophisticated infrastructure on this
chain, and the good and bad parts of NEO become much, much clearer when
doing that. I can think of many important things that should be improved in
NEO right now (e.g., the current fee system), and this discussion is
distracting from that.

It would be much better if the energy put into discussion here was instead
directed at documentation efforts or other forms of ecosystem building. You
seem to be worried about people being confused by standards. There is
general agreement here that the standard is fine. But I think we'd all
agree that more needs to be done to onboard new developers through
tutorials and documentation and new resources -- the community needs that
very much!

mwherman2000:
@Ejhfast I wrote NEP-10. My goal is to promote a develop a trusted ecosystem of hundreds ...maybe thousands of NEPs (and NCPs).  NEP-5 is incredibly important in this regard.  It's like the `Block 0 NEP`.

To allow the current situation of vague and misleading C# method syntax continue is untenable to the point of ludicrous.

The NEO platform is in serious trouble because of its fragile software architecture (https://github.com/neo-project/neo/issues/272#issuecomment-398721473).  More or better developer tutorials isn't going to fix the problem.

lllwvlvwlll:
The proposed changes are straight forward and further enhance the clarity of the standard: **APPROVED**

As @Ejhfast has indicated, the extent of this conversation is a poor use of resources for people who could be focusing on critical platform development (including some that @mwherman2000 has directly requested).  @mwherman2000 I would also recommend auditing your method of seeding discussion.  I am receiving many messages on the discord about the aggressive (and sometimes offensive) nature of these conversations.

mwherman2000:
#RIP

PR #46, opened 2018-05-24
NEP: Smart Contract Standard

erikzhang:
This NEP presents a set of methods that make DApps more accessible to the wallets and clients, and gives the users a more intuitive understanding of DApps.

### DIFF
--- a/dappstandard.mediawiki
+++ b/dappstandard.mediawiki
@@ -1,16 +1,16 @@
 <pre>
   NEP: <to be assigned>
-  Title: DApp Standard
+  Title: Smart Contract Standard
   Author: Erik Zhang <erik@neo.org>
   Type: Standard
   Status: Draft
   Created: 2018-05-24
-  Requires: 10
+  Replaces: 10
 </pre>
 
 ==Abstract==
 
-This NEP presents a set of methods that make DApps more accessible to the wallets and clients, and gives the users a more intuitive understanding of DApps.
+This NEP presents a set of methods that make smart contracts more accessible to the wallets and clients, and gives the users a more intuitive understanding of the contracts.
 
 ==Motivation==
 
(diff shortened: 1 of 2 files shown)


vncoelho:
Hi, @erikzhang,

@ThomasLobker just came with a nice idea, to validate the Dapps with the source.

Maybe it could be included here, something to tell the version and compiler used.

ThomasLobker:
Etherscan has a tool which allow any user to verify the source against deployed contracts. We could make the same functionality and this would improve trust and security for many contracts that have open-source code.

First option: I would propose two extra fields at contract deployment to make it more easy to retrieve and verify the source code.

`Compiler Version`: string to contain the name and version of the compiler, ie. neo-boa-0.6.9
`Source Code`: string to contain the link to the source code, ie: github.com/Project/Source.py

Alternative option: The same kind of parameters, but retrievable through contract invokement.

igormcoelho:
@ThomasLobker this is a very interesting idea. To fully validate source code against AVM/scripthash, it is necessary to have, besides compiler version, the versions of all dependencies... For instance, Neo devpack version, neo-vm version (if detached from neo-compiler), and if the version varies a little bit beyond the tagged version, it can be useful to also have the exact hash of the commit (on the compiler project). Some other things may affect a fully deterministic behavior, such as: C# version, .NET framework version ,etc, etc. But once all theses things are clearly defined, the results and rastreability will be truly amazing.

deanpress:
@ThomasLobker Like the idea, but instead of source code URL, why not compile the source code file hash instead, and then leave it to the user/application how they wish to verify the source code (by comparing file hashes)? Both storing the source code URL in the smart contract or comparing file hashes has some amount of centralization, but the latter is more storage efficient.

ThomasLobker:
@deanpress we could add a checksum field, but this should not replace the link to the source code.

My primary goal is to let any external compiler retrieve and compile the source code to confirm the authenticity of the contract on the blockchain. If you have only the checksum without the source, then this process cannot be (semi) automated anymore. Once the check is done and it results in the same script hash, the external compiler should keep a copy of the source code locally so it cannot be tampered with anymore.

saltyskip:
Can we merge the ideas from here into this proposal?
#69

Processing how sign transaction, and relay it off to the node, is probably the most crucial aspect of DAPP/Client communication protocol. 

This proposal should be an extension or build on top of a proposal that specifies the signature/transaction relay process like in #69

erikzhang:
This proposal is for dapps and #69 is for wallets, I think.

saltyskip:
#69 is for both, we can think about it like DAPP/Wallet communication protocol.

erikzhang:
This should be implemented in manifest: https://github.com/neo-project/neo/issues/287

shargon:
> This should be implemented in manifest: neo-project/neo#287

I think that this could be done in manifest, and it's easier to update.

erikzhang:
Now I think that some methods cannot be moved to the manifest. For example, the `name()` method in NEP-5 should be moved to a separate standard, because all contracts may require a name. This is why I reopened this PR again.

PR #52, opened 2018-06-07
NEP: Smart Contract Verification and Invocation with System Assets

dicarlo2:


nep-11.mediawiki
```diff
@@ -0,0 +1,48 @@
...
+  Replaces: NEP-7
+</pre>
+
+==Abstract==
+
+This NEP defines a scheme for smart contracts to handle system asset transfers.
+
+==Motivation==
+
+Smart contract authors cannot reliably refuse system asset transfers in smart contract invocations. They cannot refuse or react to system asset transfers made with transactions other than invocation transactions to the smart contract address. There is additional complexity in handling the possibility of multiple invocations in the same invocation transaction that expect system assets, typically requiring tracking between calls. At a high level, there are two main issues that this NEP aims to solve:
```
RavenXce: NEP-7 triggers on all transactions, not only invocation transactions.

nep-11.mediawiki
```diff
@@ -0,0 +1,48 @@
...
+
+# Allow smart contracts to refuse system asset transfers.
+# Allow smart contracts to react to system asset transfers.
```
RavenXce: NEP-7 does exactly these.

shargon:
I think we need to have a versioning system before incorporating non-retroactive features

nep-11.mediawiki
canesin: @RavenXce yes, this is a NEP to replace NEP-7 with a better design.

erikzhang:
One big drawback of this proposal is that if users want to participate in ICO, they must use a wallet that constructs `InvocationTransaction`. He could not use any wallet to transfer his assets directly into ICO address.
The advantage of NEP-7 is that users can trigger the default method of a contract without constructing `InvocationTransaction`.

RavenXce:
I also think forcing receives to pass Verification is inherently harder to code safely as it mixes concerns. This is as `Runtime.CheckWitness(contract)` will now pass even though the contract is just receiving assets.

Typically a passing this check is only necessary when extracting assets from contract. So we can just make sure that that does not happen unless we really want to send out assets. 

This proposal makes safe dynamic invokes during a receive impossible, for example. (proxy contracts)

dicarlo2:
> One big drawback of this proposal is that if users want to participate in ICO, they must use a wallet that constructs InvocationTransaction. He could not use any wallet to transfer his assets directly into ICO address. The advantage of NEP-7 is that users can trigger the default method of a contract without constructing InvocationTransaction.

I view this as a downside for developers. This means that not only do we have to watch for notifications from `InvocationTransaction`s, but also every other kind of transaction. This unnecessarily increases the api surface area for something that is trivial for wallet developers to incorporate.

Furthermore, why would a wallet allow a user to participate in an ICO without supporting `InvocationTransaction`s for transfers?

dicarlo2:
@RavenXce I don't quite follow your comment, can you expand a bit more on your points? (Sorry, just want to make sure I fully understand)

dicarlo2:
> Every address in NEO is a smart contract address.

This would look at addresses the same way NEP-7 does, i.e. is it currently a smart contract address.

igormcoelho:
I don't think I clearly understood the idea... can you give a simple implementation example and show how it works (and perhaps contrast on how current NEP-7 works?). This way, the advantages/drawbacks will be clearer for the community to discuss better.

RavenXce:
> @RavenXce I don't quite follow your comment, can you expand a bit more on your points? (Sorry, just want to make sure I fully understand)

Because by returning `true` a smart contract's Verification trigger, it would mean that a deposit would also pass a Runtime.CheckWitness call. This means that it is unsafe to ever accept deposits followed by an arbitrary dynamic invoke on another contract, since the smart contract has passed witness checks. 

An attacking smart contract can drain the NEP-5 tokens of the calling smart contract in this case, for example. Previously this check only has to be done on withdrawal of system assets. So a use case where sending 0.1 gas (as payment) to contract X which invokes contract Y (which may or may not be malicious) as part of the process is no longer possible.

Also, now during the Verification trigger, a smart contract would always have to check whether it is called due to to receiving assets (target of output) or due to it's assets being withdrawn (target of input). It seems more clear for developers if we have a separate trigger for each case (this may be debatable I guess).

Regarding invoking with Invocation `script`, the idea does sound good from a smart contract dev perspective, but the core implementation would messy as NEO doesn't seem to be designed for that. 

This is as the format of `script` is such that the appcall opcode is squeezed between the invoke parameters and the invocation script hash, and they are all in a serialized format. Furthermore, the invocation `script` was designed to be able to be call multiple appcalls within a single transaction, and on multiple smart contracts if desired. There are lots of advantages for these, such as airdrops, etc. So I don't really see a clean way of doing what you suggested without changing how the invocation transaction format is like.

dicarlo2:
@RavenXce That's a very valid point and use-case for separate triggers, thanks for the explanation. If I were to summarize, essentially NEP-11 opens up the possibility for an author to introduce a vulnerability in their code whenever they call another smart contract, such that it's impossible to safely call another smart contract when receiving assets.

> Regarding running Verification with Invocation script (params), the idea does sound good from a smart contract dev perspective, but the core implementation would messy as NEO doesn't seem to be designed for that.

Perhaps I'm missing what you mean by this (and the following paragraph), but this is already how the `Verification` trigger works, it simply runs the script with `Verification` as the trigger instead of `Application`.

> It could be a crazy idea, but someone could even deploy a simple wallet address... so, would NEP-11 behave differently if the wallet address has been deployed or not?

This is exactly the same way NEP-7 works. Quite frankly, I'm a little surprised that this point of NEP-11 is under so much confusion, and really lends further to the fact that NEP-7 was likely rushed before everyone really understood the implications. Take a look at the following lines:
 - https://github.com/neo-project/neo/pull/266/files#diff-eaba41a63c108398e3a5f9c4e5eb33b7R386
 - https://github.com/neo-project/neo/pull/266/files#diff-c087fce63c8ae4e62f21b01fe60d0a2cR609

dicarlo2:
I still think either NEP-7 is not the right approach or could be improved to resolve more of the issues that plague smart contracts (see https://github.com/neo-project/neo/pull/172#issuecomment-392363926). Give me some time to think on NEP-7/NEP-11 to see if I can modify this proposal to come up with a solution that makes everyone happy.

dicarlo2:
On your points:

1. I was actually planning on relaxing this requirement later with another NEP that would allow for temporary storage in `Verification` (and/or `VerificationR` see below) triggers. This not only would allow for tracking within `Verification` triggers, but it would allow solving a current race condition with `Verification` triggers that causes, for example, the need for a refund notification near the end of an ICO.[1] With that said, this does mean we would be back at requiring some sort of tracking between calls, but as I mentioned in [2], this is true of NEP-7 as well.
2. I have to give it some more thought (both for and against), but perhaps we could change this proposal to instead always run a new trigger, `VerificationR`, but with the same semantics as this proposal (i.e. by running the script rather than an arbitrary new method "receive").
3. I'd really like to avoid this because it makes writing applications that interact with the blockchain more complex. For example, a blockchain explorer now needs to listen for and react to notifications not only from invocation transactions, but any arbitrary transaction. More generally, imagine a client api that allows for the developer to relay a transaction and then wait for the events from that transaction - now we have to support that for every transaction, not just invocation transactions. This is how neo-one[3][4] works, and it would make things more complex for developers to have to also consider if an arbitrary transaction may cause notifications, and furthermore, what those notifications might even be. With InvocationTransactions, we require the ABI of the smart contract so that we can automatically parse out the events from the notifications - now the api would almost require an ABI for arbitrary transactions.

[1] In today's system there's a possible condition that `Verification` succeeds where `Application` would fail because of other transactions in the new block. E.g. `mintTokens` may pass the `Verification` check because they're checked independently of the other transactions in the block, whereas it might fail during the `Application` and require a refund notification because the ICO reached the limit due to previous transactions in the block.
[2] https://github.com/neo-project/neo/pull/172#issuecomment-392363926
[3] https://github.com/neo-one-suite/neo-one/blob/master/packages/neo-one-client/src/types.ts#L597
[4] https://github.com/neo-one-suite/neo-one/blob/master/packages/neo-one-client/src/types.ts#L572

dicarlo2:
I'll work on a rough implementation over the next few days of NEP-11 (possibly with a `VerificationR` trigger rather than a `Verification` trigger) so that we can have something more concrete to discuss.

dicarlo2:
I've added an implementation sketch and updated NEP-11. Here's a summary of the changes:

 - Use a separate trigger, `VerificationR`, like NEP-7, to handle the case @RavenXce mentioned. We still invoke the script as specified in the original NEP-11 proposal.
 - Introduce temporary storage (by using `StateMachine` instead of `StateReader`) in consensus so that smart contracts can track used outputs across calls and even across transactions within the current block. This also has the side-benefit of solving the refund problem in the typical `mintTokens` `Verification` trigger implementations - simply by running `mintTokens` using `VerificationR`, the smart contract will be able to refuse all transactions that would mint past the ICO cap.

The above explanation might not be the best, so please use it in tandem with looking at the code to see how it's implemented.

canesin:
@RavenXce and @localhuman what you guys think about it ? Does it improve the situation ? What do you feel can still be better ?

RavenXce:
I'm actually not sure that temporary storage works. I don't think nodes can reach consensus like this as  the order verification runs depends on when nodes get the transactions to be put into mempool. Mempool is then sorted by gas payment amt etc, and txn execution is ran in that order.

Since nodes can receive txns at different times, then verification is no longer deterministic if writes are allowed, due to undefined ordering. @erikzhang or @shargon can probably correct me if I'm wrong here.

erikzhang:
I think @RavenXce is correct.

And as I commented at https://github.com/neo-project/neo/pull/276#discussion_r197683981, there may be a zero-cost attack.

dicarlo2:
> I'm actually not sure that temporary storage works. I don't think nodes can reach consensus like this as the order verification runs depends on when nodes get the transactions to be put into mempool. Mempool is then sorted by gas payment amt etc, and txn execution is ran in that order

Good catch. Actually, the way it works in consensus is different from the mempool (which we still run without temporary storage), but it's still not deterministically ordered. We could make it deterministic based on the order of the transactions in the block proposed by the leader. This may reduce TPS though in the worst case.

> And as I commented at neo-project/neo#276 (comment), there may be a zero-cost attack.

Another good catch. Perhaps during consensus instead of rejecting the transaction outright we still include it in the block but mark it as failed verification, so that it still costs the GAS but otherwise does not persist. Otherwise, we could change it to work similar to NEP-7 and only allow the free gas cost. But then, NEP-11 would likely depend on any economic model changes - while today every script would almost certainly stay within the gas limit, this may not be true in the future if the economic model changes.

igormcoelho:
I still don't know what is NEP all about... tried to understand the code directly, but I would appreciate if someone had a simple application smart contract or tx verification code example to demonstrate how this should actually work.
I also don't know what are the considered drawbacks of NEP-7.. is it the need of four triggers, while this proposal will have three (removing ApplicationR)? Is it to remove the strings "received" and "receiving" from NEP-7? (to be honest, I think they are redundant anyway given the possibility of exploring triggers...). Is it just to create something that is fully backwards compatible and avoid any kind of versioning in the blockchain (that could allow incompatible changes to be passed on)?
Really, if someone could elaborate on the exact reasons NEP7 is being rejected, it would help a lot to discuss a new one.

erikzhang:
There is no Global Asset in NEO 3.0 any more.

PR #55, opened 2018-07-07
NEP: Approval Voting with Time-locked Contracts

Edgegasm:


lerider:
I like the general idea that only NEO that has voted get the system fee pool.

I am not fully convinced that we want to prevent exchanges from voting, on two aspects.

1. Exchanges have the same dependency, if not more, on a high performing network. They should use the voting power to vote for high performing nodes. They will already be running high performance full node(s) themselves, so even voting "selfishly" could be beneficial for the network. They don't get more CN's without more nominees anyway.

2. Locked voting periods could be used to ease distribution of system fees if the lock time is short, for example 10 blocks, so it may be worth to consider for that reason. For longer voting periods, it may potentially create instability around the voting time; a lot of new votes right before the new period (new voters who want system fees), and a lot of withdrawn votes right after (old voters who want to liquidate). If the lock period is kept short for this reason, then it doesn't affect exchanges liquidation possiblities, and they can just lock their cold wallet NEO in either case.

Edgegasm:
Thanks for your thoughts about exchanges @lerider, you and @ThomasLobker have both made some fair points on the topic. I still believe exchanges may hold too much influence, but the advent of decentralized exchanges & the GAS incentives to vote will likely reduce that influence somewhat anyway.

As for time-locking, the consensus I've seen so far is that ideally we'd use a special vote transaction or in-wallet lock instead of a separate voting contracts as initially proposed, so that users aren't required to send funds away from their wallets. This seems like the better approach, it accomplishes the same thing but would provide a more comfortable experience for voters. Would this be feasible to implement?

I'm also curious to your thoughts on approval voting as our election mechanism. Satisfaction approval voting works similarly to the current implementation in neo-gui where a user can select any number of nodes, but in this approach their total vote power is split evenly across the selected candidates. It would be intuitive to use and easy for voters to be informed; they can just base their votes on the candidate datasheet on information such as node latency, transaction fees, trustworthiness (number of valid blocks proposed by the node) etc.

Edgegasm:
Proposal withdrawn due to better suited alternatives (e.g /neo-project/neo/issues/317)

PR #58, opened 2018-07-21
NEP-51: Token Standard Requiring Optional Methods

anfn101:
See issue: https://github.com/neo-project/proposals/issues/59

### DIFF
--- a/nep-51.mediawiki
+++ b/nep-51.mediawiki
@@ -1,6 +1,6 @@
 <pre>
   NEP: 51
-  Title: Token Standard
+  Title: Token Standard Requiring Optional Methods
   Author: Alan Fong <afong@cityofzion.io>
   Type: Standard
   Status: Draft
--- a/nep-51.mediawiki
+++ b/nep-51.mediawiki
@@ -127,7 +127,7 @@ Approves the <code>to</code> account to transfer <code>amount</code> tokens from
 public static event transfer(byte[] from, byte[] to, BigInteger amount)
 </pre>
 
-MUST trigger when tokens are transferred, including zero value transfers.
+The <code>transfer<c/ode> event is raised after a successful execution of the <code>transfer</code> method.
 
 A token contract which creates new tokens MUST trigger a <code>transfer</code> event with the <code>from</code> address set to <code>null</code> when tokens are created.
 
--- a/nep-51.mediawiki
+++ b/nep-51.mediawiki
@@ -25,66 +25,113 @@ All methods are required in this standard.
 
 ====totalSupply====
 
-* Syntax: <code>public static BigInteger totalSupply()</code>
+<pre>
+public static BigInteger totalSupply()
+</pre>
 
-* Remarks: "totalSupply" returns the total token supply deployed in the system.
+Returns the total token supply deployed in the system.
 
 ====name====
 
-* Syntax: <code>public static string name()</code>
+<pre>
+public static string name()
+</pre>
 
-* Remarks: "name" returns the token name.
+Returns the name of the token. e.g. <code>"MyToken"</code>.
 
+This method MUST always return the same value every time it is invoked.
 
 ====symbol====
 
-* Syntax: <code>public static string symbol()</code>
+<pre>
+public static string symbol()
+</pre>
 
-* Remarks: "symbol" returns the token symbol.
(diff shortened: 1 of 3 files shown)


nep-51.mediawiki
```diff
@@ -0,0 +1,93 @@
...
+* Remarks: "approve" will approve the '''to''' account to transfer '''amount''' tokens from the '''originator''' acount. 
+
+===Events===
+
+====transfer====
+
+* Syntax: <code>public static event Action<byte[], byte[], BigInteger> transfer</code>
+
+* Remarks: The "transfer" event is raised after a successful execution of the "transfer" method.
+
```
localhuman: Do we want an `approve` event?

anfn101: For previous discussion on optional events, we can refer to the discussion on the original NEP-5 here: https://github.com/neo-project/proposals/pull/4

In defense of approve, it allows people to transfer on behalf on someone, which people may find useful.

RavenXce:
This lacks the necessary details for the new (and also existing) methods, so we can't discuss anything. Better to base of the current NEP-5 doc https://github.com/neo-project/proposals/blob/master/nep-5.mediawiki.

anfn101:
Have updated to be more in line with current version.

canesin:
@RavenXce we have been recommending this implementation due to a needed change in the VM please contact me or @localhuman in discord or email, as you feel best, so we clarify the needed changes in the way you call some contracts.

We all understand that a new token standard that just reinstitute a old one is not ideal, but unfortunately a oversight let a new standard proposal at the same time change the at the time already finalized NEP5, now we have a vocal person spamming very damaging warns across social channels that will put projects following it at risk.

@anfn101 this proposal should mark NEP5 status as "replaced" so people don't fall in that mistake.

nep-51.mediawiki
mwherman2000: @totalvamp @afong @tyler @lllwvlvwlll @erikzhang  https://github.com/neo-project/proposals/pull/58 is non compliant with the current direction that has previously been established for NEPs. It should only contain the incremental merthods. Nothing more.  That is was agreed to.  #GangOfZion being destructive again.

@erikzhang, to you as the NEP Editor, I recommend against you accepting this NEP.

mwherman2000:
@totalvamp @afong @tyler @lllwvlvwlll @erikzhang #58 is non compliant with the current direction that has previously been established for NEPs. It should only contain the incremental merthods. Nothing more. That is was agreed to. #GangOfZion being destructive again.

@erikzhang, to you as the NEP Editor, I recommend against you accepting this NEP.

Edgegasm:
@erikzhang @mwherman2000 

Michael's opposition to #58 is based on a supposed community consensus for a new NEP direction & demand against retroactive changes. 

However, this _current direction_ had no community consensus; instead, Michael has justified it based on his own preferences that went largely undiscussed for 30 days. 

In addition, the only retroactive changes that have been made so far are Michael's [own handiwork](https://github.com/neo-project/proposals/commit/8c6b427e2ada2e71f92081686c4a832351e9eaa0#diff-e5a1452ea0de8fa24b9c328151e94aff), where he retroactively changed a finalized NEP and subsequently re-proposed it under his [own name](https://github.com/neo-project/proposals/pull/60/commits/e7331dc4c362c48d8e05675c033bcc38ce2809d6#diff-be542232e0c21a3a90e2065df599ca44).

I suggest this objection is ignored entirely, and NEP-51 is finalized to supersede NEP-5.

mwherman2000:
@edgegasm You can read through the community based discussion, acceptance, and finalization of NEP-10 here: https://github.com/neo-project/proposals/issues/40

Edgegasm:
On 12th May, you pointed out your concept of removing optional methods from NEPs going forward.

On 14th May, Erik agreed that there should not be optional methods; just new NEPs for separate methods.

On 17th May, you retroactively removed the existing methods from a finalized NEP. 

In what dimension is a conversation solely between you and Erik over the span of five days a 'community based discussion?' Do you see how _ludicrous_ that is from the perspective of someone that has been noting your own belligerence towards members of CoZ, for supposedly basing decisions on private discussions rather than community involvement? I find it to be wholly unacceptable and hypocritical.

canesin:
Just to publicly update I talked with @RavenXce in private to clarify. The intention was not to be disrespectful to any member of the broader community but unfortunately not every issue can be discussed in the open as it could damage projects.

saltyskip:
#61 or #60 seem to be a better solution as it does not break compatibility of existing projects, and allows us to maintain NEP-5 as the base token standard, with or without extensions

erikzhang:
> #61 or #60 seem to be a better solution as it does not break compatibility of existing projects, and allows us to maintain NEP-5 as the base token standard, with or without extensions

Agree.

Issue #59, opened 2018-07-21
NEP-51 (5.1), Token Standard Requiring Optional Methods


anfn101:
**Proposal**
----------------------------------------------------------
Name: NEP-51
Title: Token Standard Requiring Optional Methods
Author: afong@cityofzion.io
Status: Draft
Created: 21 July 2018
Resolution: https://github.com/neo-project/proposals/pull/58

Abstract
-----------------------------------------------------------
In the absence of NEP-5 reversion to Final state going through (issue #57), I propose NEP-51 (5.1), which requires the optional methods recently removed from the NEP-5 standard. 

See: https://github.com/neo-project/proposals/pull/58 for a draft of NEP-51, which is just the original NEP-5 with optional methods made required.

mwherman2000:
@erikzhang I've created https://github.com/neo-project/proposals/pull/60 to be reviewed and approved.

@birmas @anfn101 @lllwvlvwlll

birmas:
I don't understand why #60 has been submitted in favour of this issue.

Let's assume we all support nep10 to convey the functionality of our contract: This proposal would mean a single entry, in supported standards,  of Nep-51 whereas #60 would require nep5 and "nep-latest" to be declared.

It was suggested on discord that this nep would never be approved by the neo project admins because "that isn't the direction the community is heading" - I would like to see some details to back that statement if that is indeed that case

As someone about to deploy a nep5(1) contract to mainnet i completely support these additional methods.

mwherman2000:
@birmas Re: [neo-project/proposals] NEP-51 (5.1), Token Standard Requiring Optional Methods (#59)
RE: I also have little faith in the opinion of someone who seems to be using their nep contribution status for their own personal gain: https://www.google.com.au/search?q=site%3Agithub.com+mwherman2000+bounty+-"neo-project"&oq=site%3Agithub.com+mwherman2000+bounty+-"neo-project"

Chris, you really need to grow up.  This was a bounty contest sponsored by the Alchemint to which I applied my knowledge and experience.

You're now officially a member of the Gang of Zion.

PR #60, opened 2018-07-21
NEP-LATEST Token Exchange Basic Extensions

mwherman2000:
==Abstract==

The NEP-LATEST proposal defines three (3) additional methods designed to extend the base level token
functionality found in NEP-5 to support the needs of third-party agents such as exchanges.  Specialized exchanges
may require additional or different functionality.  They should create and submit additional NEP proposals that support their needs

### DIFF
--- a/README.mediawiki
+++ b/README.mediawiki
@@ -77,7 +77,7 @@ First review [[nep-1.mediawiki|NEP-1]]. Then clone the repository and add your N
 |-
 | [[nep-Exchanges.mediawiki|LATEST]]
 | Token Exchange Basic Extensions
-| Michael Herman
+| Tyler Adams, luodanwg, tanyuan, Alan Fong, Michael Herman
 | Standard
 | Draft
 |-
--- a/README.mediawiki
+++ b/README.mediawiki
@@ -76,7 +76,7 @@ First review [[nep-1.mediawiki|NEP-1]]. Then clone the repository and add your N
 | Final
 |-
 | [[nep-Exchanges.mediawiki|LATEST]]
-| Token Exchange Extensions
+| Token Exchange Basic Extensions
 | Michael Herman
 | Standard
 | Draft
--- a/README.mediawiki
+++ b/README.mediawiki
@@ -71,7 +71,7 @@ First review [[nep-1.mediawiki|NEP-1]]. Then clone the repository and add your N
 |-
 | [[nep-10.mediawiki|10]]
 | Composite Smart Contracts
-| Michael Herman, Joe Stewart
+| Michael Herman
 | Standard
 | Final
 |-
--- a/README.mediawiki
+++ b/README.mediawiki
@@ -75,6 +75,12 @@ First review [[nep-1.mediawiki|NEP-1]]. Then clone the repository and add your N
 | Standard
 | Final
 |-
+| [[nep-Exchanges.mediawiki|LATEST]]
+| Token Exchange Extensions
+| Michael Herman
+| Standard
+| Draft
+|-
 | 
 | Superconductive Exchange
 | 
--- a/nep-10.mediawiki
+++ b/nep-10.mediawiki
@@ -1,7 +1,7 @@
 <pre>
   NEP: 10
   Title: Composite Smart Contracts
-  Author: Michael Herman (mwherman@parallelspace.net) aka @mwherman2000, Joe Stewart (hal0x2328@splyse.tech) aka @hal0x2328
+  Author: Michael Herman (mwherman@parallelspace.net) aka @mwherman2000
   Type: Standard
   Status: Final
   Created: 2018-04-17
--- a/nep-Exchanges.mediawiki
+++ b/nep-Exchanges.mediawiki
@@ -1,0 +1,55 @@
+<pre>
+  NEP: LATEST
+  Title: Token Exchange Extensions
+  Author: Michael Herman (mwherman@parallelspace.net)
+  Type: Standard
+  Status: Draft
+  Created: 2018-07-21
+</pre>
+
+==Abstract==
+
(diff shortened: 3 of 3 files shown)


mwherman2000:
There's small issue with the formatting of a few characters like ", [ and ] ...maybe a character set issue.  I don't have time to look at this any more this morning.

mwherman2000:
@erikzhang  Can you assign a reviewer/approver?  Thank you

@anfn101 @lllwvlvwlll @birmas

erikzhang:
The extension can be accepted. However, it is likely to be discarded in NEO 3.0, as there are plans to add multiple APIs to implement this extension's functionalities.

birmas:
If that's the case then nep5 should have been left as is to begin with and the need for these temporary, useless, neps wouldn't be needed.

My mainnet contract will not support "nep-latest" it will continue to support the previously "final" version of nep5

erikzhang:
@birmas For contracts, what's the difference?

birmas:
@erikzhang the difference is the manner this NEP-LATEST is proposed. 

* By retrospectively modifying the "final" NEP5 which projects have used and deployed to mainnet. 
* By closing down this perfectly acceptable (and complete!) proposal for NEP-51? https://github.com/neo-project/proposals/issues/59

When I said I will not support NEP-LATEST in my contract I mean: the response I give for NEP-10 supportedStandards() will not include NEP-LATEST. It will continue to return NEP-5 and NEP-10.

erikzhang:
The proposals #58 and #59 want to create a new token standard that includes NEP-5 and optional methods. Once this standard is used, the NEP-5 will be discarded.
But in https://github.com/neo-project/proposals/pull/61#issuecomment-407069002 we have learned that a large number of old contracts have already supported the NEP-5 standard, so it cannot be discarded.
This proposal create an extension to NEP-5. This will preserve the compatibility of the old contracts. And when the NEO 3.0 is online, we can easily discard this proposal, leaving only the NEP-5 part.
So it has nothing to do with manner, I only choose from a technical point of view.

birmas:
That makes sense but then why not just leave nep5 as it was in its final version?

This whole situation is because the"final' standard was retrospectively modified?

birmas:
"The extension can be accepted. However, it is likely to be discarded in NEO 3.0, as there are plans to add multiple APIs to implement this extension's functionalities."

As you said, it seems like this whole process was something that's going to result in it being discarded anyway.

The amount of time and the division this "change process" has caused in the community is shocking.

decentralisedkev:
@erikzhang If we added a new NEP that was seen as an improvement on NEP5 and a full token standard, then we can still support the older contracts, while telling new contracts to implement the newer standard instead.

I don't think we should have went back and edited NEP5, if NEP10 was incompatible with optional methods, it would have been better to create a new token standard that was compatible with it or not implement NEP10.

ThomasLobker:
These methods are not limited to exchanges and they can in fact be very useful in many cases (handling in-dApp payments for example). I would recommend changing the title to _Allowance Extensions_. As mentioned [in my comment here](https://github.com/neo-project/proposals/pull/61#issuecomment-407069002) a couple of NEP-5 tokens are already compatible with these extensions, so in my opinion this is be a valuable proposal and it should be accepted.

mwherman2000:
RE: I would recommend changing the title to Allowance Extensions

I, for one, am not stuck on the draft name.  However, I don't think "Allowance Extensions" is descriptive enough.  What is the precise, *collective* purpose of *allowance*, *transferFrom*, and *approve* methods? [I've deliberate phrasing this in a naive tone.]

How about something like "Token Extension: Third Party Transfers"? ...also think about this as a naming pattern for similar future extensions e.g. NFT, requisitions/remittance processing.

mwherman2000:
RE: "The extension can be accepted. However, it is likely to be discarded in NEO 3.0, as there are plans to add multiple APIs to implement this extension's functionalities."

@erikzhang There has been no (open, public) discussion about NEP-5 being discarded in the future.  Can you explain? ...perhaps in a new issue/discussion if necessary.

mwherman2000:
NEP-10 Documentation
```csharp
using Neo.SmartContract.Framework;
using Neo.SmartContract.Framework.Services.Neo;
using System;
using System.Numerics;

namespace NEO.NEP.Examples
{
    public class TokenExtensionThirdPartyTransfers : SmartContract
    {
        public static object Main(string operation, params object[] args)
        {
            object result = false;

            if (operation == "supportedStandards")
            {
                result = (object)supportedStandards();
            }
            else if (...)
            {
                ...
            }
            ...

            return result;
        }

        public static string[] supportedStandards()
        {
            string[] result = { "NEP-5", "NEP-10", "NEP-LATEST" };

            return result;
        }
    }
}
```

RavenXce:
At the moment this is just a skeleton?

Details of exactly how the operations should work especially with regards to the unhappy paths / contentious cases (0 transfer amount, etc.), non-common cases (invoked from another contract, etc.) should be in the spec, lest the same mistakes that were made in the original NEP-5 are made again.

Also, these threads are important for user UX and programmer safety. The implementation we want to use + the caveats we're willing to accept should be clear:
https://github.com/ethereum/EIPs/issues/738
https://github.com/CORIONplatform/solidity/issues/107

mwherman2000:
@erikzhang Is it possible to move this PR forward - at least to the point where its given a number?  
If not, what is holding this up (besides the recent events on the NEO blockchain)?

erikzhang:
> Details of exactly how the operations should work especially with regards to the unhappy paths / contentious cases (0 transfer amount, etc.), non-common cases (invoked from another contract, etc.) should be in the spec, lest the same mistakes that were made in the original NEP-5 are made again.

I think that this proposal must be clearly stated before it is accepted.

PR #65, opened 2018-08-20
NEP: NEO Name Service Standard

PhyrexTsai:
This NEP presents standard with a set of contracts and methods that make NEO Name Service possible.
NEO Name Service can benefit with wallets and clients by human-readable identity instead of hexadecimal number.

### DIFF
--- a/nep-nns.mediawiki
+++ b/nep-nns.mediawiki
@@ -30,24 +30,24 @@ The NNS system has thsee main parts:
 * Registry Smart Contract
 
 Transfers ownership of a node to a new address. May only be called by the current owner of the node.
-```csharp
+<code>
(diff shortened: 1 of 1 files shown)


erikzhang:
Why do we need a standard, not a separate service?

PhyrexTsai:
For extensibility, NNS designed with a standard that means every service build top on the NNS should follow the rule of this spec

saltyskip:
I too believe that this should be a separate service and not a standard. It is up to the designer of the name service on how it should be structured. This puts an unnecessary restriction on any service that is developer.

This proposal should be closed

erikzhang:
It should be a separate service and not a standard.

PR #66, opened 2018-09-03
Staking for low-cost (network) costs

igormcoelho:
This is a draft o the proposal recently discussed here (https://github.com/neo-project/neo/issues/357). I believe it is feasible to implement in really short time, as it could fit together in the SimplePolicyPlugin. Please, be welcome to contribute, I hope more (many more) names go together with us signing this NEP, and that could somehow help Neo becoming an even better, low-cost and business-friendly ecosystem.

[EDIT] - If Smart Transaction identification proposal (https://github.com/neo-project/proposals/pull/67) is approved, then I suggest this one is not created as a NEP, but simply as a feature that uses its Smart Transaction capabilities. This will allow a much more smooth implementation process for staking.

### DIFF
--- a/nep-staking.mediawiki
+++ b/nep-staking.mediawiki
@@ -1,0 +1,99 @@
+<pre>
+  NEP: XXX
+  Title: Staked low-cost tx
+  Author: Igor M. Coelho <igor.machado@gmail.com>, Vitor N. Coelho <vncoelho@gmail.com>, et al (I hope so!)
+  Type: Draft
+  Created: 2018-8-25
+</pre>
+
+==Abstract==
+
+This NEP proposes a standard for low-cost tx on high-staked accounts.
+Each account with a minimum value (on governance token) will be able to emit a limited number of "tickets" that guarantee access to low-cost network fees.
+If misbehavior (such as network attacks) is detected by automated filters using staked tickets, the accounts will be blocked for a given period of time.
+
+==Motivation==
+
+Pricing is a fundamental tool to control access on a network and it is deemed necessary to avoid network attacks (CITATION BITCOIN?).
+Neo technology organizes fees in two categories: network and system. Network fees intend to reward consensus nodes, charging (or not) every operation on the network.
+System fees control the nature of the operations executed on the smart contracts, guiding the development of efficient software (CITE NEO White Paper).
+However, it is hard to decide "optimal" prices when several actors are involved in a blockchain ecosystem, so the discussions on price adjustments tend to be quite challenging and slow to react to reality changes.
+Usually, a single actor (and price) is considered, but this NEP intends to create an alternative pricing structure for an important actor on the network: big business and exchanges.
+Big business (and exchanges) that bet on Neo technology should be rewarded and kept safe from sudden general price adjustments.
+This NEP proposes two different network prices: a general price and a staked low-cost price, which is accessible only by account posessing high quantities of governance token (Neo).
+If abuse is detected by automated filters (developed by community), the account may be block for a given period of time, which will highly desencourage misbehavior from high-staking actors, and even extending these benefits to trusted customers.
+
+==Specification==
+
+This NEP is fully backwards compatible and is implemented based on UTXO attributes and control parameters.
+
+===Staking===
+
+If this NEP is adopted, a staking transaction can be made in the following way.
+- (1) Transaction is made to an output account holding at least `minimumStakingValue` of governance tokens (Neo)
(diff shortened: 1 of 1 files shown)


igormcoelho:
@vncoelho time to invite everyone to help us discussing this :)

igormcoelho:
@PeterLinX @erikzhang @canesin @localhuman @shargon @lerider your opinion will also be very valuable here!

belane:
@RavenXce I understand your point, let's agree between all the most appropriate values, that help protect the network and guarantee the use of legacy contracts.

igormcoelho:
Ok @RavenXce , that's very nice to hear from you! Recently in a discussion with @erikzhang  (https://github.com/neo-project/neo/issues/371), he made me see that there are more features we could explore from Neo, to avoid putting unnecessary extra logic. So, using this new knowledge, I think I can reformulate the current proposal in a much simpler one (to implement), which is also easier to validate on relay/consensus nodes.

I think we can use this opportunity to create a slightly more general NEP for "Smart Transactions", which are transactions that perform more elaborate logic. This will create a bigger umbrella to guarantee that similar situations will be dealt much faster! I will start another template for Smart Transactions and reference this one (on how it could be implemented there), because I don't know yet how many people will agree on this new and slightly broader format.

igormcoelho:
@RavenXce this is the proposal for an identification system on Smart Transactions: https://github.com/neo-project/proposals/pull/67
I plan to use this simple header system to build two operation codes: Stake Propose (`STP`) and Stake Consume (`STC`). So, we can detect these symbols easily on P2P nodes and verify if the according operation is correct.

igormcoelho:
Just an example for a DEX that needs to claim/withdraw payments. This DEX uses a staked Neo to backup free transactions and creates one (or many) private/publickeys for staking referal. This DEX can create a public website service that automatically signs transactions using their private key, if transactions seem legit and not abusive for the network. So, the user simply enters this signature in his wallet during transaction process (we will obviously need to coordinate these efforts with Wallet developing teams), but this whole process could be virtually 100% smooth for users and DEX in the end.

igormcoelho:
Ok @RavenXce , if the owner of the stake uses a single-signature contract (which is unlikely for bigger fortunes), it could reuse the private key to sign staked transactions in a Script field. That's possible, true, nothing new is inserted (except the part that verifies the staking on relay nodes plugin). 
But let me emphasize that Remarks already exist on Neo, and since the network is open to any kinds of "marks", I don't see an issue in using Remark field to "mark" things that are useful to the whole process (for example, people write "airdrop", but also write random stuff no one knows). They will still be able to do that, but IF they write "NEP-11 AIR" or "NEP-11 STP", it's very clear what they mean. In the case of "STP", the advantage is to propose other public keys that will only be used for the purpose of stake validation (even if contract is single-sig, multi-sig, timelock, ... doesn't matter). And if some privatekey is lost in the staking process, no funds will be stolen, because it's trivial to replace them (just re-issue another self-transaction). And to mention that NEP-11 (or any other number) won't change a single line on Neo, it's just a common standard to make it easier to detect staking and other different situations, with much less computational efforts on relay nodes. I will create a working example on neocompiler.io, and even if this NEP-1X does not go through, staking will work anyway as long as relay nodes know what to look for on transactions.

igormcoelho:
@shargon, do you think extra 290 bytes to every staked free tx will be damaging to the network? this way it only uses native Neo marks... no NEP is needed.

igormcoelho:
@RavenXce @shargon sorry for the many messages, I swear I'm trying to make this as simple as possible... on general, I think we have two options:
(1) put all validation logic attached to a Script on every transaction, which will be at least ~300 bytes on every transaction, which is applying for a free execution
(2) allow the validation plugin to verify the staking conditions, except pubkey signature, which will be passed as a Script. The invocation + verification Script will have 76 bytes + remark info, around ~120 bytes. This will require a NEP, I think, at least to minimally organize what we are doing... but we can do without any NEP and just assume this is how staking works.

vncoelho:
@igormcoelho, I think that the proposal is very well written.

In the beginning I thought about removing the filter, but later I understood that you mentioned filters in terms of someone attaching invalid UTXOs, which would waste Consensus computational time.

Congratulations for the proposal.

erikzhang:
@igormcoelho There is no UTXO in NEO 3.0 now. Can I close this proposal?

vncoelho:
@erikzhang and @igormcoelho, should we close the proposal or adjust it for the context of Native NEO NEP5 Contract?

I still believe that staking could be another way for giving NEO holders a way to send "free" transactions as a matter of trust.
On the other hand, GAS distributions should also designed in a way that the those holding NEO would intrinsically have such amounts of GAS.

igormcoelho:
Not need anymore,self claim of gas can do that automatically.

PR #67, opened 2018-09-22
Proposal for Smart Transaction identification system

igormcoelho:
This NEP intends to propose a 10-byte header with human-readable characters, embedable in general Remark fields, that can be used to easily understand the nature of Smart Transactions on Neo network. This NEP is fully backwards compatible and is only used for identification purposes (no behavior will change on the network). This can also be used to accelerate the detection of specific Smart Transaction and to implement reactive behavior based on the content of such transactions.

### DIFF
--- a/nep-smarttx.mediawiki
+++ b/nep-smarttx.mediawiki
@@ -1,5 +1,5 @@
 <pre>
-  NEP: XXX
+  NEP: 11 (??)
   Title: Smart Transactions
   Author: Igor M. Coelho <igor.machado@gmail.com>, Vitor N. Coelho <vncoelho@gmail.com>, et al (I hope so!)
   Type: Draft
--- a/nep-smarttx.mediawiki
+++ b/nep-smarttx.mediawiki
@@ -1,0 +1,77 @@
+<pre>
+  NEP: XXX
+  Title: Smart Transactions
+  Author: Igor M. Coelho <igor.machado@gmail.com>, Vitor N. Coelho <vncoelho@gmail.com>, et al (I hope so!)
+  Type: Draft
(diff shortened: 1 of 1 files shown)


igormcoelho:
@erikzhang @PeterLinX please take a look at this NEP when you can.  The idea is create a simple 10-byte mark that can be used on any Remark field, to be used in different purposes... the first one will be the staking idea, but it can help build so many different stuff.
This is fully backwards compatible, but it can help to build deeper stuff that would be more complicated without it.

igormcoelho:
@shargon @belane @RavenXce @jsolman

shargon:
I don't see the advantage, if this flag is set based on rules, should do it who really need it, the explorers, etc etc

The people lie, so this flags are not trusted

igormcoelho:
Good to see you here @shargon.. perhaps it's not needed, true, but the purpose is not to avoid people to lie. It's just a simple standard to allow people to explain what the transaction is about. In the case of Staking, for example, it would be useful for the plugin to know that a "STP" operation code is there, much better than scanning the whole transaction to find the exact pattern of the Stake redeem script. It's also a place to put extra information, in the case of Staking, a list of public keys that are allowed to use the staked funds to backup transactions. If people lie, that's fine, the idea is just to give a hint on how to proceed with Stake redeem, it will just be an invalid operation, just like any random Remark.  Perhaps it's an overcomplication, but I really think it helps structuring something bigger. In my opinion, structured information is interesting, even if people lie about them.
But like I said, everything can be build without this, I just think it's more complicated and less organized.

igormcoelho:
I withdraw this proposal.

PR #70, opened 2018-10-09
Mintable/Burnable NEP-5 Token

zlumer:
A NEP describing mintable/burnable NEP-5 compatible token.

Key points for discussion:
1. Should zero value mints/burns fire the corresponding event?
2. Should mint method be part of this specification or it's implementation can be left up to developer?

### DIFF
--- a/nep-mintable-burnable-nep-5-token.mediawiki
+++ b/nep-mintable-burnable-nep-5-token.mediawiki
@@ -1,0 +1,137 @@
+<pre>
+  NEP: <to be assigned>
+  Title: Mintable/Burnable NEP-5 Token
+  Author: Dmitry Radkovskiy <zlumer@gmail.com>, Ivan Kocheshev <ivan.kocheshev@gmail.com>
+  Type: Meta
+  Status: Draft
+  Created: 2018-10-01
+</pre>
+
+==Abstract==
+
+This NEP describes a smart contract interface for mintable and burnable NEP-5 tokens.  
+
+==Motivation==
+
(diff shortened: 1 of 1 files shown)


erikzhang:
There is a description of the `transfer` event in [NEP-5](https://github.com/neo-project/proposals/blob/master/nep-5.mediawiki#transfer-1):

> A token contract which creates new tokens MUST trigger a transfer event with the from address set to null when tokens are created.
A token contract which burns tokens MUST trigger a transfer event with the to address set to null when tokens are burned.

So is it necessary to have `minted` and `burned`?

zlumer:
Since this NEP assumes that we fully conform to NEP-5 spec, I guess we can omit the `minted` and `burned` events. In that case token mints can be found in tx list as `transfer` events from `null` address, and token burns are `transfer` events to `null` address in accordance with NEP-5. The wallets clients will have enough info to display them as mints/burns.

saltyskip:
Minting and burning can have a much wider application compared to just an initial issuing. It also enables a coin with dynamic supply. This can be applied in things like loyalty programs and tether assets

Here is an example of such a contract that makes use of minting and burning to create a dynamic supply token

https://github.com/O3Labs/Blockchain-Loyalty-Program

Given the scope of minting and burning, I believe that these should be seperated into distinct events, if this standard were to move forward

corck:
Some tokens, like the one I work on, have not hardcap, so hardcap() describe what is returned if there is not hardcap. Could be `-1`

PR #71, opened 2018-10-09
Multi-Token Contract

zlumer:
This NEP describes a smart contract interface for NEP-5-like multi-token contract.

This new standard is meant to serve as an entry point for developers that need to deploy a new token for ICO or other purposes, but may not be ready yet to pay the corresponding GAS price.

### DIFF
--- a/multi-token-contract.mediawiki
+++ b/multi-token-contract.mediawiki
@@ -1,0 +1,201 @@
+<pre>
+  NEP: 
+  Title: Multi-Token Contract
+  Author: Dmitry Radkovskiy <zlumer@gmail.com>, Ivan Kocheshev <ivan.kocheshev@gmail.com>
+  Type: Meta
+  Status: Draft
+  Created: 2018-10-01
+</pre>
+
+==Abstract==
+
+This NEP describes a smart contract interface for NEP-5-like multi-token contract.
+
+==Motivation==
+
+Currently, there exists a NEP-5 token standard for developers to implement tokens that can be easily integrated to wallets and exchanges.
+
+Similar token standards exist on other blockchains (e.g. [https://github.com/ethereum/EIPs/blob/master/EIPS/eip-20.md ERC20]).
+However, there is an important difference between ERC20 and NEP-5 usage on main networks.
+While on Ethereum a simple ERC20 token contract deployment costs around 750K gas or $1.3 (at the time of writing, using average gas price), on NEO a simple NEP-5 contract deployment will cost 500 GAS or $2900. These prices only go up with GAS price and we've already seen $15000 price tag for a simple NEP-5 token this spring.
+
+While this GAS is used to pay all NEO holders, it makes the barrier for entry too high for many developers that are just seeking funding through ICOs and are not yet funded with initial capital.
+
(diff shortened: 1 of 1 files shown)


zlumer:
This NEP has some similartities with #41 but both the intention and implementation differ enough to justify a separate standard.

saltyskip:
Is this standard necessary if GAS fees are reduced for issuing contracts, it seems like the long term plan is to reduce these fees substantially?

As seen in discussion of this issue

https://github.com/neo-project/neo/issues/365#issuecomment-428528366

canesin:
I dislike this proposal, it is a work around to a issue - how to properly price use of storage. It also puts many tokens under a single contract owner (I know owner permissions can be worked out - but why the complexity ?). Besides that it propagates a standard that has issues today, NEP5 still needs to see the adoption of the previously optional methods that many contracts today live have broken implementations of.

deanpress:
+1 @canesin and @shargon. This is a dApp, not suitable for a standard.

zlumer:
@saltyskip in case GAS fees are reduced substantially, we provided a way to deprecate existing token on that type of contract and migrate to a separate NEP-5. The key point is to enable developers to issue tokens right now, without having to wait for the GAS price to get lower.

zlumer:
@canesin I agree that this proposal may be seen as a mere workaround, but we intended it to be more that that: not only there's a huge need for less expensive deployment of new tokens on NEO, there's also significantly less work required for token deployment using this standard: token owner only needs to call a contract once with required parameters.
Not sure about complexity, please elaborate.
This NEP is fully open for improvements, including improvements based on concerns about NEP-5, we definitely need to have this discussion before proceeding.

zlumer:
Overall this is a great improvement for NEO, we obviously need this change and we need it right now.
ETH has [over `8 000 000` smart contracts](https://blockchair.com/ethereum/calls?q=type(create),failed(false)#) while NEO has [only around `100`](https://neotracker.io/browse/contract/1). If we actually want to participate in competition with other blockchains, then we need to seriously step up the game. Our proposal is a good start for this. I'd be happy to continue this conversation and improve this proposal.

shargon:
@zlumer My goal is not have 8m SmartContract in 2018, this have no sense for me. I prefer quality instead of quantity.

shargon:
To be clear. This is not a proposal for reduce the neo fee as you said, or as you try to show in your Twitter. This is a proposal to allow you with the same fee to resell tokens to third parties, probably on your platform, but you don't intends to eliminate or lower the deploy fee. if you really care about this, go here https://github.com/neo-project/neo/issues/365

WyattMufson:
I don't see the point of this. It is not a solution to the problem of high GAS fees, which is being discussed elsewhere as @shargon mentioned.

jeroenptrs:
@zlumer you'd be spamming the chain with temporary info, forever stored in the blocks. This is in no circumstance a feasible solution. There's been discussions going on that my colleagues above already linked multiple times.

I'm sorry, but if you're really want `to continue this conversation`, you shouldn't do it in this NEP but in the appropriate discussion, instead of diluting existing conversations. This has no point and makes no sense.

erikzhang:
I think this can be an application of https://github.com/neo-project/proposals/pull/41.

PR #73, opened 2018-11-04
NEP: Node Monitor interface standard

ixje:
To quote the abstract
> A monitoring standard and interface enables extensive data collection capabilities on nodes in a predictable manner. Such capabilities and measures are essential in order to monitor node performance as well for enabling insights into the overal network for current and potential users. This proposal describes an architectural solution for enabling monitoring data collection as well as an expanded set of attributes that can be collected.

For more motivation why we believe this is valuable see the **Motivation** section of the standard.


ixje:
Peter Lin, Tom Zhang, insomnia, wingyungchan; please let me now how you want to be described in the authors section. I want credit you all for the contributions made (interface discussion, example code, attribute ideas, presenting the initial idea to Erik Zhang etc). Thanks!

saltyskip:
Some info about fees could be useful as well. E.g average fee, total fee, median fee. Would be useful to determine what fee is reasonable to submit.

Nice proposal!

ixje:
@saltyskip thanks!

Regarding fees; 
good idea. I think there are 2 possible sources
1. collect them from processed blocks 
2. return the fees attached to TX's in the mempool (e.g. aggregated as `{ fee x: tx count, fee y: tx count}`) and then a stats website can collect these. I'm assuming this idea is inspired by [the bitcoin equivalent](https://jochen-hoenicke.de/queue/#0,24h)

The first can collect historical average/total/median and doesn't need anything additional in this proposal. The second is valuable if you want to prioritise your TX on the network when it's busy which does need an addition to this proposal. Is this inline with what you have in mind?

nep-monitor.mediawiki
```diff
@@ -0,0 +1,171 @@
...
+| uint32
+| Time in seconds from Unix EPOCH since the node started running without interruption.
+|}
+
+==Rationale==
+
+===Monitor interface===
+'''Shared vs separate interface'''
+
+By establishing a separate channel on the regular P2P/WebSocket port, as opposed to creating the service on its own port, we avoid losing the ability to collect statistics whenever a node is actively participating in the network.
```
phetter: Splitting channels by interface or port works well for monitoring traffic using classical network analysis tools.

Each node will naturally be capable of collecting stats, but it isn't clear that these should be managed and delivered on the same node that is computing chain. The data could be offloaded to a monitoring node that is essentially a monitoring data mirror that only computes monitoring client requests and delivery. Each node would have a monitor mirror. This adds more nodes to the network, but helps with scalability and lightens node overhead.

phetter: Changes to existing "chain" node code to facilitate the mirror-node monitoring approach could be as simple as additional logging calls. The mirror could just consume logs using standard log management tools, for example. Optimal performance must be observed during selection.

phetter: In the case of persistent stat collection, database integration is heavier performance-wise as the chain node has to handle queries, but may ultimately be a requirement for the types of data being requested. It does lend some consideration to the notion of chain node optimization via off-box database coupling via a logging mechanism,  just spitballing here.

nep-monitor.mediawiki
```diff
@@ -0,0 +1,171 @@
...
+
+Data attribute broadcasting is to occur after every block persist. In the [https://github.com/neo-project/neo current official C# client] this means on every <code>Blockchain.PersistCompleted</code> event. A visual representation of the idea can be seen below
```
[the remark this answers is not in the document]
ixje: Thanks for the input!

regarding _stream mode_. I don't think that stream mode is ideal in this proposal. As we all know, blocks do not get generated in exact equal intervals, while the majority of the current default attributes only change per block. So then it becomes a trade-off; 
- fixed interval with a majority data duplication
- variable/triggered interval with no/minimal data duplication

regarding _push mode_,  what interval is the data pushed with?

takeaways:
2. that might be an implementation caveat worth noting. 
3. I think we need to do some profiling on this to see if pushing "more bytes per client" (=no request profile) performs better/worse than "custom data per client" (=with request profile). So we need to look at performance of 
- collect data + craft 1 `Message` then distribute * _n_ peers
- collect data + craft 1 `Message` * _m_ unique profiles , then distribute to peers
I have a feeling that unique profiles will be slower because there's logic that needs to filter data per profile, each profile `Message` requires a new double SHA256 checksum, and there's logic for distributing the specific message to the specific client. I'm still open for it, hoping to get input from others as well!

nep-monitor.mediawiki
ixje: We're aware of an easier initial implementation when the interface is on a separate port, but at the same time it means more conscious work for the node hoster to eventually set it up. Obviously I'm biased here, but I rather have a higher upfront implementation cost and then later have a low barrier for anyone hosting a node to share their data, than vice versa. 

I don't see much benefit of a monitoring node, unless you support custom request profiles. The full data set still has to be collected on the _chain node_ and send to this monitoring node. While we can't foresee it, I don't think we'll see such a huge number of data collection clients that off-loading becomes necessary. If that eventually does become the case, then you can always add that later by adding a client filter on the monitor channel that only accepts the off loading node and the rest can make use of that. 

Logging calls are heavy and disk I/O bound, which I believe will bring down the node performance as now the disk is shared with LevelDB access for committing blocks. The current proposal allows for all data to stay in memory (the static parts like `Description` and `Agent` can be cached, the rest only needs a 1 time snapshot that gets overwritten each round)

> In the case of persistent stat collection, database integration is heavier performance-wise as the chain node has to handle queries, but may ultimately be a requirement for the types of data being requested.
Can you give an example of an attribute that requires this? We tried to write it as such that no historical data has to be collected for longer than the block interval. This is because we believe it is up to statistic websites (i.e. HappyNodes) to track this.

jseagrave21:
@ixje BLUF: I think real-time and historic data streams are necessary for a clear statsical perspective.

After reviewing the discussion I am thinking there could be perhaps two streams a client could connect to (one at a time).

**Real-time:** One would push each tx added to the mempool which could be delivered in json format from which [unconfirmed] TPS could be calculated, average fees in use, average size of each tx, etc. See https://www.blockchain.com/btc/unconfirmed-transactions for my inspiration.

**Historic:** The other would push the json data mentioned in the PR calculated after each new block. This would not be a constant rate but could be plotted using the block's timestamp, allowing a similar graph to the aforementioned "Bitcoin equivalent"

This would increase the CPU load considerably but I don't think it would be too burdensome.

ixje:
The mempool thing looks cool, but I personally try to focus at splitting responsibility (pushing data vs tracking) and keeping node load as low as possible. We have to keep the 100K TPS end goal in mind and node load plays a role in that. It still an idea that could perhaps be implemented at a later stage via new NEP to extend the current proposal once we learn more about the impact and usage of the interface.

aboimpinto:
In the begin of November 2018 I made this presentation to CoZ 
[Neo statistics and Analytics](https://gist.github.com/aboimpinto/b6aa443e202f22506afb2082f8c5d8a6)

and I didn't know on the same time there was a discussion going on to add some statistics and analytics capability to the NEO node software.

I would be interested to know more about this need in order to be able to align my idea to be able to be integrated as a plugin to the node.

erikzhang:
@shargon Do you think it's necessary to implement this?

shargon:
Improve the monitor and log verbosity always is good for find problems

erikzhang:
Do we need a standard? Can we use the current p2p protocol to build monitor node?

shargon:
It could be a plugin, we should add events in a lot of sites for help plugins like this to get more information.

aboimpinto:
The standard plugin interface should be enough. 
There should be able to hook a plugin that is called every time a new block is generated. The information that can be provide to the plugin is: Block and MemPool information

With this block information it's possible to create the extra log information, store it somewhere else, create statistics / analytics from it, etc..etc...
Same with the Mempool.

With this implementation, the node it self don't need to care about the log information, statistics and analytics and any person can develop his own plugin using this information.

vncoelho:
Nice, a draft of mempool stats was done here: https://github.com/neo-project/neo/pull/564

This could be merged in this plugin here if you want: https://github.com/neo-project/neo-plugins/tree/master/CoreMetrics

ixje:
A plugin would defeat one of the reasons I proposed a standard. From the abstract above

> Such capabilities and measures are essential in order to monitor node performance as well for enabling insights into the **overall network** for current and potential users. 

A plugin can only provide insights for the node owner. Global network insights are necessary for among others:

- Developers & investors assessing network suitability for their business case
- Developers looking to improve network performance
- Token holders looking to make an educated vote for consensus node candidates
- (Future) Token holders to build or maintain trust in the network
- Node hosters looking for hosting requirements and maintenance tools

By making it part of the P2P protocol we increase the likelihood that other clients will add it as well. It also lowers the risk that `neo-cli` forget to enable/use the plugin which would mean losing valuable insights.

vncoelho:
I agree, @ixje.
In a multi-agent system perspective, different autonomous agents are playing different roles on the network.
Trust in metrics from a P2P protocol is not a trivial design. However, I believe in the benefits of it.

The first solution is to enable those interested in monitoring the network (or reporting their behavior) to use it through a plugin.

nep-monitor.mediawiki
```diff
@@ -0,0 +1,171 @@
...
+The monitor interface is the access point where collection clients can access the data attributes. When a client connects to a node over the P2P or Websocket port a handshake protocol is started. First, the node sends a [http://docs.neo.org/en-us/network/network-protocol.html#network-message network message] with a <code>version</code> command and the appropriate payload. In the original handshake implementation the client is expected to respond with a similar <code>version</code> command network message and its own payload or be disconnected from the node. According to this proposal, we allow the client to provide an alternative response in the form of a network message holding the <code>monitor</code> command type and a zero length payload to indicate that the client is not a regular peer but a client that wishes to collect data attributes. The node separates the connection into its own monitoring channel on which no [http://docs.neo.org/en-us/network/network-protocol.html#network-message regular network messages] from the client are accepted and only data from the node is broadcasted. The <code>monitor</code> payload format is described in the next section.
```
vncoelho: Maybe hidden agents monitoring the network could be enough. Plugins developed for different purposes could handle this task.

ixje:
> The first solution is to enable those interested in monitoring the network (or reporting their behavior) to use it through a plugin.

I don't think you can monitor anything with a plugin, only report/share data. An important item to keep in mind is to make it becomes a "push data" design, not pull data (like RPC) as that's going to kill performance .

vncoelho:
Plugin or not is just a matter of code organization.

By means of a Plugin I mean that the core code can be modified in the way that those interested in reporting/collecting the data can do it regarding their interests.

Maybe this standard is a good idea, @ixje. But I am not sure if it is really needed. Maybe adding complexity where it is not needed.
Let me insight it a little bit more.

ixje:
I don't understand the complexity you're referring to. We're just adding 1 extra branch in the handshake to take us to the monitoring channel if the command is `monitor` and from there the node is the only one that can broadcast and doesn't process/accept any messages from the client on that channel.
image

vncoelho:
Thanks for taking time doing this figure, @ixje. Did you draft it now?
I got the idea, but it requires the one being monitored to agree with such periodically reporting, right?

ixje:
It's actually already in the standard, but I had a feeling not everybody opens that in the pretty format (most probably just look at the `Files changed` tab which doesn't render the image).

As the standard is proposed now; the node broadcasts its statistics to the connected clients on each block commit event.

vncoelho:
I think that @erikzhang and @jsolman will have a good vision about it.
As I mentioned, maybe it is not needed now. Let's discuss it further.

EdgeDLT:
Plugins might make it easy to add any necessary monitoring functions, but doesn't that make it quite 'opt-in?' Consensus nodes at least should be open to some kind of monitoring procedure, and this information should be easy to retrieve for users.

I can't speak on the implementation but I think this is very important, fundamentally, for decentralization.

ixje:
That depends on who's perspective we're looking from. Yes it is not needed now for e.g. consensus voters as that's not implemented yet or not used yet for selecting consensus nodes. But it is too late to implement the feature by the time they need to start voting for consensus nodes, because then there's insufficient time to collect sufficient data for making an educated vote. Therefore this needs to planned ahead of time, not last minute.

vncoelho:
@EdgeDLT, we currently adjusted the code for watch-only node on the P2P of consensus (https://github.com/neo-project/neo/commit/7e82bcad2d95c223bc255710bd7676e85d26168b).
I think that a wide spectrum of metrics can already be covered with the features we already have.

We need a community to report that periodically. In the DevCon I heard something about Swisscom and NGD.

ixje:
> Consensus nodes at least should be open to some kind of monitoring procedure, and this information should be easy to retrieve for users.

I agree. Currently a third party service would not be able to collect the necessary data and make it easy accessible/understandable for the various skill levels of active or potential NEO network users. That's one of the topics this NEP could address.

vncoelho:
An example of logs of this watch-only can be seen here: https://neocompiler.io/#!/ecolab/cnnodesinfo

image
Specialized tools can currently tease out this data and report it.

erikzhang:
What if a node has too many monitors connected?

ixje:
> What if a node has too many monitors connected?

It can be restricted just like the regular P2P protocol. For example there can be a restriction setting for regular P2P and one for the monitor channel (as in; both have their own pool). The currently proposal would emit ~95 bytes of data per client per broadcast. That's far less than the average data emitted in 15 seconds over the normal P2P protocol. Another idea is to white list accepted third parties e.g. a neo-scan like service but for monitoring data. This are implementation details that can be discussed of course.

vncoelho:
My worry was also the same question @erikzhang asked.

That is why I mentioned a little bit about complexity.
I did not like these white list solutions, not scalable from a MAS perspective.

The amount of data sounds good, @ixje.
We should also consider that monitoring summarized data can be subjected to manipulation, while hidden agents monitoring sounds to me like a more promising way of checking the network.

ixje:
> We should also consider that monitoring summarized data can be subjected to manipulation, while hidden agents monitoring sounds to me like a more promising way of checking the network.

Rogue actors that want to spoof their node performance will do so regardless of the solution we choose. They're in full control of their and will always be capable of spoofing. I don't think we should worry about that too much.

PR #74, opened 2018-11-15
NEP: Merkle State Proofs

igormcoelho:
We are proudly to propose a first draft on how Merkle State Proofs could be easily used to validate storage and voting information, also to provide a simple transition to a Neo blockchain without UTXO model.
We would be happy to have more names in this NEP, so if you want to contribute, please join us :)

### DIFF
--- a/nep-state.mediawiki
+++ b/nep-state.mediawiki
@@ -1,0 +1,40 @@
+<pre>
+  NEP: <to be assigned>
+  Title: <NEP title>
+  Author: <list of authors' names and optionally, email addresses>
+  Type: <Standard | Informational | Meta>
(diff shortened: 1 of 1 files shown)


vncoelho:
@igormcoelho, an initial implementation can be done such as the Storage Changes Dumper: https://github.com/neo-project/neo-plugins/commit/db04e756a64c149e86d117fe89879c72f319c7df

An optional plugin can be designed for dumping the merkle tree states on every block.
I think it should also be stored on cache/db in order reply RPC calls.

erikzhang:
This should be implemented in NEO 3.0.

PR #76, opened 2018-11-18
NEP: Non fungible tokens with expirable transfer

quelibrio:
to be reviewed and improved

### DIFF
--- a/nep-X.mediawiki
+++ b/nep-X.mediawiki
@@ -1,40 +1,63 @@
 <pre>
   NEP: <to be assigned>
-  Title: <NEP title>
-  Author: <list of authors' names and optionally, email addresses>
-  Type: <Standard | Informational | Meta>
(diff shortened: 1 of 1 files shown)


erikzhang:
@neo-project/everyone I modified the title. We should discuss this proposal.

quelibrio:
cool i'll write you to elaborate things

vncoelho:
@erikzhang, do you prefer this as a specially contract functionality implemented by a thirdy party?

erikzhang:
Yes.

quelibrio:
Guys don't delete pull request please. We didn't implement this to the end because we found one flaw in the concept. Each time we calculate if the token has to be returned we call the smart contract which manages those assess and that costs additional gas. If we do that too often its a too expensive operation. But if the person who wants tokes returned invokes the contract its more effective. However its a bit strange scenario. We consider reimplementing this in python.

PR #77, opened 2018-11-25
WIP: dao voting proposal

giancarloGiuffra:
we came up with this proposal for the junction hackathon


igormcoelho:
This is an amazing proposal! Looking forward to see an implementation.

erikzhang:
@giancarloGiuffra Any updates?

corollari:
@erikzhang The proposal got stuck due to our unability to implement it. Currently there doesn't exist any implementation of threshold homomorphic encryption, so implementing this proposal would require us to roll our own crypto primitives.

PR #78, opened 2018-11-25
Oracles blockchain for NEO 3.0

shargon:
The objective of this proposal is to be the first chain capable of not needing oracles to interact with non-deterministic content. Our mission is to make NEO capable of performing REST API calls and other types of non-deterministic functionalities that can be extended in the future.


igormcoelho:
Well done @shargon! This would bring amazing oportunities for external executions and interaction with real world. Protecting IP could be done through vpn, I guess that wont be an issue. Congratulations!

belane:
We believe that it is a great functionality and it will provide Neo with great value by incorporating the Oracles into the blockchain.

Currently Shargon and I have an open discussion of where it would be better to include the information from the Oracles.
- Include the data in the header of the block, which would reduce the size of the block and consume less external requests.
- Include the data into the TX itself which would allow all nodes to verify the transaction when receive it and before relay it, decreasing the processing in the consensus nodes.

nep-oracles.mediawiki
```diff
@@ -0,0 +1,54 @@
...
+  Title: Neo oracles
+  Author: Shargon and Belane
+  Type: Standard
+  Status: Draft
+  Created: 2018-11-22
+</pre>
+
+==Abstract==
+
+The objective of this proposal is to be the first chain capable of not needing oracles to interact with non-deterministic content. Our mission is to make NEO capable of performing REST API calls and other types of non-deterministic functionalities that can be extended in the future.
```
vncoelho: Hey, SharLane, why you mention REST API calls? Would it be the same as an RPC call?

nep-oracles.mediawiki
```diff
@@ -0,0 +1,54 @@
...
+
+[[File:oracles/oracles-image-1.png|center|header]]
+
+A section called "MetaData" will be added to the block that will store, following the key-value structure, the information necessary for the execution that will allow the persistence of the result over time. ''This header (or the entire block) can be compressed using Gzip.''
+
+In order to provide the system with a non-deterministic execution, Smart Contracts will be provided with a set of Access Proxies to all those elements that could cause an unstable result (through SYSCALL).
+
+The result of the execution of a non-deterministic function will be stored in the metadata of the block header, allowing the future persistence of the result thus giving the system determinism in time. The execution of a Smart Contract will check if the result has been stored, in which case, the value will be reused. Otherwise, the complete execution of the contract will be performed and the result will be stored in the metadata of the block according to the following cycle:
```
vncoelho: Now I got it, metadata will storage results of that specific executions.

Did you think about a zero knowledge proof to be sure that that non-determinism, at least, followed the correct pattern? Toff question...aehauheauheaea

nep-oracles.mediawiki
```diff
@@ -0,0 +1,54 @@
...
+
+[[File:oracles/oracles-image-2.png|center|flow]]
+
```
vncoelho: I did not get the `Return Result`.
Will it call the execution again? What is the different between the `Save`?

nep-oracles.mediawiki
```diff
@@ -0,0 +1,54 @@
...
+Another option is to include the information of the results of the Oracles in the remark field of the transaction itself, being able to validate each TX for specific results.
```
vncoelho: As I understood until, I think I prefer in the TX, but I need to understand it further.

nep-oracles.mediawiki
```diff
@@ -0,0 +1,54 @@
...
+
+The origin/creator of the transaction will include the results of the calls in the TX itself and will sign it. Then, both the nodes and the consensus will verify that this information matches the same calls to the Oracles.
+Basically the internet proxy would be a function of the style:
+
+[[File:oracles/oracles-image-3.png|center|source]]
+
```
vncoelho: More organization, please.. Where is the indentation in this figure? jajajajajaja

nep-oracles.mediawiki
```diff
@@ -0,0 +1,54 @@
...
+===Consensus===
+
+If, during the execution of the block (for each consensus node), values other than those stored in the metadata of the returned block are obtained, that block won’t be signed and the TXs containing those calls will be discarded. This is because the determinism of the non-deterministic call cannot be guaranteed at the moment of execution.
+The verification of transactions with Oracle operations will always be carried out in background, the generation time of a block will never be increased. If one of the consensus nodes does not have the result of an oracle transaction prior to the creation or verification of a block, it will be discarded.
+
+===Security concerns===
+
+# The execution of the oracle syscalls by the consensus nodes could leak its IP address. Therefore it is recommended to have several network interfaces and the use of VPN for this purpose.
+# The processing of these Syscall will increase the consumption of the nodes during the transactions verification, therefore, operations should be more expensive, and with a reduced timeout.
+# The use of metadata avoid duplicate information when there are several TXs calling the same contract or API.
```
vncoelho: I did not get this last point of duplicated information.

trueinsider:
I imagine that with storing each URL content on-chain ledger size could become quite big quite fast. Is there a possibility in using this somehow with NeoFS, so that metadata would be stored there instead of directly in block?

erikzhang:
In order to ensure the consistency of contract execution, you cannot rely on NeoFS.

trueinsider:
@erikzhang I see, I didn't know about consistency guarantees that NeoFS provides (or don't). That makes sense then, thanks

On another note, I can think of several different approaches to Oracles:
1) Querying arbitrary deterministic data. Even though the current proposal mentions non-deterministic data it's still, in a sense, deterministic. Because URL content should be the same each time it's queried by consensus nodes during verification, otherwise it'll fail. This, however, doesn't allow us to query data that will be different on each call (e.g. random numbers).
2) Querying specially-formatted non-deterministic data. When using Oracles you explicitly trust the resource you're getting data from, only thing that we can ensure is that that data definitely came from that resource and wasn't modified by Oracle or someone else. For that purpose we can enforce the data format, which contains data itself and signature that allows us to verify that that date came from the expected resource and wasn't modified. This approach won't allow us to query arbitrary APIs, but still will allow us to query data from off-chain after some preparation. Example of such API is random.org Signed API https://api.random.org/json-rpc/2/signed using which you can guarantee that number was generated by no one else than random.org itself.
3) Querying arbitrary non-deterministic data. This is a slight modification of previous approach. We allow SCs to query arbitrarily-formatted data while verifying it's correctness is SC dev responsibility. If there is a signature present like in above-mentioned API then SC dev should extract and verify it. Though this will require interop/opcode for signature verification in SCs. If there is no signature then authenticity of data can't be verified but this still can be used for some non-mission-critical data.

Last two approaches also allows for another implementation strategy, which doesn't bring any overhead to block structure or transaction verification.
We can't use consensus because data is non-deterministic, but since it's signed there is no need for consensus anyway (we can verify that it came from expected source w/o consensus).
This could be implemented like this:
1. Submit data request (and pay GAS for it) to Oracle SC
2. One of the oracle nodes sees request, fetches data and submit it on-chain. First one to submit would receive GAS sent along with request to this data.
3. Original SC can then query this data from on-chain and continue execution (this should be separate tx from the one in 1).

That way we don't need to change the block structure, Oracles nodes is decoupled from Consensus nodes and there is incentive for people to run Oracle nodes, guaranteeing that there always will be someone to put requested data on-chain as he is incentivized to do so.
It's possible (if we want) to implement some kind of Oracle node registration mechanism so that only trusted people will be able to run one.
It's also possible to implement another rewarding rule instead of "first one wins" to ensure more fair reward distribution process (e.g. round-robin).

Another advantage of this approach is that it could be safely backported to Neo 2.0, we only need to make sure that Oracle SC hash won't change in Neo 3.0. And we could use CGAS to pay to Oracle (which should have same contract hash as GAS native contract in Neo 3.0 for backwards compatibility reasons)

erikzhang:
This should be implemented in NEO 3.0.

PR #45, opened 2018-05-21
NEP: Distributed Payments for UTXO Assets

igormcoelho:
This NEP intends to provide a standard way of providing distributed payments on NEO network for UTXO assets, such as NEO and GAS. This way, payments can occur without the intervention of the contract owner and the interested party can send/receive the coins by itself.
We propose standard function signatures that could be used for integration with current wallets and smart contracts, including different implementations (with different tradeoffs).

PR #79, opened 2018-11-28
NEP: Fees paid by the Contract

belane:
A proposal to establish how smart contracts can pay the costs of users' operations.

### DIFF
--- a/nep-feesbycontract.mediawiki
+++ b/nep-feesbycontract.mediawiki
@@ -19,12 +19,12 @@ Currently, however, the fee payment system is limited so that only the sender pa
 
 Allowing projects to pay for the running costs of the network for the users may give NEO the edge in attracting projects interested in this feature, helping users make calls to contracts or token transfers without even having any GAS in their wallet.
 
-Transfers will carry fee thus they will be considered priority transactions. Projects can deploy contracts that will allow commission on contract funds to be charged based on the logic established by the contract itself.
+Transfers will carry fee and so they will be considered priority transactions. Projects can deploy contracts that will allow commission on contract funds to be charged based on the logic established by the contract itself.
 
 
 ==Specification==
 
-NEO calculates a transaction’s fee based on the GAS difference<sup>1</sup> between the Inputs and Outputs no matter the source or destination. Thus, a new Input and Output (from the contract) can be added to the transaction with the GAS difference that is to be paid for the network fee.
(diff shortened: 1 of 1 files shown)


shargon:
A NEP is required just for define the way to use this feature if you want to make a Smart Contract in the future.

vncoelho:
@RavenXce, we had a NEP for that #45, but we closed it due to the possibility of removing UTXO in the future.

vncoelho:
@shargon and @belane, I like the proposal and the idea, congratulations.
I think it is a important NEP for the Ecosystem.

As I mentioned above, we closed our proposal related to Distributed Payment in favor of the possibility of the UTXO removal in Neo 3.0.
In this sense, I think that this proposal also fits in this idea, maybe you could postpone it in order to implement it directly without the UTXO model (it that is the case).

What do you think?

erikzhang:
If the contract pay for the invocation, I think the gas in the contract will soon be squandered.

belane:
I think it will depend on the rules set by the smart contract owner.
For example, the smart contract can have the following rules:

- Limit maximum free fee value.
- Allow only this feature to registered users.
- Allow only X free transfers per day/total/..

It's an interesting feature that allows projects to assume the total cost of the service, getting rid of the user from worrying about having gasoline in their wallet.
It also allows users to not have to exchange project's tokens by GAS to invoke the contract.

Another idea, the owner could have a high NEO balance in the smart contract and pay users' fee with the generated GAS.

@erikzhang  As you say, if the contract does not provide any protection measure, such as maximum fee value, balance could be quickly squandered. So maybe we should highlight the importance of these rules.

erikzhang:
https://github.com/neo-project/neo/blob/289850131500c88b2237e3e6181d1882ccdab876/neo/Network/P2P/Payloads/Transaction.cs#L28

@belane I think it can already be implemented on NEO 3.0. Just put the contract's hash to the `Transaction.Sender`.

belane:
Cool! It could be used even in NEO 2 without modifications in the core, just some adjustments on the client side, but I think it is better to leave it for NEO3 where we won't have to look for an UTXO to use it.

So, It's important to establish a standard that serves as a basis for projects that want to use this feature, because a bad implementation could have security risks for the project.

erikzhang:
Do you think a standard is necessary? Perhaps a template is enough.

belane:
Yes, maybe a template is enough.

erikzhang:
So I will close it first. If we think it is necessary to set a standard later, we can reopen it.

PR #92, opened 2019-06-20
Cross-contract Call Attack Prevention Standard

WyattMufson:
This proposal introduces a standard to protect contracts from cross-contract call attacks.

### DIFF
--- a/nep-ccap.mediawiki
+++ b/nep-ccap.mediawiki
@@ -1,0 +1,89 @@
+<pre>
+  NEP: <to be assigned>
+  Title: Cross-contract Call Attack Prevention Standard
+  Author: Wyatt Mufson <wyatt@ryu.games>
+  Type: Standard
+  Status: Draft
+  Created: 2019-06-20
+</pre>
+
+==Abstract==
+
+This NEP describes a standard for cross-contract call attack prevention in digital asset smart contracts.
+
+==Motivation==
+
+The motivation for this contract comes from the lack of cross-contract call attack prevention documentation and practices in smart contract development.
+The current proposals for digital assets on the NEO blockchain do not address this issue and therefore put users' assets at risk.
+
+It was ultimately inspired by the [[https://github.com/ontio/OEPs/pull/48 | OEP-10 proposal]] I previously authored.
+
+==Specification==
+
+===Overview===
+
+This NEP introduces a process for preventing malicious contracts to invoke smart contracts conforming to this standard.
+By default, users will only be able to invoke the contract directly from their wallet.
+In order to be able to invoke the contract from another contract they will need to manually approve that second contract.
+Users will also have the ability to unapprove contracts, as well as view whether a contract has been approved.
+
+This NEP introduces three methods to be implemented within the <code>Main</code> function: <code>approveContract</code>, <code>unapproveContract</code> and <code>isApproved</code>.
+
+It also introduces an internal method, <code>RequireApproved</code> to be included within any method that changes the state.
+For example, an NEP5 contract that also follows this NEP will call <code>RequireApproved</code> at the beginning of the <code>transfer</code> method.
+
+===Public Methods===
+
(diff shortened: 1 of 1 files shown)


EdgeDLT:
Hey Wyatt, good to see you bringing some of those ideas back over to team green 😉

From what I can understand, it seems the planned [Manifest and Permission system](https://github.com/neo-project/neo/issues/287) would resolve the same problem in a more comprehensive manner. Can you verify?

vncoelho:
Exactly, @edgedlt.
I believe it is more a requeriment than a proposal, the ABI implementation is specifing constraints.

WyattMufson:
@EdgeDLT ah hadn't seen that before. Is the manifest able to be updated after deploying? And if so, does it cost 100-1000 GAS like Contract.Migrate?

Seems like the manifest probably solves the issue more than this proposal.

EdgeDLT:
Implementation is [here ](https://github.com/neo-project/neo/pull/766)if you'd like to take a look. [Apparently](https://github.com/neo-project/neo/pull/766#issuecomment-494219076), permissions will be updatable without migration.

WyattMufson:
Yeah looks good to me. Would just be curious on the cost/difficulty to update the manifest.

WyattMufson:
It looks like `Contract.Update` costs the same as `Contract.Create`. While the manifest is more comprehensive, it does look like it would be too expensive with the current fee structure: https://github.com/neo-project/neo/blob/master/neo/SmartContract/InteropService.NEO.cs#L31

WyattMufson:
Also is this included in NEO 2.x or just in 3.0?

EdgeDLT:
I expect just 3.0, 2.x is LTS now it seems.

Recommend checking against the 3.0 fee structure.

WyattMufson:
@EdgeDLT Hm. Would this standard have usage for NEO 2.x, since 3.0 is still a year away? Any idea what the fee will be for `Contract.Update`? @erikzhang ?

shargon:
Manifest and permission system are only for neo3

WyattMufson:
So then this looks like a good solution for NEO 2.x, right? @EdgeDLT what are your thoughts?

@shargon is it clear yet what the gas fee will be? Doesn't it look like `Contract.Update` and `Contract.Create` will be the [same gas fee](https://github.com/neo-project/neo/blob/master/neo/SmartContract/InteropService.NEO.cs#L31)? If this is prohibitively high, like in 2.x then I think this NEP could be useful in 3.0.

If 3.0 is not prohibitively expensive to update, what would people's thoughts be? I know 2.x is not getting new features but that doesn't mean projects aren't planning on launching on NEO before 3.0.

EdgeDLT:
I think we should move ahead on this.

When a NEO3 TestNet exists we can probably assume that most projects will focus on building for that. Until then, there are probably going to be others deploying or migrating contracts on NEO 2.x. There's not really a downside with providing a standard that can offer them some extra security.

WyattMufson:
I agree with @EdgeDLT on this

WyattMufson:
@erikzhang @shargon

nep-ccap.mediawiki
```diff
@@ -0,0 +1,89 @@
...
+
+The <code>RequireApproved</code> method takes no arguments.
+If the results of <code>GetCallingScriptHash()</code> and <code>GetEntryScriptHash()</code> do not match, then the <code>callerHash</code> must be in the whitelist for the wallet.
+If not, this method must throw an exception.
+
+==Rationale==
+
+This NEP was designed with the usability and security of digital assets on the NEO blockchain in-mind.
+The goal is to enable a new type of asset on NEO that will be optimized for use in smart contracts while maintaining the security of its users.
+Contracts following this NEO are meant to also be following a standard for other crypto assets - such as NEP5, NEP10 and NEP11.
```
corollari: NEO -> NEP

PR #104, opened 2019-09-09
Add readOnly to NEP-3

shargon:
Camelcase for json property names

### DIFF
--- a/nep-3.mediawiki
+++ b/nep-3.mediawiki
@@ -65,7 +65,7 @@ Method object has the following structure:
 
 <code>returnType</code> indicates the return type of the method. It can be one of the following values: <code>Signature</code>, <code>Boolean</code>, <code>Integer</code>, <code>Hash160</code>, <code>Hash256</code>, <code>ByteArray</code>, <code>PublicKey</code>, <code>String</code>, <code>Array</code>, <code>InteropInterface</code>, <code>Void</code>.
 
-<code>readOnly</code> is <code>Boolean</code> which describe if the method require write access to the storage or not.
+<code>readOnly</code> is <code>Boolean</code> which describe if the method require write access to the storage or not. Also, syscalls like <code>Neo.Native.Deploy</code>, <code>Neo.Contract.Create</code>, <code>Neo.Contract.Update</code> or <code>Contract.Destroy</code> will be banned if <code>readOnly</code> is True.
 
 ===Event===
 
--- a/nep-3.mediawiki
+++ b/nep-3.mediawiki
@@ -32,7 +32,13 @@ The NeoContract ABI is defined by JSON format, which has the following basic str
 <pre>
 {
   "hash": "0x562851057d8afbc08fabc8c438d7cc771aef2195",
-  "entrypoint": "main",
+  "entryPoint": 
+    {
+    "name": "Main",
+    "parameters": [],
+    "returnType": "Boolean",
+    "readOnly": "Boolean",
+    },
   "methods": [],
   "events": []
 }
@@ -40,7 +46,7 @@ The NeoContract ABI is defined by JSON format, which has the following basic str
 
 <code>hash</code> is the script hash of the contract. It is encoded as a hexadecimal string in big-endian.
 
-<code>entrypoint</code> is a Method object which describe the details of the entrypoint of the contract.
+<code>entryPoint</code> is a Method object which describe the details of the entrypoint of the contract.
 
 <code>methods</code> is an array of Method objects which describe the details of each method in the contract.
 
@@ -54,7 +60,8 @@ Method object has the following structure:
 {
   "name": "transfer",
   "parameters": [],
-  "returntype": "Boolean"
+  "returnType": "Boolean",
+  "readOnly": "Boolean"
 }
 </pre>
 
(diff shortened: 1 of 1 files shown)


nep-3.mediawiki
```diff
@@ -32,15 +32,21 @@ The NeoContract ABI is defined by JSON format, which has the following basic str
 <pre>
 {
   "hash": "0x562851057d8afbc08fabc8c438d7cc771aef2195",
-  "entrypoint": "main",
+  "entryPoint":
```
erikzhang:
```
  "entryPoint": {}
```

nep-3.mediawiki
```diff
@@ -54,15 +60,18 @@ Method object has the following structure:
...
 </pre>
 
 <code>name</code> is the name of the method, which can be any valid identifier.
 
 <code>parameters</code> is an array of Parameter objects which describe the details of each parameter in the method.
 
-<code>returntype</code> indicates the return type of the method. It can be one of the following values: <code>Signature</code>, <code>Boolean</code>, <code>Integer</code>, <code>Hash160</code>, <code>Hash256</code>, <code>ByteArray</code>, <code>PublicKey</code>, <code>String</code>, <code>Array</code>, <code>InteropInterface</code>, <code>Void</code>.
+<code>returnType</code> indicates the return type of the method. It can be one of the following values: <code>Signature</code>, <code>Boolean</code>, <code>Integer</code>, <code>Hash160</code>, <code>Hash256</code>, <code>ByteArray</code>, <code>PublicKey</code>, <code>String</code>, <code>Array</code>, <code>InteropInterface</code>, <code>Void</code>.
+
+<code>readOnly</code> is <code>Boolean</code> which describe if the method require write access to the storage or not.
```
erikzhang: I think it's not only the storage, but also the SYSCALLs like `Contract.Destroy`.

nep-3.mediawiki
```diff
@@ -156,7 +165,8 @@ It is strongly recommended that each contract have the following entry point:
       "type": "Array"
     }
   ],
-  "returntype": "Any"
+  "returnType": "Any",
+  "readOnly": "Boolean"
```
erikzhang:
```suggestion
  "readOnly": "false"
```

erikzhang:
Can we discuss it again, why do we need `readOnly`? What is its use?

shargon:
https://github.com/neo-project/neo/issues/927

Is for ensure that is not possible to execute a write operation in these methods, in order to prevent the write access

erikzhang:
Why to prevent the write access?

shargon:
As @igormcoelho said:

> offchain services execute operations by themselves (unless forced to send to chain); and two, a readonly storage context is provided

Before this change, the developer promise to not modify the state if the method is `safe`, after this change is not a promise, is a must.

Also, the Abi will be compleated by the compiler acording to the Attributes, and developers doesn't need to create/modify the Abi.

erikzhang:
Can we remove `readOnly` from ABI, and extend the syscall `System.Contract.Call`, add a new parameter `readOnly` to it?

shargon:
But then, how the wallet know when they need to execute the context as readOnly?

erikzhang:
Why a wallet need to execute the context as readOnly?

shargon:
Imagine that you want to create a website like etherscan.io, and allow to execute readOnly methods (without relay), you need to know what methods are `readOnly` and what methods don't. 

https://etherscan.io/address/0x01eacc3ae59ee7fbbc191d63e8e1ccfdac11628c#readContract

erikzhang:
I think readOnly and relay are irrelevant.

shargon:
How do you do this website without know which methods are readOnly and which methods not?

erikzhang:
I don't know. Why people need to know it? If I make this website, I will provide a uniform list of methods without distinguishing between read-only or not.

lock9:
🤔  🤔  🤔 
Lol, you are right @erikzhang , this change makes no sense 😅 . I don't think people would even send TX to the blockchain that does not alter the storage. And even if they did, there is no difference, because making a method read-only won't remove the dependency of the storage.

Maybe we need a "no storage" flag instead of a readOnly. A method that has no access to the storage (stateless) can be used in optimizations since these are certain to be "thread safe".

shargon:
> Maybe we need a "no storage" flag instead of a readOnly. A method that has no access to the storage (stateless) can be used in optimizations since these are certain to be "thread safe".

This is precisely for this, allow to create methods that ensure that no changes will be made on the blockchain, has the same behavior of `view` in ETH

lock9:
@shargon  but read-only still depends on the state, even if it is read-only, we can have different results if the content of the storage is changed. This is not true for real stateless methods

shargon:
maybe we need three states instead of two

`storage=stateLess, writable, readOnly`

But is very important this information in the abi.

lock9:
Why do we need the readonly? I think we only need stateful and stateless methods. And I believe all methods should be either of them, always. If nothing is said, it is stateful. What do you think @shargon ? I can't see the gains of only having "read only". Can you point it to me?

erikzhang:
> I think we only need stateful and stateless methods. And I believe all methods should be either of them, always.

I think even the same method can have different state dependencies on each execution.

```c#
if (needToWrite)
{
    //Write something to storage
}
else
{
    //return constant
}
```

So we should specify an option each time we call a method, which can include an agreement on whether the called method has permission to access the state.

shargon:
For me is the same case as `view` in ethereum, but they don't do any check at runtime, is only for the abi

https://solidity.readthedocs.io/en/latest/contracts.html#view-functions

*"This means library view functions do not have run-time checks that prevent state modifications"*

shargon:
Any news? this should be closed or updated?

corollari:
Another reason that makes this interesting is that it allows libraries that interact with the blockchain, such as neon-js, to detect which calls require actual transactions and which can be done using a simple RPC call to any node, while providing a unified API.

Example:
```
contract.getName() // -> RPC call
contract.setName("hey") // -> transaction
```

erikzhang:
https://github.com/neo-project/neo/blob/24da2bd8df2612643b3a4840cca2bea77b305daf/neo/SmartContract/Manifest/ContractManifest.cs#L59

The `readonly` is in manifest, and it is named as `safemethods`. So it is easy to be checked.

This proposal is to move it from manifest to abi. It is a different thing.

corollari:
I see, you are right in that my comment is invalid then.

lock9:
I vote to move this to the ABI, we can use an attribute at smart contract level. As @corollari mentioned, this is helpful to know when we can use RPC directly and it may also be useful for optimizations. Also, the ABI is an adequate place for it because it is where we store information about the methods in our contract.

PR #105, opened 2019-09-26
Update NEP3 to camelcase

shargon:
Change NEP3 according to https://github.com/neo-project/neo-devpack-dotnet/pull/109


shargon:
I have no write access, could you merge it?

PR #111, opened 2020-03-18
Amend NEP6 for NEO3

shargon:
Now the script it's in base 64 format

*Based in my conversation with guil from neow3j*

### DIFF
--- a/nep-6.mediawiki
+++ b/nep-6.mediawiki
@@ -100,13 +100,13 @@ Contract object has the following structure:
 
 <pre>
 {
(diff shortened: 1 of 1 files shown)


erikzhang:
The address format should be changed too.

ixje:
1) The implementation link to the NEP6Wallet.cs in this NEP is invalid 
2) Both linked implementations are now invalid due to this change

shargon:
`NEP6Wallet.cs` from neo3? what it's wrong?

ixje:
> `NEP6Wallet.cs` from neo3? what it's wrong?

1. The link gets a `404`. 
2. You're right on the latest implementation of `NEP6Wallet.cs` to correctly support it. Was watching at an old copy 😑

PR #112, opened 2020-03-18
Amend NEP3

shargon:
It doesn't include `Map` in the accepted parameters

### DIFF
--- a/nep-3.mediawiki
+++ b/nep-3.mediawiki
@@ -62,7 +62,7 @@ Method object has the following structure:
 
 <code>parameters</code> is an array of Parameter objects which describe the details of each parameter in the method.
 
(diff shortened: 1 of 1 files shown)


nep-3.mediawiki
```diff
@@ -129,6 +129,9 @@ ParameterType enum has the following values:
 | Array
 | An array of objects. The type of elements can be any value of ParameterType.
 |-
+| Map
+| A map of objects. The type of elements inside the key/value collection can be any value of ParameterType.
```
ixje: @shargon I don't think this description is correct.
* A `Map` only takes a PrimitiveType as key (Boolean, ByteArray, Integer)
* A `Map` takes `StackItem` as a value. `Any` and `Void` are not `StackItem`'s.

src: https://github.com/neo-project/neo-vm/blob/754bca3568c50695c571cb796c500fd66c07d0fe/src/neo-vm/Types/Map.cs#L13

ixje: Bullet 2 also holds for the `Array` description above it.

shargon: Please, could you change it?

PR #118, opened 2020-03-25
Revert NEP-3 and NEP-6

### LINKED ISSUE #115: Create ANOTHER NEP standard for fundamental changes
There were changes directly on the NEP-3. This brings a lot of confusion to the community, IMHO.
If NEP-3 (or NEO-6) is modified in some fundamental way, then, let's do another NEP standard as any other blockchain project. In the example from above, removing an entrypoint is a fundamental change. For sure, no doubts.

If only pieces of text and no functionality is modified, then, it's fine, it's just an amendment and the NEP number should remain untouched.

_original posted by @gsmachado_

What do you think? @neo-project/core

erikzhang:
Closes #115

### DIFF
--- a/nep-3.mediawiki
+++ b/nep-3.mediawiki
@@ -32,12 +32,15 @@ The NeoContract ABI is defined by JSON format, which has the following basic str
 <pre>
 {
   "hash": "0x562851057d8afbc08fabc8c438d7cc771aef2195",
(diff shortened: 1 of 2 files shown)


shargon:
The revert should be in master-2x, isn't it?

erikzhang:
No. I have removed the master-2.x branch.

PR #122, opened 2020-05-25
NEP: Extended Token Standard

steven1227:
Here this NEP defines an extended token standard for the Neo Smart Economy. The extended standard interface for contracts that manage multiple types of tokens including Non-Fungible tokens and Fungible tokens. This Standard can help the developer to manage different kind of tokens much easier. It can also reduce the transaction fee by grouping those tokens together.

### DIFF
--- a/nep-TokenStandard.mediawiki
+++ b/nep-TokenStandard.mediawiki
@@ -9,11 +9,11 @@
 
 ==Abstract==
 
-This NEP defines an extended token standard for the Neo Smart Economy. The extended standard interface for contracts that manage multiple type of tokens including Non-Fungible tokens and Fungible tokens, which are minted with NEO based mortgage.
+This NEP defines an extended token standard for the Neo Smart Economy. The extended standard interface for contracts that manage multiple type of tokens including Non-Fungible tokens and Fungible tokens, which are minted with Nep5 based collateral.
 
 ==Motivation==
 
-Non-Fungible tokens (NFT) have been popular in the crypto word for a while. Since the wide usage of NFT in the collectibles, game items, digital art, event tickets, the simple NFT standard NEP-11 limits the functionality of transferring multiple tokens in one transaction. In addition, in some case, user have to use the non-fungible token and fungible token together. On the other hand, minting NFT without valued mortgage is pointless. Because non-divisible NEO token is Governing Token for the Neo Smart Economy, it adds the value to the minted NFT.
+Non-Fungible tokens (NFT) have been popular in the crypto word for a while. Since the wide usage of NFT in the collectibles, game items, digital art, event tickets, the simple NFT standard NEP-11 limits the functionality of transferring multiple tokens in one transaction. In addition, in some case, user have to use the non-fungible token and fungible token together. On the other hand, minting NFT without valued collateral is pointless. The Nep5 token adds the value to the minted NFT.
 
 ==Specification==
 
@@ -99,7 +99,7 @@ Mint a NFT to an address.
 
 The user who invoke this function MUST be the creator of this type of NFT. If not, this method SHOULD throw an exception.
 
-The invocation of mintNonFungible function SHOULD have at least one <code>NEO</code> as mortgage
+The invocation of mintNonFungible function SHOULD have sent <code>Nep5</code> as collateral. 
 If not, this method SHOULD throw an exception.
 
 The parameter <code>type</code> SHOULD be a valid NFT type.  If not, this method SHOULD throw an exception.
@@ -132,14 +132,14 @@ public static bool burnNonFungible(byte[] from, byte[] tokenId)
 </pre>
 Burn a NFT with <code>tokenId</code> from account <code>from</code>. The parameter <code>tokenId</code> SHOULD be a valid index of that NFT type. If not, this method SHOULD throw an exception.
 
-The account who burns this NFT will get the corresponding <code>NEO</code> mortgage
+The account who burns this NFT will get the corresponding <code>Nep5</code> collateral
 
 
 ====burnFungible====
 <pre>
 public static bool burnFungible(byte[] from, byte[] tokenId, BigInteger amount)
 </pre>
-Burn <code>amount<code> of fungible token with <code>tokenId</code> from account <code>from</code>. The parameter <code>tokenId</code> SHOULD be a valid index of that fungible token type. If not, this method SHOULD throw an exception.
+Burn <code>amount</code> of fungible token with <code>tokenId</code> from account <code>from</code>. The parameter <code>tokenId</code> SHOULD be a valid index of that fungible token type. If not, this method SHOULD throw an exception.
 
 
 
--- a/nep-TokenStandard.mediawiki
+++ b/nep-TokenStandard.mediawiki
@@ -1,0 +1,162 @@
+<pre>
+  NEP:
+  Title: Extended Token Standard
+  Author: Steven Liu <liurendong@neo.org>
+  Type: Standard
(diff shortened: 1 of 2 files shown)


erikzhang:
Don't modify README.mediawiki when creating PR.

erikzhang:
And don't assign NEP number by yourself.

steven1227:
> And don't assign NEP number by yourself.

OK. Will change it soon

nep-TokenStandard.mediawiki
```diff
@@ -0,0 +1,162 @@
...
+If the Mint is not processed, the function SHOULD return <code>false</code>.
+
+====mintFungible====
+<pre>
+public static bool mintFungible(byte[] type, byte[] to,  BigInteger amount)
+</pre>
+Mint an <code>amount</code> of Fungible Token to an address.
+
+The user who invoke this function MUST be the creator of this type of Fungible Token . If not, this method SHOULD throw an exception.
+
```
vncoelho, line 120: By `type` you mean the `hash` of the contract, @steven1227?

steven1227:
The word **mortgage** maybe inappropriate. Let us change it to collateral

nep-TokenStandard.mediawiki
steven1227: Here `type` is the unique mark of one kind NFT/FT. In this standard, the hash of this contract is just the address of this contract, has no use inside the contract. Each kind of NFT/FT in this standard will be created by the creator with an assigned `type`. The `type` can be used to clarify which NFT/FT a token belongs, and also used to distinguish if it is an NFT or FT by using bit masking on this `type`

steven1227:
The reason why the neo ecosystem is needing the new extended token standard is to combine the FT and NFT together in one system and decrease the expense of multiple times transfer transaction invocation. The other regulation this the collateral of the NFT. In most cases, the NFT should have value inside and the NEO token used as collateral makes it naturally has the value in the neo ecosystem.

Below is the explanation of NEP-5 token (sorry for my terrible drawing skill).
WechatIMG1

The NEO-5 token standard is great when the user wants to have some initial token offering and makes the basic type of smart economic available on the neo ecosystem. However, those tokens are fungible tokens which means there is still a scenario that we need to define digital items with identical ownership. Here comes the NEP-11 proposal which is the NFT standard in the neo ecosystem:

WechatIMG22

The purpose of the NEP-11 standard is to make the NFT available in the neo ecosystem. The NFT standard is different from NFT in the other blockchain is that the NEP-11 make the NFT is divisible.
This makes the NFT power it is kind of making the NFT also can be fungible when there is a specific case. However, this makes the developer such as Dapp/Wallets/Explorers difficult to the developer and maintain.

WechatIMG33

 In addition, it also makes the transaction of some scenarios very expensive. Imaging in one game, you have a big car, then inside the car, each equipment is an NFT such as tire, exhaust pipe.  So for such a case, you have to create a number of NFT contracts as per type. When the user exchanges the car to another account, you have two make two different transactions. It increases the fee of the transaction.

Here is the new design of the token standard that includes both the FT and NFT. The creator can first create a type of token. It can be FT or NFT. Once the token type is defined, the user can mint the related token. In such a way, one contract can manage the unlimited type of FT/NFT, and user can batch transfer both the NF and NFT in one transaction. The standard is compatible with the NEP-11 which is also managing each NFT as indivisible and can make the divisible ones as a fungible token.

WechatIMG4

Finally, here when minting the NFT, the user has to use at least one NEO as collateral. This is required for making the NFT has backing value. The backing valued NFT tokens represents unique digital or physical assets.
WechatIMG5

EdgeDLT:
This is likely a good direction to move. Ethereum has shown some success in this regard with ERC-1155, with Enjin also moving in the direction of backing each item with another token (ENJ in their case) in order to "imbue" the NFT with value.

I think this is a very interesting proposal, especially if it possible to melt/burn the asset to retrieve the collateral. Say if a game goes offline and NFTs would become worthless; this means that the user can always get something back out of it.

One important change would be to use GAS for collateral instead of NEO. Indivisibility means that NEO has no granularity for this use case, and we should not be locking the governance token into random NFTs, we should be encouraging NEO holders to use their NEO to vote.

nimmortal:
Hello guys!
We at blockchaincuties have implemented something similar to the proposed standard.
However, we got issues with methods like `tokensOf`, because it can be too expensive to read the whole storage. From my point of view, contract interface looks good, but need more technical details how to implement `batchTransfer` and other things to make it cost efficient.

imyourm8:
As i understand **type** is part of the token id?

imyourm8:
> Hello guys!
> We at blockchaincuties have implemented something similar to the proposed standard.
> However, we got issues with methods like `tokensOf`, because it can be too expensive to read the whole storage. From my point of view, contract interface looks good, but need more technical details how to implement `batchTransfer` and other things to make it cost efficient.

Yes, that's why enumeration should be done through the events.

steven1227:
> This is likely a good direction to move. Ethereum has shown some success in this regard with ERC-1155, with Enjin also moving in the direction of backing each item with another token (ENJ in their case) in order to "imbue" the NFT with value.
> 
> I think this is a very interesting proposal, especially if it possible to melt/burn the asset to retrieve the collateral. Say if a game goes offline and NFTs would become worthless; this means that the user can always get something back out of it.
> 
> One important change would be to use GAS for collateral instead of NEO. Indivisibility means that NEO does not have the required granularity for this use case, and we should not be locking the governance token into random NFTs, we should be encouraging NEO holders to use their NEO to vote.

Good point, may be we can use GAS instead and maybe set a minimum GAS fee requirement?

steven1227:
> Hello guys!
> We at blockchaincuties have implemented something similar to the proposed standard.
> However, we got issues with methods like `tokensOf`, because it can be too expensive to read the whole storage. From my point of view, contract interface looks good, but need more technical details how to implement `batchTransfer` and other things to make it cost efficient.

Thanks, BlockchainCuties. Yes, we understand the expensive fee situation in Neo2. However, in Neo3, the fee is adjusted. For the detailed implementation, I (or anyone) can make one when the Neo3 testnet is launching.

steven1227:
> As i understand **type** is part of the token id?

The `tokenId` is specific for each NFT and no matter which type of NFT it is.  The `type` is an id to distinguish the different types of tokens are. They are different meaning and usage

steven1227:
> > This is likely a good direction to move. Ethereum has shown some success in this regard with ERC-1155, with Enjin also moving in the direction of backing each item with another token (ENJ in their case) in order to "imbue" the NFT with value.
> > I think this is a very interesting proposal, especially if it possible to melt/burn the asset to retrieve the collateral. Say if a game goes offline and NFTs would become worthless; this means that the user can always get something back out of it.
> > One important change would be to use GAS for collateral instead of NEO. Indivisibility means that NEO does not have the required granularity for this use case, and we should not be locking the governance token into random NFTs, we should be encouraging NEO holders to use their NEO to vote.
> 
> Good point, may be we can use GAS instead and maybe set a minimum GAS fee requirement?

@EdgeDLT Recently after discussion with other people who are interested in this proposal, how about use `NEP5` token as collateral? In this way, contract can define some specific method that enable user can only stake specific NEP5 token. Then, the NEO/GAS can be also used by using CNEO/CGAS

EdgeDLT:
NEP-5 sounds good to me. It will also work nicely with NEO/GAS when we migrate the standard to Neo3.

erikzhang:
@neo-project/everyone Please check https://github.com/erikzhang/neo-asset-combiner

Issue #97, opened 2019-07-18
Solid State Transfers


igormcoelho:
I propose a NEP that will help us give much more control over token operations, specially on Neo 3. This proposal is simple: allow better balance tracing over storage-based assets, like NEP-5.
On UTXO, it's all connected, the past transactions, past balances, and current balance, giving great auditability capabilities.
On storage-based assets, such as NEP-5, we only execute `transfer` operation on the chain, and our balance is automatically updated, but no record is kept for how much existed before (and after).

This NEP requires that assets prove that they exist, before actual transfer.

Example for NEP-5:
`transfer myaddress youraddress 100`

How much did I have before? zero? 1000? We don't know, unless we process the whole chain at that point (or restore the point with some state trie). This proposal is simple, yet powerful:

`value` = `balanceOf myaddress`
`assert(value == 250)` # ok, now I'm sure what my funds were before transfer
`transfer myaddress youraddress 100`
`value2` = `balanceOf youraddress`
`assert(value2 == 600)` # ok, now I'm sure what your funds were after transfer

If this transaction passes on chain, we have a evidence for two things: I had 250; and I have transferred 100 to you; and now you have 600.

More elaborate versions of this could require saving previous and after balances, for both addresses, but personally I think that a single "before" register is enough to guarantee quite a few nice properties.

Justification:
Neo Blockchain is quite special, in the sense that it doesn't add a solid state reference to block header itself, giving fundamental bugfixing capabilities to the network (that already could have prevented hundreds of hard forks on it). There are many proposals for state tracking on Neo3, in many different formats, however, I believe this proposal here adds much guarantees and auditability to NEP-5 tokens, with or without state tracking. 

This is something users (or user wallets) can do for them, it's easy and gives strong certainty of instantaneous states on the chain, just by looking at operations themselves.

Drawbacks:
This may not be wanted on situations where several transactions compete to enter a block, in very high throughput operations (on same address), as this NEP also fundamentally breaks the possibility of "double-spending" of any kind of token, in any situation. Yet, for common daily transfers it looks quite nice.

igormcoelho:
@vncoelho @shargon @jsolman @erikzhang

igormcoelho:
@erikzhang this is a simplified version of NEP https://github.com/neo-project/proposals/issues/102

[...]

PR #123, opened 2020-10-22
Proposal for Solid States

igormcoelho:
This NEP explains and specifies Solid State Transfers, which is a mechanism for easy accounting of NEP-5 balances on Neo2 and Neo3 network, regardless of any other storage audit technology (MPT, etc).
This is complementary to all other security techniques, has no drawbacks for users, but huge advantages.

Neo developers for clients of different programming languages will be strongly benefited, since clients will "naturally break" if they find any transaction in block which is invalid... so this helps "cleaning out" poorly implemented nodes from Neo p2p network.

If you like this idea, feel free to contribute and join us as co-author.

See discussion at: https://github.com/neo-project/proposals/issues/97

### DIFF
--- a/nep-solid.mediawiki
+++ b/nep-solid.mediawiki
@@ -1,0 +1,122 @@
+<pre>
+  NEP: XYZ
+  Title: Solid State Transfers
+  Author: Igor Machado Coelho <igormachado@gmail.com>, Vitor Nazário Coelho <vncoelho@gmail.com>, Fernando Díaz Toledano <shargon@gmail.com>, "Who Else? Join Us!" <you@you.com>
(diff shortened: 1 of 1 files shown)


igormcoelho:
@shargon @vncoelho now it's officially proposed

vncoelho:
An unity for states,like UTXO of state proofs.
I support this direction and can surely follow progres on that line of reasoning, brother.
The power and flexibility of Verification should be enhanced while keeping 1BF.

igormcoelho:
EdgeLT is more than welcome to join us, as a very early supporter of the idea... is the text minimally acceptable? :)

Issue #108, opened 2019-08-29
Invoke contracts' payable method on asset Transfers


belane:
I would love that smart contracts could react to asset transfers, I'm not talking about the `Mint` function, but rather that a smart contract can react, execute, upon receiving NEO or any asset.
(remember our old friend `AplicationR` from `NEP7`?)

After studying several options we found that the solution is right in front of us.
In Native Contracts (NEO/GAS) we can add a call to the address in if the destination is a contract.

In the `Transfer` method of the native contracts we can include a check of whether the destination account is a contract address, check if it has the payable flag and if so, invoke a specific method of the contract, for example `bool Payable()`

`Payable` method can be included in the `NEP5` standard and must return `True` if everything is correct or throw an exception otherwise.

In this way smart contracts can react to payments without users having to build a special transaction, invoke `Mint` or whatever.
Therefore, Mint invocations won't need a second `APPCALL`, saving network and blockchain resources.

It also gives to the contract full management of received payments such as denying money receipts, issuing refunds, etc ...

igormcoelho:
@erikzhang @shargon I think I found an interesting solution for this. Please correct me if I missed some recent development.

Suppose we don't make contracts pay for dynamic invoke during deployment, but we make users pay (or not) for it, during invocations (a tiny extra cost).

User wallets will likely disable this, unless they need it (as a failed dynamic invoke would cause FAULT if non paid).

Now, before invoking the "payable" method on other contract, we execute some Interop that requires Manifest/MethodList and verify that the remote contract *actually requires* some payable "reaction".

Ok, now its easy for wallets: if destination requires payable action, it enables dynamic invoke, otherwise it doesn't (and user saves gas).

In any case, this NEP will still be necessary for NEP5 to declare its "willigness" to call remote payable, but I think this resolves the "dynamic invoke dillema".

erikzhang:
Maybe we can add a new SYSCALL: `Contract.CallIfExists()`. Or we can just add a new flag to `CallFlags`.

igormcoelho:
Good idea @erikzhang , automatically checks manifest, simplify a lot coding.
Regarding users paying dynamic invoke fees per execution call, only if needed, instead of contract paying bigger amounts... do you see something like this reasonable? Good thing is that becomes easier to execute parallel tx, if dyn invoke fees are not required in that precise tx.

erikzhang:
[...]

PR #124, opened 2020-10-27
NEP5 call onPayment after transfer

shargon:
- Add `onPayment` method

Close https://github.com/neo-project/proposals/issues/108

### DIFF
--- a/nep-5.mediawiki
+++ b/nep-5.mediawiki
@@ -76,7 +76,7 @@ If the <code>account</code> is an unused address, this method MUST return <code>
 This method is optional, it will be called during NEP5 transfers in order to ensure that the smart contract know how to deal with the token, if don't this method SHOULD <code>throw</code> an exception.
 
 <pre>
-public static void onPayment(byte[] from, byte[] to, BigInteger amount)
+public static void onPayment(BigInteger amount)
 </pre>
 
 ====transfer====
@@ -99,7 +99,7 @@ The function SHOULD check whether the <code>from</code> address equals the calle
 
 If the <code>to</code> address is a deployed contract, the function SHOULD check the <code>payable</code> flag of this contract to decide whether it should transfer the tokens to this contract.
 
-The function MUST call <code>onPayment</code> method AFTER fire the <code>Transfer</code> event. This call need the flag <code>IfExists</code> and the arguments <code>from</code>, <code>to</code>, <code>amount</code>.
+The function MUST call <code>onPayment</code> method AFTER fire the <code>Transfer</code> event. This call need the flag <code>IfExists</code> and the <code>amount</code> argument.
 
 If the transfer is not processed, the function SHOULD return <code>false</code>.
 
--- a/nep-5.mediawiki
+++ b/nep-5.mediawiki
@@ -1,7 +1,7 @@
 <pre>
   NEP: 5
   Title: Token Standard
-  Author: Tyler Adams <lllwvlvwlll@gmail.com>, luodanwg <luodan.wg@gmail.com>, tanyuan <tanyuan666@gmail.com>, Alan Fong <anfn@umich.edu>, Jaime Kindelan <srbelane@gmail.com>
+  Author: Tyler Adams <lllwvlvwlll@gmail.com>, luodanwg <luodan.wg@gmail.com>, tanyuan <tanyuan666@gmail.com>, Alan Fong <anfn@umich.edu>, Jaime Kindelán <srbelane@gmail.com>
   Type: Standard
   Status: Final
   Created: 2017-08-10
--- a/nep-5.mediawiki
+++ b/nep-5.mediawiki
@@ -71,6 +71,14 @@ The parameter <code>account</code> SHOULD be a 20-byte address. If not, this met
 
 If the <code>account</code> is an unused address, this method MUST return <code>0</code>.
 
+====onPayment====
+
+This method is optional, it will be called during NEP5 transfers in order to ensure that the smart contract know how to deal with the token.
+
+<pre>
+public static event onPayment(byte[] from, byte[] to, BigInteger amount)
+</pre>
+
 ====transfer====
 
 <pre>
@@ -90,6 +98,8 @@ If the method succeeds, it MUST fire the <code>transfer</code> event, and MUST r
 The function SHOULD check whether the <code>from</code> address equals the caller contract hash. If so, the transfer SHOULD be processed; If not, the function SHOULD use the SYSCALL <code>Neo.Runtime.CheckWitness</code> to verify the transfer.
 
 If the <code>to</code> address is a deployed contract, the function SHOULD check the <code>payable</code> flag of this contract to decide whether it should transfer the tokens to this contract.
+
+The function MUST call <code>onPayment</code> method with the flag <code>IfExists</code> and the arguments <code>from</code>, <code>to</code>, <code>amount</code>.
 
 If the transfer is not processed, the function SHOULD return <code>false</code>.
 
--- a/nep-5.mediawiki
+++ b/nep-5.mediawiki
@@ -73,10 +73,10 @@ If the <code>account</code> is an unused address, this method MUST return <code>
 
 ====onPayment====
 
-This method is optional, it will be called during NEP5 transfers in order to ensure that the smart contract know how to deal with the token.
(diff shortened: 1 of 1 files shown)


erikzhang:
`onPayment` method should be in the target contract. And the `from` parameter is not needed.

nep-5.mediawiki
```diff
@@ -91,6 +99,8 @@ The function SHOULD check whether the <code>from</code> address equals the calle
 
 If the <code>to</code> address is a deployed contract, the function SHOULD check the <code>payable</code> flag of this contract to decide whether it should transfer the tokens to this contract.
 
+The function MUST call <code>onPayment</code> method AFTER fire the <code>Transfer</code> event. This call need the flag <code>IfExists</code> and the arguments <code>from</code>, <code>to</code>, <code>amount</code>.
```
ixje, line 102: This conflicts with the existing description of the `transfer` method.

> If the method succeeds, it MUST fire the transfer event, and MUST return true, even if the amount is 0, or from and to are the same address.

We can't say if `transfer()` succeeds it must fire the `transfer` event, but then also say that after the `transfer` event we call `onPayable`, which in its turn is supposed to be able to modify execution according to this text from #108 
> must return True if everything is correct or throw an exception otherwise. 

I think it is supposed to say "must call onPayment BEFORE firing the Transfer event"

nep-5.mediawiki
```diff
@@ -71,6 +71,14 @@ The parameter <code>account</code> SHOULD be a 20-byte address. If not, this met
 
 If the <code>account</code> is an unused address, this method MUST return <code>0</code>.
 
+====onPayment====
+
+This method is optional, it will be called during NEP5 transfers in order to ensure that the smart contract know how to deal with the token, if don't this method SHOULD <code>throw</code> an exception.
```
ixje, line 76:
```suggestion
This method is optional, it will be called during NEP5 transfers to give the smart contract the opportunity to further process the event and optionally stop execution by throwing an exception.
```

ixje (changes requested):

ixje:
> `onPayment` method should be in the target contract. And the `from` parameter is not needed.

I think `from` can actually be useful. The target contract can have different execution based on the `from` address

roman-khimov:
I think first of all we need a new standard here. NEP-5 is NEP-5 and it shouldn't be changed.

ixje:
yes that also

roman-khimov:
And also, `onPayment` definition can't really be a part of NEP-5 because it should also be invoked for non-NEP-5 receiver contracts. NeoFS mainnet contract is not NEP-5, but it'd be nice to get an `onPayment` invocation for it too.

shargon:
> I think first of all we need a new standard here. NEP-5 is NEP-5 and it shouldn't be changed.

We use different standards for neo2 and neo 3. Before mainnet it's only an amend.

ixje:
> > I think first of all we need a new standard here. NEP-5 is NEP-5 and it shouldn't be changed.
> 
> We use different standards for neo2 and neo 3. Before mainnet it's only an amend.

This just breaks more existing tools unnecessary

nep-5.mediawiki
```diff
@@ -93,6 +93,12 @@ If the <code>to</code> address is a deployed contract, the function SHOULD check
 
 If the transfer is not processed, the function SHOULD return <code>false</code>.
 
+The function MUST call <code>onPayment</code> method AFTER fire the <code>Transfer</code> event. This call need the flag <code>IfExists</code> and the <code>amount</code> argument. If the receiver doesn't want to receive this transfer it SHOULD <code>throw</code> an exception.
```
erikzhang, line 96: Now we have the ability to catch exceptions in the contract. So throwing an exception may not be able to cancel the transfer. ABORT must be called.

erikzhang:
> I think first of all we need a new standard here. NEP-5 is NEP-5 and it shouldn't be changed.

I think it is compatible with NEP-5. Because there is no `onPayment` method in any deployed contract. So nothing actually changed.

roman-khimov:
> I think it is compatible with NEP-5

It's not enough IMO. A standard only works if it's the standard, it doesn't change in any way and any reference to it has exactly the same meaning. Otherwise "NEP-5 compliance" doesn't really mean anything, it could NEP-5 of this or that or some other revision and no one would be able to tell what exactly is meant by this.

vncoelho:
I also believe it is still compatible with NEP-5.
Perhaps we could call it NEO3-NEP-5. In principle, it is basically an improvement that is connected with NEO 3 capabilities. The `onpayment` brings a subtle change in the way contracts can optionally react.

ixje:
> I also believe it is still compatible with NEP-5.

compatibility != compliancy
Just because the added behaviour doesn't necessary burn the system down, doesn't mean it is compliant. As @roman-khimov said, what's the point of a standard (without some form of revision identifier) if you can't identify the **EXACT** requirements it has. That's like saying you don't need to tag any software builds because what's the point?

EdgeDLT:
`onPayment` has use for any kind of contract, not just NEP-5 ones, right? Then it should be its own standard regardless of how compatible it is with existing NEP-5.

I'd rather you amend Neo3's definition of NEP-5 to declare that it must follow a new onPayment NEP, rather than put onPayment straight into NEP-5. Ideally we can avoid amendment entirely and just leave NEP-5 alone. I think a NEP-5 token upgraded to take advantage of new functionality is not a NEP-5 token, it is something new, NEP-20 etc.

vncoelho:
> onPayment has use for any kind of contract, not just NEP-5 ones, right? Then it should be its own standard regardless of how compatible it is with existing NEP-5.

This is a good point, @EdgeDLT.

vncoelho:
@EdgeDLT, in fact, this `onPayment` has other uses as you mentioned, for example, a deployed multi-sig that does not want to receive tokens, airdrops, etc...
Or even a wallet that want to keep it maximum amount limited to `X`.

vncoelho:
I think it would abort execution, @EdgeDLT, not even a need to refunding.

erikzhang:
> onPayment has use for any kind of contract, not just NEP-5 ones, right?

I'm afraid not. The parameter list is NEP-5 specific. For example, if it is NFT transfer, then the parameter list of `onPayment` should contain `tokenid`.

vncoelho:
But that sounds like a good point, @erikzhang for generic use of it.
However, we should also consider that the main need for now is for NEO/GAS.

erikzhang:
If a contract is deployed but the `onPayment` method doesn't exist, should the transfer be processed?

vncoelho:
Maybe yes, @erikzhang, because any valid address should be able to receive assets.
`onPayment` could be seen as a suggested standard and possible protection.

roman-khimov:
> If a contract is deployed but the onPayment method doesn't exist, should the transfer be processed?

We can make `onPayment` mandatory for payable contracts.

erikzhang:
> We can make onPayment mandatory for payable contracts.

Or if there is no `onPayment` method, we think it is not payable. We don't need `payable` field anymore.

nep-5.mediawiki
```diff
@@ -93,6 +93,12 @@ If the <code>to</code> address is a deployed contract, the function SHOULD check
 
 If the transfer is not processed, the function SHOULD return <code>false</code>.
 
+The function MUST call <code>onPayment</code> method AFTER fire the <code>Transfer</code> event. This call need the flag <code>IfExists</code> and the <code>amount</code> argument. If the receiver doesn't want to receive this transfer it MUST call <code>ABORT</code>.
```
erikzhang, line 96:
```suggestion
The function MUST call <code>onPayment</code> method AFTER fire the <code>Transfer</code> event if the receiver is a deployed contract. If the receiver doesn't want to receive this transfer it MUST call <code>ABORT</code>.
```

igormcoelho:
I think @EdgeDLT has a good point.. are we able to create some new standard, NEP-XX which is named like "NEP-5 on Neo3", just including this amend?
This way, amend would have a new number, but could keep popularly calling NEP-5 (on both Neo2 and Neo3).

erikzhang:
I will assign new number for this amendment after it is accepted.

erikzhang:
I created a new PR with a new NEP number: #126

shargon:
Closed in favor of https://github.com/neo-project/proposals/pull/126

PR #125, opened 2020-11-09
Add description for _initialize and _deploy

bettybao1209:


shargon (commented):
Only mention that `Extra` can be `null`

nep-14.mediawiki
```diff
@@ -65,6 +65,14 @@ Method object has the following structure:
 
 <code>returntype</code> indicates the return type of the method. It can be one of the following values: <code>Signature</code>, <code>Boolean</code>, <code>Integer</code>, <code>Hash160</code>, <code>Hash256</code>, <code>ByteArray</code>, <code>PublicKey</code>, <code>String</code>, <code>Array</code>, <code>Map</code>, <code>InteropInterface</code>, <code>Void</code>.
 
+The method <code>_initialize</code> will be generated if static variables are defined in smart contracts. It will be executed automatically to initialize the static variable BEFORE other methods in contracts are called by using syscall <code>System.Contract.Call</code> and <code>System.Contract.CallEx</code>.
```
vncoelho, line 68: The method <code>_initialize</code> will be generated if static variables are defined in smart contracts. It will be executed automatically to initialize the static variable BEFORE other methods in contracts are called by using syscall <code>System.Contract.Call</code> and <code>System.Contract.CallEx</code>.
-> 
The method <code>_initialize</code> will be generated if static variables are defined in smart contracts. It will be **automatically executed, during contract registration**, to initialize the static variable BEFORE other methods in contracts are called by using syscall <code>System.Contract.Call</code> and <code>System.Contract.CallEx</code>.

vncoelho: @bettybao1209  @shargon, @erikzhang, maybe we need to also explain that when we increase the **version** the **_initialize** will be again called, right?

roman-khimov: > static variables are defined in smart contracts

I think it's not correct to talk about `static variables` here, it's a different level of abstraction. NEP-14 should only care about `_initialize` existence and calling it when needed, what it does is up to contract.

nep-14.mediawiki
```diff
@@ -65,6 +65,14 @@ Method object has the following structure:
...
+
+The method <code>_deploy</code> is defined by users which will be called by the system AFTER the contract is deployed or updated by using the syscall <code>System.Contract.Create</code> and <code>System.Contract.Update</code>. The <code>update</code> argument should be FALSE for deploying the contract and be TRUE for updating.  
+
+<pre>
+public static void _deploy(bool update)
```
roman-khimov, line 73: This should be defined as method object defined above.

roman-khimov (commented):
And I'd rather put `_initialize` and `_deploy` into some additional section like "Special methods", because "Method" section is for method object description.

nep-14.mediawiki
bettybao1209: Done.

nep-14.mediawiki
bettybao1209: I have made some updates based on your suggestions. Please review it again.

nep-14.mediawiki
```diff
@@ -65,6 +65,27 @@ Method object has the following structure:
...
+====Special methods====
+
+The method <code>_initialize</code> will be automatically generated and executed to initialize the static variable BEFORE other methods in contracts are called by using syscall <code>System.Contract.Call</code> and <code>System.Contract.CallEx</code>.
+
+The method <code>_deploy</code> is defined by users which will be automatically executed during contract registration or update by using the syscall <code>System.Contract.Create</code> and <code>System.Contract.Update</code>. The structure has the following format:
+
+<pre>
+{
+  "name": "_deploy",
+  "offset": "8",
```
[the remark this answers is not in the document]
bettybao1209: updated.

nep-14.mediawiki
```diff
@@ -65,6 +65,27 @@ Method object has the following structure:
...
+  "parameters": [
+    {
+      "name": "update",
+      "type": "Boolean"
+    }
+  ],
+  "returntype": "Void"
+}
+</pre>
+where the <code>update</code> argument should be set to FALSE for registration or TRUE for update.
```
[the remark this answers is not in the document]
bettybao1209: updated.

erikzhang:
Maybe this should be added to the NEO specifications. ABI is just a description file used to describe which methods the contract contains and how to be called, and should not include the implementation details of NEO.

PR #126, opened 2020-11-12
Replace NEP-5 with NEP-17

erikzhang:
Differences from NEP-5:

1. Added `onPayment` call if the receiver is a deployed contract.
1. Removed `payable` check. It has been replaced by `onPayment`.
1. Removed `name` method, because it should be in manifest for all the contracts, not only for the token contracts.
1. Changed the `transfer` event to capitalize the first letter `Transfer`.

### DIFF
--- a/nep-17.mediawiki
+++ b/nep-17.mediawiki
@@ -26,9 +26,9 @@ In the method definitions below, we provide both the definitions of the function
 
 <pre>
 {
-	"name": "symbol",
-	"parameters": [],
-	"returntype": "String"
+  "name": "symbol",
+  "parameters": [],
+  "returntype": "String"
 }
 </pre>
 
@@ -40,9 +40,9 @@ This method MUST always return the same value every time it is invoked.
 
 <pre>
 {
-	"name": "decimals",
-	"parameters": [],
-	"returntype": "Integer"
+  "name": "decimals",
+  "parameters": [],
+  "returntype": "Integer"
 }
 </pre>
 
@@ -54,9 +54,9 @@ This method MUST always return the same value every time it is invoked.
 
 <pre>
 {
-	"name": "totalSupply",
-	"parameters": [],
-	"returntype": "Integer"
+  "name": "totalSupply",
+  "parameters": [],
+  "returntype": "Integer"
 }
 </pre>
 
@@ -66,14 +66,14 @@ Returns the total token supply deployed in the system.
 
 <pre>
 {
-	"name": "balanceOf",
-	"parameters": [
-		{
-			"name": "account",
-			"type": "Hash160"
-		}
-	],
-	"returntype": "Integer"
+  "name": "balanceOf",
+  "parameters": [
+    {
+      "name": "account",
+      "type": "Hash160"
+    }
+  ],
+  "returntype": "Integer"
 }
 </pre>
 
@@ -87,22 +87,22 @@ If the <code>account</code> is an unused address, this method MUST return <code>
 
 <pre>
 {
-	"name": "transfer",
-	"parameters": [
-		{
-			"name": "from",
-			"type": "Hash160"
-		},
-		{
-			"name": "to",
-			"type": "Hash160"
-		},
-		{
-			"name": "amount",
-			"type": "Integer"
-		}
-	],
-	"returntype": "Boolean"
+  "name": "transfer",
+  "parameters": [
+    {
+      "name": "from",
+      "type": "Hash160"
+    },
+    {
+      "name": "to",
+      "type": "Hash160"
+    },
+    {
+      "name": "amount",
+      "type": "Integer"
+    }
+  ],
+  "returntype": "Boolean"
 }
 </pre>
 
@@ -125,14 +125,14 @@ If the receiver is a deployed contract, the function MUST call <code>onPayment</
 <pre>
 public static void onPayment(BigInteger amount)
 {
-	"name": "onPayment",
-	"parameters": [
-		{
-			"name": "amount",
-			"type": "Integer"
-		}
-	],
-	"returntype": "Void"
+  "name": "onPayment",
+  "parameters": [
+    {
+      "name": "amount",
+      "type": "Integer"
+    }
+  ],
+  "returntype": "Void"
 }
 </pre>
 
@@ -142,21 +142,21 @@ public static void onPayment(BigInteger amount)
 
 <pre>
 {
-	"name": "Transfer",
-	"parameters": [
-		{
-			"name": "from",
-			"type": "Hash160"
-		},
-		{
-			"name": "to",
-			"type": "Hash160"
-		},
-		{
-			"name": "amount",
-			"type": "Integer"
-		}
-	]
+  "name": "Transfer",
+  "parameters": [
+    {
+      "name": "from",
+      "type": "Hash160"
+    },
+    {
+      "name": "to",
+      "type": "Hash160"
+    },
+    {
+      "name": "amount",
+      "type": "Integer"
+    }
+  ]
 }
 </pre>
 
--- a/nep-17.mediawiki
+++ b/nep-17.mediawiki
@@ -124,11 +124,11 @@ The function SHOULD check whether the <code>from</code> address equals the calle
 
 If the transfer is not processed, the function MUST return <code>false</code>.
 
-If the receiver is a deployed contract, the function MUST call <code>onPayment</code> method on receiver contract with the <code>data</code> parameter from <code>transfer</code> AFTER firing the <code>Transfer</code> event. If the receiver doesn't want to receive this transfer it MUST call <code>ABORT</code>.
+If the receiver is a deployed contract, the function MUST call <code>onNEP17Payment</code> method on receiver contract with the <code>data</code> parameter from <code>transfer</code> AFTER firing the <code>Transfer</code> event. If the receiver doesn't want to receive this transfer it MUST call <code>ABORT</code>.
 
 <pre>
 {
-  "name": "onPayment",
+  "name": "onNEP17Payment",
   "parameters": [
     {
       "name": "from",
--- a/nep-17.mediawiki
+++ b/nep-17.mediawiki
@@ -82,7 +82,7 @@ The function SHOULD check whether the <code>from</code> address equals the calle
 
 If the transfer is not processed, the function MUST return <code>false</code>.
 
-The function MUST call <code>onPayment</code> method AFTER fire the <code>Transfer</code> event if the receiver is a deployed contract. If the receiver doesn't want to receive this transfer it MUST call <code>ABORT</code>.
+If the receiver is a deployed contract, the function MUST call <code>onPayment</code> method on receiver contract AFTER firing the <code>Transfer</code> event. If the receiver doesn't want to receive this transfer it MUST call <code>ABORT</code>.
 
 <pre>
 public static void onPayment(BigInteger amount)
--- a/nep-17.mediawiki
+++ b/nep-17.mediawiki
@@ -96,7 +96,7 @@ public static void onPayment(BigInteger amount)
 public static event Transfer(byte[] from, byte[] to, BigInteger amount)
 </pre>
 
-MUST trigger when tokens are transferred, including zero value transfers.
+MUST trigger when tokens are transferred, including zero value transfers and self-transfers.
 
 A token contract which creates new tokens MUST trigger a <code>Transfer</code> event with the <code>from</code> address set to <code>null</code> when tokens are created.
 
--- a/nep-17.mediawiki
+++ b/nep-17.mediawiki
@@ -3,7 +3,7 @@
   Title: Token Standard
   Author: Erik Zhang <erik@neo.org>
   Type: Standard
-  Status: Accepted
+  Status: Final
   Created: 2020-11-12
   Replaces: 5
 </pre>
--- a/nep-17.mediawiki
+++ b/nep-17.mediawiki
@@ -184,6 +184,6 @@ A token contract which burns tokens MUST trigger a <code>Transfer</code> event w
 
 ==Implementation==
 
-C#: https://github.com/neo-project/neo/pull/2024
+C#: https://github.com/neo-project/neo-devpack-dotnet/blob/master/src/Neo.SmartContract.Framework/Nep17Token.cs
 
 Python: https://github.com/CityOfZion/neo3-boa/blob/development/boa3_test/examples/NEP17.py
--- a/nep-17.mediawiki
+++ b/nep-17.mediawiki
@@ -3,7 +3,7 @@
   Title: Token Standard
   Author: Erik Zhang <erik@neo.org>
   Type: Standard
-  Status: Draft
+  Status: Accepted
   Created: 2020-11-12
   Replaces: 5
 </pre>
--- a/nep-17.mediawiki
+++ b/nep-17.mediawiki
@@ -27,6 +27,7 @@ In the method definitions below, we provide both the definitions of the function
 <pre>
 {
   "name": "symbol",
+  "safe": true,
   "parameters": [],
   "returntype": "String"
 }
@@ -41,6 +42,7 @@ This method MUST always return the same value every time it is invoked.
 <pre>
 {
   "name": "decimals",
+  "safe": true,
   "parameters": [],
   "returntype": "Integer"
 }
@@ -55,6 +57,7 @@ This method MUST always return the same value every time it is invoked.
 <pre>
 {
   "name": "totalSupply",
+  "safe": true,
   "parameters": [],
   "returntype": "Integer"
 }
@@ -67,6 +70,7 @@ Returns the total token supply deployed in the system.
 <pre>
 {
   "name": "balanceOf",
+  "safe": true,
   "parameters": [
     {
       "name": "account",
@@ -88,6 +92,7 @@ If the <code>account</code> is an unused address, this method MUST return <code>
 <pre>
 {
   "name": "transfer",
+  "safe": false,
   "parameters": [
     {
       "name": "from",
--- a/nep-17.mediawiki
+++ b/nep-17.mediawiki
@@ -80,8 +80,6 @@ If the method succeeds, it MUST fire the <code>transfer</code> event, and MUST r
 
 The function SHOULD check whether the <code>from</code> address equals the caller contract hash. If so, the transfer SHOULD be processed; If not, the function SHOULD use the SYSCALL <code>Neo.Runtime.CheckWitness</code> to verify the transfer.
 
-If the <code>to</code> address is a deployed contract, the function SHOULD check the <code>payable</code> flag of this contract to decide whether it should transfer the tokens to this contract.
-
 If the transfer is not processed, the function MUST return <code>false</code>.
 
 The function MUST call <code>onPayment</code> method AFTER fire the <code>Transfer</code> event if the receiver is a deployed contract. If the receiver doesn't want to receive this transfer it MUST call <code>ABORT</code>.
--- a/nep-17.mediawiki
+++ b/nep-17.mediawiki
@@ -58,7 +58,7 @@ public static BigInteger balanceOf(byte[] account)
 
 Returns the token balance of the <code>account</code>.
 
-The parameter <code>account</code> SHOULD be a 20-byte address. If not, this method SHOULD <code>throw</code> an exception.
+The parameter <code>account</code> MUST be a 20-byte address. If not, this method SHOULD <code>throw</code> an exception.
 
 If the <code>account</code> is an unused address, this method MUST return <code>0</code>.
 
@@ -70,7 +70,7 @@ public static bool transfer(byte[] from, byte[] to, BigInteger amount)
 
 Transfers an <code>amount</code> of tokens from the <code>from</code> account to the <code>to</code> account.
 
-The parameters <code>from</code> and <code>to</code> SHOULD be 20-byte addresses. If not, this method SHOULD <code>throw</code> an exception.
+The parameters <code>from</code> and <code>to</code> MUST be 20-byte addresses. If not, this method SHOULD <code>throw</code> an exception.
 
 The parameter <code>amount</code> MUST be greater than or equal to <code>0</code>. If not, this method SHOULD <code>throw</code> an exception.
 
@@ -82,7 +82,7 @@ The function SHOULD check whether the <code>from</code> address equals the calle
 
 If the <code>to</code> address is a deployed contract, the function SHOULD check the <code>payable</code> flag of this contract to decide whether it should transfer the tokens to this contract.
 
-If the transfer is not processed, the function SHOULD return <code>false</code>.
+If the transfer is not processed, the function MUST return <code>false</code>.
 
 The function MUST call <code>onPayment</code> method AFTER fire the <code>Transfer</code> event if the receiver is a deployed contract. If the receiver doesn't want to receive this transfer it MUST call <code>ABORT</code>.
 
--- a/nep-17.mediawiki
+++ b/nep-17.mediawiki
@@ -25,7 +25,11 @@ In the method definitions below, we provide both the definitions of the function
 ====symbol====
 
 <pre>
-public static string symbol()
+{
+	"name": "symbol",
+	"parameters": [],
+	"returntype": "String"
+}
 </pre>
 
 Returns a short string representing symbol of the token managed in this contract. e.g. <code>"MYT"</code>. This string MUST be valid ASCII, MUST NOT contain whitespace or control characters, SHOULD be limited to uppercase Latin alphabet (i.e. the 26 letters used in English) and SHOULD be short (3-8 characters is recommended).
@@ -35,7 +39,11 @@ This method MUST always return the same value every time it is invoked.
 ====decimals====
 
 <pre>
-public static byte decimals()
+{
+	"name": "decimals",
+	"parameters": [],
+	"returntype": "Integer"
+}
 </pre>
 
 Returns the number of decimals used by the token - e.g. <code>8</code>, means to divide the token amount by <code>100,000,000</code> to get its user representation.
@@ -45,7 +53,11 @@ This method MUST always return the same value every time it is invoked.
 ====totalSupply====
 
 <pre>
-public static BigInteger totalSupply()
+{
+	"name": "totalSupply",
+	"parameters": [],
+	"returntype": "Integer"
+}
 </pre>
 
 Returns the total token supply deployed in the system.
@@ -53,7 +65,16 @@ Returns the total token supply deployed in the system.
 ====balanceOf====
 
 <pre>
-public static BigInteger balanceOf(byte[] account)
+{
+	"name": "balanceOf",
+	"parameters": [
+		{
+			"name": "account",
+			"type": "Hash160"
+		}
+	],
+	"returntype": "Integer"
+}
 </pre>
 
 Returns the token balance of the <code>account</code>.
@@ -65,7 +86,24 @@ If the <code>account</code> is an unused address, this method MUST return <code>
 ====transfer====
 
 <pre>
-public static bool transfer(byte[] from, byte[] to, BigInteger amount)
+{
+	"name": "transfer",
+	"parameters": [
+		{
+			"name": "from",
+			"type": "Hash160"
+		},
+		{
+			"name": "to",
+			"type": "Hash160"
+		},
+		{
+			"name": "amount",
+			"type": "Integer"
+		}
+	],
+	"returntype": "Boolean"
+}
 </pre>
 
 Transfers an <code>amount</code> of tokens from the <code>from</code> account to the <code>to</code> account.
@@ -86,6 +124,16 @@ If the receiver is a deployed contract, the function MUST call <code>onPayment</
 
 <pre>
 public static void onPayment(BigInteger amount)
+{
+	"name": "onPayment",
+	"parameters": [
+		{
+			"name": "amount",
+			"type": "Integer"
+		}
+	],
+	"returntype": "Void"
+}
 </pre>
 
 ===Events===
@@ -93,7 +141,23 @@ public static void onPayment(BigInteger amount)
 ====Transfer====
 
 <pre>
-public static event Transfer(byte[] from, byte[] to, BigInteger amount)
+{
+	"name": "Transfer",
+	"parameters": [
+		{
+			"name": "from",
+			"type": "Hash160"
+		},
+		{
+			"name": "to",
+			"type": "Hash160"
+		},
+		{
+			"name": "amount",
+			"type": "Integer"
+		}
+	]
+}
 </pre>
 
 MUST trigger when tokens are transferred, including zero value transfers and self-transfers.
--- a/nep-17.mediawiki
+++ b/nep-17.mediawiki
@@ -1,15 +1,16 @@
 <pre>
-  NEP: 5
+  NEP: 17
   Title: Token Standard
-  Author: Tyler Adams <lllwvlvwlll@gmail.com>, luodanwg <luodan.wg@gmail.com>, tanyuan <tanyuan666@gmail.com>, Alan Fong <anfn@umich.edu>
+  Author: Erik Zhang <erik@neo.org>
   Type: Standard
-  Status: Final
-  Created: 2017-08-10
+  Status: Draft
+  Created: 2020-11-12
+  Replaces: 5
 </pre>
 
 ==Abstract==
 
-The NEP-5 Proposal outlines a token standard for the NEO blockchain that will provide systems with a generalized interaction mechanism for tokenized Smart Contracts.  This mechanic, along with the justification for each feature are defined.  A template and examples are also provided to enable the development community.
+This proposal outlines a token standard for the NEO blockchain that will provide systems with a generalized interaction mechanism for tokenized Smart Contracts.  This mechanic, along with the justification for each feature are defined.  A template and examples are also provided to enable the development community.
 
 ==Motivation==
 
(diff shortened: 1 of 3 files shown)


erikzhang:
@shargon @ixje @roman-khimov @vncoelho @EdgeDLT @igormcoelho @neo-project/ngd-shanghai

nep-17.mediawiki
```diff
@@ -0,0 +1,109 @@
...
+
+The function MUST return <code>false</code> if the <code>from</code> account balance does not have enough tokens to spend.
+
+If the method succeeds, it MUST fire the <code>transfer</code> event, and MUST return <code>true</code>, even if the <code>amount</code> is <code>0</code>, or <code>from</code> and <code>to</code> are the same address.
+
+The function SHOULD check whether the <code>from</code> address equals the caller contract hash. If so, the transfer SHOULD be processed; If not, the function SHOULD use the SYSCALL <code>Neo.Runtime.CheckWitness</code> to verify the transfer.
+
+If the <code>to</code> address is a deployed contract, the function SHOULD check the <code>payable</code> flag of this contract to decide whether it should transfer the tokens to this contract.
+
+If the transfer is not processed, the function SHOULD return <code>false</code>.
```
ixje, line 85: There are some optional behaviours (SHOULDs) where I wonder if they should be MUSTs instead
1. `from` and `to` SHOULD be 20-byte addresses and **SHOULD** throw an exception if not. I think the latter should be a MUST. Right now, if you pass a stack item with less then 20-bytes the ApplicationEngine will throw an exception because it can't convert the data to a `UInt160`. 
2. We say "if `transfer` succeeds it **MUST** return `true`", but at the same time we say "if transfer is not processed it **SHOULD** return `false`". Under the assumption that "not processed == transfer failed", then I think the latter should say "**MUST** return `false`"

ixje (commented):
First off, happy to see this new NEP 👍 I think an extension to NEP5 could have done, but completely superseding is fine with me as well.

Note that there is a downside to removing the `name` function for light wallets. Currently a light wallet can create a single script and use a single `invokescript` RPC call to fetch the name, symbol, decimals and totalSupply.  Changing this requires another RPC call (`getcontractstate`) and fetch the name from there (I'm guessing from the `extra` field, of which the content aren't standardised)

shargon (commented):
Add `transferFrom` and `allowance`?

nep-17.mediawiki
```diff
@@ -0,0 +1,109 @@
+<pre>
+  NEP: 17
```
vncoelho, line 2: Since we are on the 17, maybe we should jump 3 and make NEP-20. aheuaheua

nep-17.mediawiki
```diff
@@ -0,0 +1,109 @@
...
+
+If the method succeeds, it MUST fire the <code>transfer</code> event, and MUST return <code>true</code>, even if the <code>amount</code> is <code>0</code>, or <code>from</code> and <code>to</code> are the same address.
+
+The function SHOULD check whether the <code>from</code> address equals the caller contract hash. If so, the transfer SHOULD be processed; If not, the function SHOULD use the SYSCALL <code>Neo.Runtime.CheckWitness</code> to verify the transfer.
+
+If the <code>to</code> address is a deployed contract, the function SHOULD check the <code>payable</code> flag of this contract to decide whether it should transfer the tokens to this contract.
+
+If the transfer is not processed, the function SHOULD return <code>false</code>.
+
+The function MUST call <code>onPayment</code> method AFTER fire the <code>Transfer</code> event if the receiver is a deployed contract. If the receiver doesn't want to receive this transfer it MUST call <code>ABORT</code>.
```
vncoelho, line 87: I think that this part "If the receiver doesn't want to receive this transfer it MUST call <code>ABORT</code>." should be moved to the <code>payable</code> part a little bit above.

It could be: The function MUST call <code>onPayment</code> method AFTER fire the <code>Transfer</code> event if the receiver is a deployed contract and <code>payable</code>  was enabled.

nep-17.mediawiki
```diff
@@ -0,0 +1,109 @@
...
+
+===Events===
+
+====Transfer====
+
+<pre>
+public static event Transfer(byte[] from, byte[] to, BigInteger amount)
+</pre>
+
+MUST trigger when tokens are transferred, including zero value transfers.
```
vncoelho, line 101: MUST trigger when tokens are transferred, including zero value transfers.
-> 
MUST trigger when tokens are transferred, including zero value transfers **and self-transfers**.

vncoelho:
@shargon, what would be `transferFrom` and `allowance`?

ixje:
> @shargon, what would be `transferFrom` and `allowance`?

you `allow` another account to transfer from an account you own. The other account uses `transferFrom`

nep-17.mediawiki
erikzhang: Fixed.

nep-17.mediawiki
EdgeDLT: @vncoelho I thought so too... especially if we are adding approve/transferFrom!

erikzhang:
> Add transferFrom and allowance?

With `approve` and `transferFrom`, you need two steps. In Neo3, we don't need them. With `transfer` and `onPayment`, you only need one step.

nep-17.mediawiki
erikzhang: It's meaningless.

nep-17.mediawiki
erikzhang: Updated.

nep-17.mediawiki
```diff
@@ -0,0 +1,107 @@
...
+
+The function MUST return <code>false</code> if the <code>from</code> account balance does not have enough tokens to spend.
+
+If the method succeeds, it MUST fire the <code>transfer</code> event, and MUST return <code>true</code>, even if the <code>amount</code> is <code>0</code>, or <code>from</code> and <code>to</code> are the same address.
+
+The function SHOULD check whether the <code>from</code> address equals the caller contract hash. If so, the transfer SHOULD be processed; If not, the function SHOULD use the SYSCALL <code>Neo.Runtime.CheckWitness</code> to verify the transfer.
+
+If the transfer is not processed, the function MUST return <code>false</code>.
+
+The function MUST call <code>onPayment</code> method AFTER fire the <code>Transfer</code> event if the receiver is a deployed contract. If the receiver doesn't want to receive this transfer it MUST call <code>ABORT</code>.
```
vncoelho, line 85: The function MUST call <code>onPayment</code>.
Are we mentioning the `onPayment` of the current contract and the `onPayment` dynamic invoked by the deployed contract which is receiver? Perhaps both?

erikzhang: The deployed contract (which is receiver).

vncoelho: Maybe modify to something in this direction:

The function MUST call <code>onPayment</code> method AFTER fire the <code>Transfer</code> event if the receiver is a deployed contract. If the receiver doesn't want to receive this transfer it MUST call <code>ABORT</code>.
-> 
The function MUST call deployed contract  <code>onPayment</code> method AFTER fire the <code>Transfer</code> event. If the receiver (<code>to</code>) doesn't want to receive this transfer, or did not implement the <code>onPayment</code> method, it MUST call <code>ABORT</code>. Note that this is dynamic invoke is not needed for standard addressees.

nep-17.mediawiki
vncoelho: That is true. It would be a good remembering for the general public and is a little bit didactic...aheuahea
Anyway, it is meaningless and would leave here with 3 empty slots.

nep-17.mediawiki
```diff
@@ -0,0 +1,107 @@
...
+  Author: Erik Zhang <erik@neo.org>
+  Type: Standard
+  Status: Draft
+  Created: 2020-11-12
+  Replaces: 5
+</pre>
+
+==Abstract==
+
+This proposal outlines a token standard for the NEO blockchain that will provide systems with a generalized interaction mechanism for tokenized Smart Contracts.  This mechanic, along with the justification for each feature are defined.  A template and examples are also provided to enable the development community.
```
igormcoelho, line 13: It would be interesting to also invite previous authors of NEP-5 to join this NEP-17 for discussion, or even, authorship (if they agree too).

igormcoelho: P.S.: just noticed you informed that on the TODO list, requiring people to submit Pull Request. That's fine then.

vncoelho: Ping @lllwvlvwlll to discuss with other authors of NEP-5.

Tommo-L:
> Removed name method, because it should be in manifest for all the contracts, not only for the token contracts.

Then we'd better add `author`, `description`,  `email`,  `website` in manifest file, too

erikzhang:
> Then we'd better add author, description, email, website in manifest file, too

They can be in `Extra`.

nep-17.mediawiki
```diff
@@ -0,0 +1,107 @@
...
+
+====balanceOf====
+
+<pre>
+public static BigInteger balanceOf(byte[] account)
+</pre>
+
+Returns the token balance of the <code>account</code>.
+
+The parameter <code>account</code> MUST be a 20-byte address. If not, this method SHOULD <code>throw</code> an exception.
```
igormcoelho, line 61: "20-byte address" -> "20-byte scripthash"? Maybe we can have more precision here? I know concepts are equivalent, but many people actually think that a Base58 String is passed here, which is not the case... Other possibility is to adopt UInt160 standard here, since we are dealing with C# devpack as reference, correct?

nep-17.mediawiki
igormcoelho, line 85:
```suggestion
If the receiver is a deployed contract, the function MUST call <code>onPayment</code> method on receiver contract AFTER firing the <code>Transfer</code> event. If the receiver doesn't want to receive this transfer it MUST call <code>ABORT</code>.
```

igormcoelho: I would also add: "If onPayment method does not exist on receiver contract, no action is required."

nep-17.mediawiki
```diff
@@ -0,0 +1,107 @@
...
+The function MUST return <code>false</code> if the <code>from</code> account balance does not have enough tokens to spend.
+
+If the method succeeds, it MUST fire the <code>transfer</code> event, and MUST return <code>true</code>, even if the <code>amount</code> is <code>0</code>, or <code>from</code> and <code>to</code> are the same address.
+
+The function SHOULD check whether the <code>from</code> address equals the caller contract hash. If so, the transfer SHOULD be processed; If not, the function SHOULD use the SYSCALL <code>Neo.Runtime.CheckWitness</code> to verify the transfer.
+
+If the transfer is not processed, the function MUST return <code>false</code>.
+
+The function MUST call <code>onPayment</code> method AFTER fire the <code>Transfer</code> event if the receiver is a deployed contract. If the receiver doesn't want to receive this transfer it MUST call <code>ABORT</code>.
+
```
igormcoelho, line 86:
```suggestion
The method <code>onPayment</code> should receive the amount as parameter (this is expected only at receiving contracts, not on the NEP-17 token contract):
```

nep-17.mediawiki
```diff
@@ -0,0 +1,107 @@
...
+====Transfer====
+
+<pre>
+public static event Transfer(byte[] from, byte[] to, BigInteger amount)
+</pre>
+
+MUST trigger when tokens are transferred, including zero value transfers and self-transfers.
+
+A token contract which creates new tokens MUST trigger a <code>Transfer</code> event with the <code>from</code> address set to <code>null</code> when tokens are created.
```
igormcoelho, line 101: Does <code>null</code> exists as <code>from</code>? Is it the same as an empty byte array / buffer?

nep-17.mediawiki
roman-khimov, line 61: BTW, specifying methods via NEP-14 descriptors allows to use `Hash160`. And neo-project/neo#1891 would even allow to remove this check.

nep-17.mediawiki
```diff
@@ -0,0 +1,107 @@
...
+
+Transfers an <code>amount</code> of tokens from the <code>from</code> account to the <code>to</code> account.
+
+The parameters <code>from</code> and <code>to</code> MUST be 20-byte addresses. If not, this method SHOULD <code>throw</code> an exception.
+
+The parameter <code>amount</code> MUST be greater than or equal to <code>0</code>. If not, this method SHOULD <code>throw</code> an exception.
+
+The function MUST return <code>false</code> if the <code>from</code> account balance does not have enough tokens to spend.
+
+If the method succeeds, it MUST fire the <code>transfer</code> event, and MUST return <code>true</code>, even if the <code>amount</code> is <code>0</code>, or <code>from</code> and <code>to</code> are the same address.
```
roman-khimov, line 79: Either it's `transfer` everywhere or `Transfer` everywhere.

nep-17.mediawiki
```diff
@@ -0,0 +1,107 @@
...
+
+The function SHOULD check whether the <code>from</code> address equals the caller contract hash. If so, the transfer SHOULD be processed; If not, the function SHOULD use the SYSCALL <code>Neo.Runtime.CheckWitness</code> to verify the transfer.
```
roman-khimov, line 81: Do we really need caller contract hash check after neo-project/neo#1925?

nep-17.mediawiki
roman-khimov, line 85: I think we can leave receiver behavior (and `onPayment` definition) for another NEP (because the receiver doesn't have to support full NEP-17), here it's important to call `onPayment`, but it's not clear from this document how a contract should detect `if the receiver is a deployed contract`.

nep-17.mediawiki
```diff
@@ -0,0 +1,107 @@
...
+<pre>
+public static void onPayment(BigInteger amount)
+</pre>
+
+===Events===
+
+====Transfer====
+
+<pre>
+public static event Transfer(byte[] from, byte[] to, BigInteger amount)
```
roman-khimov, line 96:
```
         {
            "parameters" : [
               {
                  "type" : "ByteArray",
                  "name" : "from"
               },
               {
                  "name" : "to",
                  "type" : "ByteArray"
               },
               {
                  "name" : "amount",
                  "type" : "Integer"
               }
            ],
            "name" : "transfer"
         }
```

nep-17.mediawiki
roman-khimov, line 101: It'd be better to have `NULL` here for VM `NULL`.

nep-17.mediawiki
roman-khimov: I'd rather fail if `onPayment` does not exist. `onPayment` is mandatory for contracts that want to receive/own/spend tokens IMO.

vncoelho: This would make deployed need to update its rules over time or make always true. 
I am not sure about make mandatory.

vncoelho: I am thinking again here, the default will be Abort anyway,right?
Thus if not implemented it will not accept tokens anyway.

igormcoelho: > I'd rather fail if onPayment does not exist. onPayment is mandatory for contracts that want to receive/own/spend tokens IMO.

Looks like a circular/reflexive argument to me... 
- we fail if method doesn't exist, because onPayment is mandatory to receive tokens.. and since it's mandatory, we should fail if it doesn't exist :) 
- We could not fail if it doesn't exist, and not make mandatory the existance of onPayment to receive tokens.. and since it's not mandatory, we are obliged to fail if it doesn't exist (opposite logic)

As long as this is clear, we should write on text anyway, the preferred logic.

I think this is a very important matter, regarding future standards, that will certainly mimic this. A deployed contract released today should be unable to receive some token kind developed in 5 years from now? Example: NFT will have its own "onPayment" strategy, so if it's not predicted on contract, it will not be able to receive it.

Some more general "strategy" could be a method "allowReceivingAssets" or typical "payable" flag. Maybe, the sense of "payable" is still valid, and we should require contracts to be "payable" (in manifest) and call NEP-5 "onPayment" following its own structure (for NFT it's different, and so on). Just to make sure that if we follow this logic here, we will block any future developed token kinds to be received by any contract developed in this moment (something that a complementary strategy of "payable=true" would resolve for any future "token kinds").

igormcoelho: A more flexible structure to me for "payable assets" on general, seems to be:
- contract wants to send assets (NEP-5, NFT, ...) to another contract
- if contract is not deployed, it will accept
- if contract is deployed, check if it's payable
- if it's not payable, reject.
- if it's payable, send assets and call specific action method (for NEP-5 is "onPayment")

nep-17.mediawiki
```diff
@@ -0,0 +1,107 @@
...
+
+===Methods===
+
+====symbol====
+
+<pre>
+public static string symbol()
+</pre>
+
+Returns a short string symbol of the token managed in this contract. e.g. <code>"MYT"</code>. This symbol SHOULD be short (3-8 characters is recommended), with no whitespace characters or new-lines and SHOULD be limited to the uppercase latin alphabet (i.e. the 26 letters used in English).
```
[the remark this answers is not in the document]
igormcoelho: I agree with you @roman-khimov... maybe we (and all wallet developers) count with "common sense" here, but it may come some individual with token symbol of zero chars, or even 10000 chars. Software interfaces may not be prepared for that...
I would put some strict checking on at least one char, and some strict maximum length (maybe 8).

nep-17.mediawiki
erikzhang: The method signature of `onPayment` is below.

nep-17.mediawiki
erikzhang: https://github.com/neo-project/neo-vm/blob/f88da9fd63fdae2d57507385cef9497727a3c2db/src/neo-vm/Types/StackItem.cs#L14

nep-17.mediawiki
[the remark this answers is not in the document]
erikzhang: Updated.

nep-17.mediawiki
```diff
@@ -0,0 +1,107 @@
...
+==Specification==
+
+In the method definitions below, we provide both the definitions of the functions as they are defined in the contract as well as the invoke parameters.
+
+===Methods===
+
+====symbol====
+
+<pre>
+public static string symbol()
```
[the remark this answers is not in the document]
erikzhang: Updated.

nep-17.mediawiki
```diff
@@ -0,0 +1,107 @@
...
+</pre>
+
+Returns a short string symbol of the token managed in this contract. e.g. <code>"MYT"</code>. This symbol SHOULD be short (3-8 characters is recommended), with no whitespace characters or new-lines and SHOULD be limited to the uppercase latin alphabet (i.e. the 26 letters used in English).
+
+This method MUST always return the same value every time it is invoked.
+
```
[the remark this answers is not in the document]
erikzhang: What's the difference?

nep-17.mediawiki
```diff
@@ -0,0 +1,107 @@
...
+====decimals====
+
+<pre>
+public static byte decimals()
+</pre>
+
+Returns the number of decimals used by the token - e.g. <code>8</code>, means to divide the token amount by <code>100,000,000</code> to get its user representation.
```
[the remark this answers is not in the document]
erikzhang: The NeoVM limit integers to 256-bits.

nep-17.mediawiki
erikzhang: I have modified it and used `Hash160` in the ABI description.

nep-17.mediawiki
erikzhang: Fixed.

nep-17.mediawiki
erikzhang: Although we provide a simple implementation method, we still have to explain the logic clearly in the proposal.

nep-17.mediawiki
erikzhang: We only need to explain the logic clearly in the proposal. As for how to implement it, there can be templates or examples. The specific way is to call `System.Blockchain.GetContract`, and then check the return value.

nep-17.mediawiki
erikzhang: What do you mean? This is vm `null`.

nep-17.mediawiki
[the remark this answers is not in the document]
roman-khimov: Explicitly specifying behavior for updated contracts. It's not to replace the sentence above, but to add one.

nep-17.mediawiki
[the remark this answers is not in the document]
roman-khimov: True, but people are interacting with the node via JSON RPC and while it technically doesn't limit numbers in any way, in practice there is `MAX_SAFE_INTEGER` in the C# code. I'd say it's too restrictive (although it follows RFC 7159 recommendation) because it doesn't allow even for 100M with 8 decimals (which is easily reached for GAS), but at the same time even if we're to remove this limitation (it's not hard to do that), probably we'll have client compatibility problems with anything not fitting in at least int64 limitations.

nep-17.mediawiki
roman-khimov: I think this will lead to contracts implementing this check while they no longer really need to do that. It's something to be documented for `Neo.Runtime.CheckWitness`, here we can rely on its behavior.

nep-17.mediawiki
roman-khimov: Just nit-picking, but in VM it's either referred to as `NULL` (`PUSHNULL`/`ISNULL`) or `Null` (as specific VM type), not `null`.

nep-17.mediawiki
roman-khimov: `System.Blockchain.GetContract` seems to be too heavy for the task to me (because we don't really need contract's data), but maybe it's OK (0.01 GAS, same as notification, and it's not hard to `DROP` the result). And you're right, if we're resorting to these standard APIs they can be omitted from the specification.

What do you think about making an additional companion NEP (NEP-18, for example) specifically for `onPayment`? It can simplify things in that this NEP will refer to it for `onPayment` behavior definition and contracts could then easily declare that they're supporting NEP-18 (meaning that they have proper `onPayment` and can process incoming tokens). For example, NEO and GAS contracts are NEP-17-compatible, but not NEP-18 (they don't have `onPayment`), while NeoFS mainnet contract is the opposite, it's not NEP-17 (it doesn't declare any token of its own), but it does support NEP-18 (can accept payments to it).

vncoelho: I think it is a good idea, @roman-khimov.

nep-17.mediawiki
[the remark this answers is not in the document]
vncoelho: I think the decimals can be changed after upgrade @roman-khimov. Maybe a logic can be done to migrate balances.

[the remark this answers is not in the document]
roman-khimov: It technically can be done (especially given `_deploy` powers), but doing so will cause too much confusion and errors because the old value (considered to be stable) can be cached in lots of places.

nep-17.mediawiki
igormcoelho: Maybe we refer to "NeoVM Null" or "Null StackItem"?

nep-17.mediawiki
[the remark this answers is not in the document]
igormcoelho: I tend to agree with @roman-khimov, but I've seen many people using 16 already... so I agree we may put some limit, but the point is: based on what?
int64 is a lot of numbers, but it's not clear to me if a hyperinflationary coin could beat that... is there any risk for that?

nep-17.mediawiki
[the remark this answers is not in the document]
igormcoelho: I think  it doesn't help the standard @roman-khimov because migration can change any aspect of the contract, including token symbol, decimals, checks for origin, total balance, everything. And, in the end, if anything goes against the standard after the migration, it means that the previous contract was NEP-5 and after migration its not, simple as that.  If one manages to change token symbol, even without migration, then it has also offended the standard, and so on.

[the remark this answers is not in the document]
erikzhang: Maybe we will not change this for now.

nep-17.mediawiki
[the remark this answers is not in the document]
erikzhang: Maybe we should optimize RpcServer to recognize the number represented by the string. But there is no need to modify the standard.

nep-17.mediawiki
erikzhang: They are the same.

nep-17.mediawiki
erikzhang: > What do you think about making an additional companion NEP (NEP-18, for example) specifically for onPayment?

I don't think it's necessary. We directly check whether it contains the `onPayment` method, without checking whether it implements a certain NEP.

nep-17.mediawiki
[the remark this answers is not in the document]
roman-khimov: The main problem is that it's not just about RpcServer (as I've said, it's not hard to fix that), _every other client_ (or server) needs to support that too and I'm not sure we can expect that to happen.

shargon (approved):

nep-17.mediawiki
```diff
@@ -0,0 +1,171 @@
...
+
+<pre>
+public static void onPayment(BigInteger amount)
+{
+  "name": "onPayment",
+  "parameters": [
+    {
+      "name": "amount",
+      "type": "Integer"
+    }
```
roman-khimov, line 133: Maybe I've missed some discussion, but don't we also need `from` parameter here to avoid dealing with notifications in the contract called?

Qiao-Jin: `from` can be got from `ExecutionEngine.CallingScriptHash` if needed though

nep-17.mediawiki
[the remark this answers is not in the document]
erikzhang: Is there any suggestion? I don't think it's easy to limit it here.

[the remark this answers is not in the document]
roman-khimov: I doubt anyone goes beyond 8 here, but 12 seems to be viable (picocoins) and that's what we're using for internal NeoFS calculations now (after failing with 16). As a reference, Neo 2 UTXO-based tokens were limited to 8 here with int64 values.

So, 12 here and int64 limits for balance/totalsupply, seems fine to me. It also allows for some optimizations in native tokens replacing big integer calculations with int64.

nep-17.mediawiki
roman-khimov: A ("from") transfers X amount ("amount") of TOK to CONTRACT ("to"). The invocation stack will look like this:
Entry script -> TOK script ("transfer") -> CONTRACT ("onPayment")

In CONTRACT's `onPayment` I can easily get X (from the `amount` parameter), I can easily get TOK (`CallingScriptHash`), I know "to" (`GetExecutingScriptHash`). Where do I get "from" from?

Qiao-Jin: Understood. Indeed we need some kind of input to show where the token is transferred from.

erikzhang: Fixed.

nep-17.mediawiki
```diff
@@ -0,0 +1,175 @@
...
+If the method succeeds, it MUST fire the <code>Transfer</code> event, and MUST return <code>true</code>, even if the <code>amount</code> is <code>0</code>, or <code>from</code> and <code>to</code> are the same address.
+
+The function SHOULD check whether the <code>from</code> address equals the caller contract hash. If so, the transfer SHOULD be processed; If not, the function SHOULD use the SYSCALL <code>Neo.Runtime.CheckWitness</code> to verify the transfer.
+
+If the transfer is not processed, the function MUST return <code>false</code>.
+
+If the receiver is a deployed contract, the function MUST call <code>onPayment</code> method on receiver contract AFTER firing the <code>Transfer</code> event. If the receiver doesn't want to receive this transfer it MUST call <code>ABORT</code>.
+
+<pre>
+public static void onPayment(BigInteger amount)
```
Qiao-Jin, line 126: This line should also be refreshed

erikzhang: Removed.

shargon (approved):

nep-17.mediawiki
```diff
@@ -0,0 +1,174 @@
...
+    },
+    {
+      "name": "amount",
+      "type": "Integer"
+    }
+  ]
+}
+</pre>
+
+MUST trigger when tokens are transferred, including zero value transfers and self-transfers.
```
fyrchik, line 166: `be triggered` or `be emitted`?

nep-17.mediawiki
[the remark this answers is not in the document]
igormcoelho: @roman-khimov What I've seen regarding precision, I guess guys from DEX (maybe Nash) require that, is that 16 precision is mostly necessary in many scenarios (the reason is connected to the same principle of "gas distribution" that also made very difficult to transform NEO into fractional units). So, 16 precision is indeed necessary, but I've never seen more than that. Regarding the total supply, I do not imagine an upper limit on that, but maybe there's a "reasonable number".. I don't know.

nep-17.mediawiki
erikzhang: I don't feel much difference.

erikzhang:
We wait one more week to see if there are more comments.

nep-17.mediawiki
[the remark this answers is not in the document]
roman-khimov: After some additional research I'd agree that 16 really is in use on other chains and we shouldn't place any limits in this standard. This still leaves us with JSONization problem, though.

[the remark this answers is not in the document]
roman-khimov: https://github.com/neo-project/neo/issues/2082

nep-17.mediawiki
```diff
@@ -0,0 +1,174 @@
...
+
+The function SHOULD check whether the <code>from</code> address equals the caller contract hash. If so, the transfer SHOULD be processed; If not, the function SHOULD use the SYSCALL <code>Neo.Runtime.CheckWitness</code> to verify the transfer.
+
+If the transfer is not processed, the function MUST return <code>false</code>.
+
+If the receiver is a deployed contract, the function MUST call <code>onPayment</code> method on receiver contract AFTER firing the <code>Transfer</code> event. If the receiver doesn't want to receive this transfer it MUST call <code>ABORT</code>.
+
+<pre>
+{
+  "name": "onPayment",
```
fyrchik, line 127: What about asset type? How can we distinguish between receiving NEO or GAS ?

Tommo-L: asset's hash = `CallingScriptHash`

lllwvlvwlll:
I know this is probably a little late in the game, but we should consider using a different acronym for Neo3 proposals since this is not specific proposal is not compatible for Neo2 and depending on the level of continued Neo2 support, we will likely end up in a situation with interleaved proposals for both Neo2 and Neo3 which could be very confusing for developers.

nep-17.mediawiki
```diff
@@ -0,0 +1,182 @@
...
+  ],
+  "returntype": "Boolean"
+}
+</pre>
+
+Transfers an <code>amount</code> of tokens from the <code>from</code> account to the <code>to</code> account.
+
+The parameters <code>from</code> and <code>to</code> MUST be 20-byte addresses. If not, this method SHOULD <code>throw</code> an exception.
+
+The parameter <code>amount</code> MUST be greater than or equal to <code>0</code>. If not, this method SHOULD <code>throw</code> an exception.
```
Qiao-Jin, line 117: Info about parameter data is missing

erikzhang: > If the receiver is a deployed contract, the function MUST call onPayment method on receiver contract with the data parameter from transfer AFTER firing the Transfer event. If the receiver doesn't want to receive this transfer it MUST call ABORT.

Qiao-Jin: I mean, whether we need to add a brief description like why we need this parameter, what's the size limitation, etc.

nep-17.mediawiki
```diff
@@ -0,0 +1,182 @@
...
+      "type": "Hash160"
+    },
+    {
+      "name": "to",
+      "type": "Hash160"
+    },
+    {
+      "name": "amount",
+      "type": "Integer"
+    }
```
shargon, line 169: Add `data` in the event?

erikzhang: `data` can be any `StackItem`, so it is not suitable for notifications.

lllwvlvwlll:
Boa Reference: https://github.com/CityOfZion/neo3-boa/blob/development/boa3_test/examples/NEP17.py

nep-17.mediawiki
```diff
@@ -0,0 +1,184 @@
...
+==Motivation==
+
+As the NEO blockchain scales, Smart Contract deployment and invocation will become increasingly important.  Without a standard interaction method, systems will be required to maintain a unique API for each contract, regardless of their similarity to other contracts.  Tokenized contracts present themselves as a prime example of this need because their basic operating mechanism is the same.  A standard method for interacting with these tokens relieves the entire ecosystem from maintaining a definition for basic operations that are required by every Smart Contract that employs a token.
+
+==Specification==
+
+In the method definitions below, we provide both the definitions of the functions as they are defined in the contract as well as the invoke parameters.
+
+===Methods===
+
```
superboyiii, line 24: Should we add Verify() here?
```suggestion
====verify====

<pre>
{
  "name": "verify",
  "parameters": [],
  "returntype": "Boolean"
}
</pre>
```

erikzhang: Of course not.

chenzhitong:
Should we add the `onPayment` method to the standard methods of NEP-17 or C# smart contract template?

erikzhang:
@chenzhitong `onPayment` is a receiver method.

chenzhitong:
If a receiver doesn't implement the `onPayment` method, will the `transfer` execute fail?
接收者不实现`onPayment` ，转账是否会报错？

In addition, should the manifest of the NEP-7 method must be written 
```json
"permissions": [
  {
    "contract": "*",  
    "methods": "*"
  }
]
```
, because it needs to call the `onPayment` method of `to`.
是不是所有的NEP-17方法都要在manifest中添加上述的permission，因为它需要调用to的`onPayment`方法
@erikzhang

erikzhang:
```
{
  "contract": "*",  
  "methods": "onPayment"
}
```

chenzhitong:
那一个合约（ICO合约）想接收别人给他转账，是不是要在 manifest 中这么写
`“trusts”: *`
这样岂不是很不安全。
@erikzhang

erikzhang:
> "trusts": *

For example, if the ICO contract needs to receive neo, he has to trust neo. I don't think an ICO can accept any type of tokens.

satoshichou:
When will it be merged to master. The proposal seems is already implemented in native ocntract.

erikzhang:
Merge?

shargon (dismissed):

superboyiii (dismissed):
I think it's OK to merge.

RichardBelsum (dismissed):

RichardBelsum:
Celebrate🎉

PR #129, opened 2020-12-07
add safe attribute

bettybao1209:


ixje:
Relevant https://github.com/neo-project/neo/issues/1664
Doesn't matter if it is in the ABI or manifest, it should not be decided by the contract author

ixje:
https://github.com/neo-project/neo/pull/2120#discussion_r537215041

This seems to be the more relevant description

erikzhang (approved):

shargon (approved):

PR #130, opened 2021-01-14
NEP11: Non-fungible Token Standard

erikzhang:
Originally posted by @hal0x2328 in #41.


nep-11.mediawiki
```diff
@@ -0,0 +1,251 @@
...
+<pre>
+public static BigInteger totalSupply()
+</pre>
+
+Returns the total token supply currently in the system.
+
+====balanceOf====
+
+<pre>
+public static BigInteger balanceOf(byte[] owner)
```
vncoelho, line 63: This `balanceOf` is a little bit strange.
What is this total amount of NFTs? Since we do not describe the tokensID this info may become misleading.

nep-11.mediawiki
```diff
@@ -0,0 +1,251 @@
...
+
+The parameter <code>tokenid</code> SHOULD be a valid NFT.
+If not, this method SHOULD <code>throw</code> an exception.
+
+===Divisible NFT methods===
+
+====transfer====
+
+<pre>
+public static bool transfer(byte[] from, byte[] to, BigInteger amount, byte[] tokenid)
```
vncoelho, line 183: Why the divisible has `From` and non-divisible has not?

nep-11.mediawiki
erikzhang: See https://github.com/ethereum/EIPs/blob/master/EIPS/eip-721.md#specification

nep-11.mediawiki
```diff
@@ -0,0 +1,259 @@
...
+If not, this method SHOULD <code>throw</code> an exception.
+
+The function SHOULD return <code>false</code> if the token that will be transferred has more than one owner.
+
+If the method succeeds, it MUST fire the <code>Transfer</code> event, and MUST return <code>true</code>, even if the token is sent to the owner.
+
+The function SHOULD check whether the owner address equals the caller contract hash.
+If so, the transfer SHOULD be processed; If not, the function SHOULD use the SYSCALL <code>Neo.Runtime.CheckWitness</code> to verify the transfer.
+
+If the <code>to</code> address is a deployed contract, the function SHOULD check the <code>payable</code> flag of this contract to decide whether it should transfer the tokens to this contract.
```
roman-khimov, line 105: We no longer have `payable`. Probably we need something similar to NEP-17 `onPayment`.

nep-11.mediawiki
```diff
@@ -0,0 +1,259 @@
...
+
+It transfers the token with id <code>tokenid</code> to address <code>to</code>.
+
+The parameter <code>tokenid</code> SHOULD be a valid NFT.
+If not, this method SHOULD <code>throw</code> an exception.
+
+The parameter <code>to</code> SHOULD be a 20-byte address.
+If not, this method SHOULD <code>throw</code> an exception.
+
+The function SHOULD return <code>false</code> if the token that will be transferred has more than one owner.
```
roman-khimov, line 98: Do we really support multiple owners? And should we support them? I think one owner is enough (and that's what we have in the implementation), multisig scripts can be used if/when needed.

nep-11.mediawiki
```diff
@@ -0,0 +1,259 @@
...
+
+Returns an <code>iterator</code> that contains all the co-owners that own the specified token.
+
+The parameter <code>tokenid</code> SHOULD be a valid NFT.
+If not, this method SHOULD <code>throw</code> an exception.
+
+====balanceOf====
+
+<pre>
+public static BigInteger balanceOf(byte[] owner, byte[] tokenid)
```
roman-khimov, line 164: Don't we also want to support `balanceOf(byte[] owner)` to list token IDs owner possesses?

nep-11.mediawiki
```diff
@@ -0,0 +1,259 @@
...
+If not, this method SHOULD <code>throw</code> an exception.
+
+The function SHOULD return <code>false</code> if the <code>from</code> account balance does not have enough tokens to spend.
+
+If the method succeeds, it MUST fire the <code>Transfer</code> event, and MUST return <code>true</code>, even if the <code>amount</code> is <code>0</code>, or the token is sent to the owner.
+
+The function SHOULD check whether the <code>from</code> address equals the caller contract hash.
+If so, the transfer SHOULD be processed; If not, the function SHOULD use the SYSCALL <code>Neo.Runtime.CheckWitness</code> to verify the transfer.
+
+If the <code>to</code> address is a deployed contract, the function SHOULD check the <code>payable</code> flag of this contract to decide whether it should transfer the tokens to this contract.
```
roman-khimov, line 146: `Payable` here also.

nep-11.mediawiki
```diff
@@ -0,0 +1,259 @@
...
+Returns an <code>iterator</code> that contains all of the tokens minted by the contract.
+
+====properties====
+
+<pre>
+public static string properties(byte[] tokenid)
+</pre>
+
+Returns a serialized NVM object containing the properties for the given NFT.
+The NVM object must conform to the "NEO NFT Metadata JSON Schema".
```
roman-khimov, line 194: I don't think current implementation returns JSON and actually it shouldn't, it's a contract method so it should be some proper VM type like map.

nep-11.mediawiki
```diff
@@ -0,0 +1,259 @@
...
+====tokensOf====
+
+<pre>
+public static iterator tokensOf(byte[] owner)
+</pre>
+
+Returns an <code>iterator</code> that contains all of the token ids owned by the specified address.
+
+The parameter <code>owner</code> SHOULD be a 20-byte address.
+If not, this method SHOULD <code>throw</code> an exception.
```
roman-khimov, line 80: The behavior or `tokensOf()` is not described.

erikzhang: > Returns an <code>iterator</code> that contains all of the token ids owned by the specified address.

nep-11.mediawiki
erikzhang: This is different from multi-signature. The share held by each owner can be different.

nep-11.mediawiki
erikzhang: That is `tokensOf`.

nep-11.mediawiki
gsmachado: I agree with @erikzhang . There are different use cases for multiple owners and multi-sig addresses.

nep-11.mediawiki
erikzhang: It returns a VM object. But the object is described in JSON.

nep-11.mediawiki
erikzhang: Removed.

erikzhang: `onNEP11Payment`? Does it have scenarios? I think it is unlikely that the NFT will be sent to the contract by mistake.

nep-11.mediawiki
erikzhang: Removed.

nep-11.mediawiki
gsmachado: @erikzhang I agree that it's very unlikely that NFT will be sent to the contract by mistake.... but, in my opinion, this is a feature we can't imagine use cases for now, but I'm sure people (contract devs) would like to have that for convenience.

My vote is to add it.

nep-11.mediawiki
```diff
@@ -0,0 +1,259 @@
...
+This allows for high value objects to be tokenized more effectively.
+
+==Specification==
+
+===Common methods===
+
+====symbol====
+
+<pre>
+public static string symbol()
```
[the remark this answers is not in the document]
erikzhang: Done.

nep-11.mediawiki
roman-khimov: Sorry, I've confused parameterless `tokens` (and it is described in the `optional` section) and `tokensOf`.

nep-11.mediawiki
roman-khimov: Well, this `transfer` is in the "non-divisible" section and if we're to support multi-owned non-divisible NFTs some other interfaces might need to be corrected also (like `ownerOf` returning an array or iterator of owners).

nep-11.mediawiki
erikzhang: Added `onNEP11Payment`.

nep-11.mediawiki
erikzhang: Non-divisible NFT couldn't be multi-owned.

roman-khimov: Then this sentence should be removed as non-divisible NFT can't have more than one owner.

erikzhang: I moved it to `Common methods`.

roman-khimov: Divisible NFTs are supposed to implement this variant of `transfer` (as overloaded method) moving 100% of the token if there is only one owner of this token, correct?

erikzhang: Yes.

nep-11.mediawiki
```diff
@@ -0,0 +1,416 @@
...
+      "type": "string",
+      "description": "Identifies the asset to which this NFT represents."
+    },
+    "description": {
+      "type": "string",
+      "description": "Describes the asset to which this NFT represents."
+    },
+    "image": {
+      "type": "string",
+      "description": "Optional. A URI pointing to a resource with mime type image/* representing the asset to which this NFT represents. Consider making any images at a width between 320 and 1080 pixels and aspect ratio between 1.91:1 and 4:5 inclusive."
```
fyrchik, line 323: Instead of `Optional. ` we can use `required` for other fields https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.6.5.3

nep-11.mediawiki
```diff
@@ -0,0 +1,416 @@
...
+<pre>
+{
+  "name": "properties",
+  "parameters": [
+    {
+      "name": "tokenid",
+      "type": "ByteString"
+    }
+  ],
+  "returntype": "Array"
```
fyrchik, line 297: It should be `Map` (as JSON spec declares `"type": "object"`).

nep-11.mediawiki
```diff
@@ -0,0 +1,416 @@
...
+      "type": "string",
+      "description": "Describes the asset to which this NFT represents."
+    },
+    "image": {
+      "type": "string",
+      "description": "Optional. A URI pointing to a resource with mime type image/* representing the asset to which this NFT represents. Consider making any images at a width between 320 and 1080 pixels and aspect ratio between 1.91:1 and 4:5 inclusive."
+    },
+    "tokenURI": {
+      "type": "string",
+      "description": "Optional. A distinct URI for a given asset that adheres to RFC 3986"
```
fyrchik, line 327: Can be `"format": "uri"` (1) if corresponding vocabulary is included (2).
https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.7.3.5
https://github.com/json-schema-org/json-schema-spec/blob/master/schema.json#L10

nep-11.mediawiki
erikzhang: Done.

nep-11.mediawiki
erikzhang: Fixed.

nep-11.mediawiki
erikzhang: Done.

nep-11.mediawiki
```diff
@@ -0,0 +1,418 @@
...
+    {
+      "name": "to",
+      "type": "Hash160"
+    },
+    {
+      "name": "amount",
+      "type": "Integer"
+    },
+    {
+      "name": "tokenid",
```
fyrchik, line 403: We implement some checks in `neo-go` and currently actual parameter name in code (`tokenId` doesn't match `tokenid` here). Can we unify both?

nep-11.mediawiki
```diff
@@ -0,0 +1,418 @@
...
+
+===Optional methods===
+
+====tokens====
+
+<pre>
+{
+  "name": "tokens",
+  "parameters": [],
+  "returntype": "InteropInterface<iterator>"
```
fyrchik, line 280: The type is currently `Any` in C# node, I think it is better to change manifest in C# node, to be more specific.

nep-11.mediawiki
```diff
@@ -0,0 +1,418 @@
...
+<pre>
+{
+  "name": "tokensOf",
+  "parameters": [
+    {
+      "name": "owner",
+      "type": "Hash160"
+    }
+  ],
+  "returntype": "InteropInterface<iterator>"
```
AnnaShaleva, line 103: Same as https://github.com/neo-project/proposals/pull/130/files#r579098628, at the current moment the `returntype` is `Any` in C# node.

AnnaShaleva (changes requested):

nep-11.mediawiki
erikzhang: We can unify them in native contracts. But we have no way to guarantee that other contracts use the same parameter names. And the parameter name is not important, it does not affect anything.

erikzhang: Updated.

nep-11.mediawiki
erikzhang: It should be fixed in neo-project/neo#2356

nep-11.mediawiki
erikzhang: Fixed in neo-project/neo#2356

nep-11.mediawiki
```diff
@@ -0,0 +1,418 @@
...
+    {
+      "name": "owner",
+      "type": "Hash160"
+    }
+  ],
+  "returntype": "Integer"
+}
+</pre>
+
+Returns the total amount of NFTs owned by the specified address.
```
roman-khimov, line 87: What is the supposed return value of this method for divisible NFTs? They have an overloaded `balanceOf` which is quite obvious, but they at the same time have this one also (with `owner` parameter only) and if some owner has 0.5 of token A and 0.125 of token B then what is the total amount of divisible NFTs owned?

erikzhang: `0.625`

roman-khimov: But does it have any meaning? These tokens are non-fungible and it can be like you own 0.5 of some house and 0.125 of the other, you don't own 0.625 of anything.

EdgeDLT:
I think we may benefit from a couple of amendments to the NFT metadata schema. Firstly, not all NFTs are images, they can be any form of content, such as music or a piece of writing. So maybe we should replace the "image"  parameter with something more generalized (maybe "asset" or "content").

Secondly, I think we should include the content's SHA256 hash as a value in the schema. A commonly discussed problem right now in the Ethereum community is that a URI can easily be changed to point to something completely different; currently there are many cases of NFT images that are being replaced with pictures of rugs (for "rug pull"). 

Obviously an NFT isn't really used to actually store the original content so this issue isn't directly a problem for us, but I think it should always be possible to verify what content an NFT should point to. Therefore I think it is very important to store the hash.

erikzhang:
> Firstly, not all NFTs are images, they can be any form of content, such as music or a piece of writing. So maybe we should replace the "image" parameter with something more generalized (maybe "asset" or "content").

`image` is a URI pointing to the icon of the specified NFT. It does not indicate the content. That's why we have another field `tokenURI`.

> Secondly, I think we should include the content's SHA256 hash as a value in the schema.

The NFT metadata schema is expandable. You can add custom fields. Currently described are some well-known fields.

erikzhang:
Merge? Or anyone has any comments?

shargon (dismissed):

WyattMufson (approved):
It's now been more than 3 years since this was started, seems ready to merge

vncoelho (dismissed):

nep-11.mediawiki
```diff
@@ -0,0 +1,442 @@
...
+
+MUST trigger when tokens are transferred, including zero value transfers.
+
+A token contract which creates new tokens MUST trigger a <code>Transfer</code> event with the <code>from</code> address set to <code>null</code> when tokens are created.
+
+A token contract which burns tokens MUST trigger a <code>Transfer</code> event with the <code>to</code> address set to <code>null</code> when tokens are burned.
+
+==Implementation==
+
+https://github.com/neo-project/neo/blob/master/src/neo/SmartContract/Native/NonfungibleToken.cs
```
RichardBelsum, line 442: 404

erikzhang: Fixed.

nep-11.mediawiki
```diff
@@ -0,0 +1,442 @@
...
+{
+  "name": "tokensOf",
+  "safe": true,
+  "parameters": [
+    {
+      "name": "owner",
+      "type": "Hash160"
+    }
+  ],
+  "returntype": "InteropInterface<iterator>"
```
fyrchik, line 108: Can this be implemented by non-native contracts after https://github.com/neo-project/neo/pull/2439 ?

fyrchik: Oh, I see, it's just storage iterator.

metachris:
Please merge! Will be great to have a finalised, accepted and merged standard that the community can build upon.

shargon (dismissed):

WyattMufson (dismissed):

WyattMufson:
@erikzhang what else is needed to merge this is in?

PR #134, opened 2021-03-19
Authentication Scheme

erikzhang:


igormcoelho:
Amazing path @erikzhang ! It was less than a month ago that @vncoelho and I got a discussion on the matter. Structure looks nice, it would be interesting to provide some example perhaps, with Alice and Bob, and maybe sone contextualization over existing OAuth 2. Neo can certainly be such pioneer, specially with oracle technology.

erikzhang:
Any comment?

shargon (dismissed):
It's good to me :)

authentication-scheme.mediawiki
```diff
@@ -0,0 +1,221 @@
...
+=====network=====
+
+The magic number of the network.
+Indicates which network this authentication will be applied to.
+
+=====nonce=====
+
+A 64-bit unsigned integer used to identify this authentication.
+
+The server should record the expiration time of the nonce and check it when the response is received.
```
djnicholson, line 64: This requires the authentication server to be stateful. That could be avoided by including a signature for the nonce and timestamp in the request (and requiring that the client provide the signature in the response).

erikzhang: If the server is stateless, how to identify the clients?

djnicholson: State will definitely be needed to provide any useful authorization (e.g. a table mapping from Neo address to roles and/or other application-specific identity data).  I'm just showing an opportunity to remove a database table and to avoid a database write on the initial authentication request (which could become a vector for unauthenticated denial-of-service).

erikzhang: If the server does not record the nonce, it will become reusable. We need a mechanism to ensure that when a client is successfully authenticated, the nonce is immediately invalidated.

djnicholson: It would be good to add mention here that the server is required to prevent nonce re-use:

> When the server receives the response, it should check whether the nonce has expired. If the nonce has expired or not exists, the authentication fails.

It's currently not clear that this is required and a server implementor might choose to delete a nonce after successful use rather than invalidate it.  I'm also curious as to why preventing nonce re-use is required (what security property breaks of the server doesn't do that?)

I'd still lean towards having the nonce invalidation happen post-authentication though.

erikzhang: For example, I log in to a website by scanning a QR code. If the nonce can be reused, the second person who scans the code will kick my login.

chenzhitong:
Rename `nonce` to `UUID`?

chenzhitong:
To prevent phishing sites from using the QR code of a particular website. We should add the website name to the Challenge payload. When you scan the code, you will be prompted with "You will be logged in to **** website". It's like having the website name in the SMS verification code.

chenzhitong:
Do we need to add an expiration timestamp to the Challenge payload.

authentication-scheme.mediawiki
```diff
@@ -0,0 +1,221 @@
...
+
+#The user clicks the login button with the NEO icon.
+
+#The QR code for login is displayed on the webpage. The challenge payload is encoded in the QR code. In this case, the <code>callback</code> must be included.
+
+#The user opens the mobile wallet and scans the QR code.
+
+#After the wallet recognizes the QR code, a pop-up window asks the user to confirm login.
+
+#If the user cancels the login, the authentication process ends. Otherwise, go to the next step.
```
chenzhitong, line 179: How the server gets the authentication refused message? Add `State` in Response data, include scaned, confirmed, refused, to optimize user experience?

authentication-scheme.mediawiki
djnicholson: It sounds like the nonce is being used as a session ID? If that is the case there is a vulnerability:  Anyone who has observed the QR code can later hijack the session (it wouldn't be safe to use this flow in a room with security cameras, for example).  

To mitigate this, the nonce should be swapped for a session ID and this swap should only be allowed to happen once (note that there is still a race condition where an attacker can steal the session, though--to prevent that you need some sort of communication channel from the phone back to the browser).

djnicholson:
> To prevent site A from using site B's QR code to trick you into logging in. We should add the website `name` to the `Challenge payload`. When you scan the code, you will be prompted with "You will be logged in to **** website". It's like having the website name in the SMS verification code.

The challenge payload will also need to be signed by the website (or the attacking site can just modify it)

erikzhang:
> Rename nonce to UUID?

It's not UUID. It is a one-time random number.

> Do we need to add an expiration timestamp to the Challenge payload.

I don't think it's necessary.

authentication-scheme.mediawiki
erikzhang: If the authentication is cancelled, the server won't receive the response payload.

authentication-scheme.mediawiki
erikzhang: The nonce is a one-time random number. There should be a nonce to sessionId mapping table on the server side.

erikzhang:
> To prevent site A from using site B's QR code to trick you into logging in.

In QR mode, there is a `callback` field. The client should show the domain of the callback url to the user. And the user can verify whether the website they are logging in is the domain name.

authentication-scheme.mediawiki
igormcoelho: > This requires the authentication server to be stateful. 

This is an interesting point @djnicholson, I need to study more on authentication. But perhaps, one way of not storing these nonces is to have some good pseudo-random algorithm, such as mersenne twister on server, starting with some secret seed, and then strongly hash mersenne twister outputs (via SHA-256, for example) to generate the nonce. The good thing is that, as long as some timestamp is involved, server can re-execute seeding process to find the nonce, instead of storing it. The bad thing is that, as soon as seed is leaked on server, nonces could be predicted.
Is it Neo servers or other servers that require being stateful?

Jim8y (dismissed):

Jim8y:
@shargon

vncoelho (commented):
@igormcoelho

shargon (dismissed):

nep-20.mediawiki
```diff
@@ -0,0 +1,247 @@
...
+  Title: Authentication Scheme
+  Author: Erik Zhang <erik@neo.org>
+  Type: Standard
+  Status: Accepted
+  Created: 2021-03-19
+</pre>
+
+==Abstract==
+
+This NEP describes a scheme for third-party applications to authenticate users based on NEO.
```
igormcoelho: "based on NEO" ... perhaps "NEO protocol" ? or based on "NEO token"? Is a token necessary for anything related to this proposal? It seems disconnected from any onchain mechanism, so I think it's also useful to explain that this is an entirely offchain authentication scheme.

nep-20.mediawiki
```diff
@@ -0,0 +1,247 @@
...
+The third-party applications mentioned here can be any type of application, including Dapps, websites, games, etc.
+
+==Motivation==
+
+Usually, an application saves user data on its own server. The authenticated user can modify the corresponding data through application-defined logic.
+Traditional Internet applications usually use an authentication scheme based on username and password, which is not suitable for blockchain.
+This authentication scheme allows third-party applications to authenticate users based on digital signatures.
+In this way, NEO users can log in to any third-party application that supports this scheme without registering, and allow applications to share user data on the blockchain.
```
igormcoelho: It is not clear in this proposal on how it allows users "to share user data on the blockchain". Maybe remove this part? Is it really an offchain approach, or somehow interacts onchain with a NEO capable blockchain?

nep-20.mediawiki
```diff
@@ -0,0 +1,247 @@
...
+==Abstract==
+
+This NEP describes a scheme for third-party applications to authenticate users based on NEO.
+The third-party applications mentioned here can be any type of application, including Dapps, websites, games, etc.
+
+==Motivation==
+
+Usually, an application saves user data on its own server. The authenticated user can modify the corresponding data through application-defined logic.
+Traditional Internet applications usually use an authentication scheme based on username and password, which is not suitable for blockchain.
+This authentication scheme allows third-party applications to authenticate users based on digital signatures.
```
igormcoelho: "digital signatures" , perhaps "digital signatures from NEO accounts"?

nep-20.mediawiki
```diff
@@ -0,0 +1,247 @@
...
+In this way, NEO users can log in to any third-party application that supports this scheme without registering, and allow applications to share user data on the blockchain.
+
+==Specification==
+
```
igormcoelho, line 23: At this point,  maybe before Specification, it could be useful to introduce the Document Structure.
Perhaps explain that "the Specification starts with a description of the proposed Payloads, namely the Challenge and the Response payloads. Then the document presents three use cases, or Scene Modes, namely QR code, Plugin and Connection modes."

nep-20.mediawiki
```diff
@@ -0,0 +1,247 @@
...
+=====allowed_algorithms=====
+
+An array containing the signature algorithms allowed by the server.
+
+The client should use one of the algorithms to sign the response data.
+If none of the algorithms are supported by the client, the authentication fails.
+
+=====domain=====
+
+The domain of the server. It is used to prevent phishing/MITM attacks.
```
igormcoelho, line 65: How the domain relates to the callback URL? Do they need to share the same prefix or not mandatory?

nep-20.mediawiki
```diff
@@ -0,0 +1,247 @@
...
+
+=====networks=====
+
+The magic number of the supported networks. Indicates which networks this authentication will be applied to.
```
igormcoelho, line 69: This part of "networks" is not completely clear, since it would probably mean NEO networks. But what kind of NEO networks are meant to be supported here? Does this include NeoX? Does it depend on P2P transmission/layer or the broad consensus algorithm to be part of NEO Ecosystem?
Perhaps the concept of NEO Networks should be clarified for the community, at least in this document.

nep-20.mediawiki
```diff
@@ -0,0 +1,247 @@
...
+
+=====nonce=====
+
+A 64-bit unsigned integer used to identify this authentication.
+
+The server should record the expiration time of the nonce and check it when the response is received.
+The expiration time is recommended to be 5 minutes.
```
igormcoelho, line 76: Should expiration time be informed to the user? Perhaps another optional field? Otherwise, how can the user know the estimated duration of the session? If informed, it could be safely calculated over base timestamp field.

nep-20.mediawiki
```diff
@@ -0,0 +1,247 @@
...
+It must correspond to the <code>pubkey</code>, otherwise the authentication fails.
+
+=====nonce=====
+
+A 64-bit unsigned integer used to identify this authentication.
+This value must be exactly the same as the <code>nonce</code> sent by the server.
+
+When the server receives the response, it should check whether the nonce has expired.
+If the nonce has expired or not exists, the authentication fails.
+
```
igormcoelho, line 135: Note that user must use HMAC to sign... otherwise server can easily break user signature, by forcing user to repeat it over different data. Another enhancing solution (if HMAC is forgotten!) is to the user add ANOTHER nonce field in Response.

nep-20.mediawiki
```diff
@@ -0,0 +1,247 @@
...
+=====timestamp=====
+
+The timestamp of the client.
+This timestamp is used by the server to determine the time of the signature to prevent replay attacks.
+
```
igormcoelho, line 140: This somehow mitigates the HMAC situation, but not entirely safe, unless wallet ensures it will never be repeated. I still prefer some explicit alternative User Nonce field to secure this response.

nep-20.mediawiki
```diff
@@ -0,0 +1,247 @@
...
+| The action sent by the server.
+|-
+| domain
+| var_str
+| variable
+| The domain of the server.
+|}
+
+===Scene modes===
+
```
igormcoelho, line 186: I think that "Scenes modes" could be explained in the introduction, as these are new words/vocabulary to be introduced by this NEP. What are Scenes? Perhaps "we define Scenes or Scene Modes the strategies employed by the expected use cases to ensure proper authentication using this NEP".

nep-20.mediawiki
```diff
@@ -0,0 +1,247 @@
...
+====Challenge payload====
+
+<pre>
+{
+  "action": "Authentication",
+  "grant_type": "Signature",
+  "allowed_algorithms": ["ECDSA-P256"],
+  "domain": "example.com",
+  "networks": [860833102],
```
igormcoelho, line 41: What kind of Network identifiers are allowed? From NEO networks? N2, N3, NeoX, N4?
Isn't it interesting to also name the kind of NEO protocol accepted here?

igormcoelho (changes requested):
The proposal is very interesting and well written. I think some points still deserve clarification, specially regarding the nature of the proposal, if it's entirely offchain or is expected to have some onchain mechanism (it looks to be offchain, until now). Also the nature of NEO networks to be considered in this proposal (N2, N3, NeoX, N4?). This affects address formats, but I couldn't see any limitation in allowing multiple kinds of NEO networks here, as an offchain proposal. Finally, I have some thoughts on Account Delegation when reading this proposal... it seemed risky to me to mix authentication concerns with perhaps an account with NEO funds and tokens inside. So, perhaps some Native Control that registers Delegated Accounts ONLY FOR "Authentication" roles (we don't have other roles at the moment, but we could have!). It would be safer in my opinion, so transmitting on payload not only the Main Account but also the Delegated Account, that responds as an alternative identity to the Main Account (without requiring Main Account to sign payloads and expose funds to risk). The downside of this is that it begins to look like some onchain approach, that requires more efforts on the server to check if Main Account is indeed connected to the Delegated Account. So, this is a strong downside for this approach, if intending to keep NEP as entirely offchain.

nep-20.mediawiki
roman-khimov: BTW, it should "Neo" here then, "NEO" is a token.

nep-20.mediawiki
```diff
@@ -0,0 +1,247 @@
...
+In the second step, the client sends the response payload to the server to complete the authentication.
+The payloads are transmitted in the form of JSON.
+
+====Challenge payload====
+
+<pre>
+{
+  "action": "Authentication",
+  "grant_type": "Signature",
+  "allowed_algorithms": ["ECDSA-P256"],
```
roman-khimov, line 39: The biggest problem to me here is that this is completely disconnected from real Neo accounts. Neo accounts are based on scripts, System.Crypto.CheckSig is just one of many possibilities, even though it's the most often used. This standard should work for any accounts. And it can be done, it just needs to push an address and then accept invocation/verification script pair instead of pubkey/signature.

nep-20.mediawiki
```diff
@@ -0,0 +1,247 @@
...
+  "domain": "example.com",
+  "networks": [860833102],
+  "nonce": "13458238842203010919",
+  "timestamp": 1616131368,
```
roman-khimov, line 43: Some ValidUntil is needed as well.

nep-20.mediawiki
```diff
@@ -0,0 +1,247 @@
...
+====Response payload====
+
+<pre>
+{
+  "algorithm": "ECDSA-P256",
+  "network": 860833102,
+  "pubkey": "0355912bc4e61c9715c5912397ea53a5ac6c103c4893fbd9c2a9f3be13b7a3e29d",
+  "address": "NfMFWYxaUUQy9SYo6AhRiGmRxfPxe9Edj7",
+  "nonce": "13458238842203010919",
+  "timestamp": 1616131369,
```
roman-khimov, line 101: Can be time-limited as well.

ajara87 (dismissed):

nep-20.mediawiki
igormcoelho: I fully agree, in terms of portability and indeed covering "multiple NEO-based platforms". But I also understand the benefit of being simple, easy to understand, and on practice, what 99,9999% people will use. And finally, the question of onchain x offchain standard... If some onchain-dependent opcode is to be allowed here, then the whole standard becomes onchain. It's not clear to me if it was on purpose offchain, but it seems to be offchain (don't know if this is good either).

nep-20.mediawiki
erikzhang: Fixed.

nep-20.mediawiki
erikzhang: Removed.

nep-20.mediawiki
erikzhang: Fixed.

nep-20.mediawiki
erikzhang: Added.

nep-20.mediawiki
erikzhang: Not mandatory

nep-20.mediawiki
erikzhang: The `network` field corresponds to the value defined in `ProtocolSettings`. There is no need to explicitly specify supported networks, as this parameter can be negotiated between the application and the wallet.

nep-20.mediawiki
erikzhang: The next field, `timestamp`, is used for this purpose.

nep-20.mediawiki
erikzhang: Does the `timestamp` help with this?

nep-20.mediawiki
erikzhang: The `network` field corresponds to the value defined in `ProtocolSettings`.

nep-20.mediawiki
erikzhang: I agree with this viewpoint. However, this is too difficult for developers of ordinary applications. I hope this standard can serve not only the Web3 industry but also traditional internet application scenarios. Therefore, we should minimize mechanisms that are difficult to implement when designing the standard. As @igormcoelho said, the `System.Crypto.CheckSig` address accounts for 99% of usage. In the future, we can further expand the `grant_type` field to implement other authentication methods.

nep-20.mediawiki
erikzhang: This is defined by the application within the scope of this specification.

nep-20.mediawiki
erikzhang: This is defined by the application within the scope of this specification.

nep-20.mediawiki
igormcoelho: Indeed, it can be used for this.
> The timestamp of the client. This timestamp is used by the server to determine the time of the signature to prevent replay attacks. 

However, the client must never repeat it. I don't know if client needs to do something special about this, or if it will naturally never repeat the timestamp... and I suppose it would require an explicit attack from server. But if it does, the only trusted protection I could imagine is having a fresh random nonce (not repeated from the server request), or simply using HMAC. That is why, the safest option is to use HMAC, or to have a second nonce.
There was a time on C# I saw some codes doing non-deterministic signatures... some time ago, so this could never happen... it should be deterministic in this case.

nep-20.mediawiki
igormcoelho: As long as timestamp indeed prevents replay attacks, it resolves the problem. Or just use HMAC.

nep-20.mediawiki
igormcoelho: Ok, this part is clear. Only doubts was about on what protocol the server was interacting with... I suppose the same server could interact with two different NEO-based chains, at this point.

nep-20.mediawiki
igormcoelho: So perhaps, grant type is Signature now, but Script Witness later. It makes sense.

nep-20.mediawiki
```diff
@@ -17,7 +17,6 @@ The third-party applications mentioned here can be any type of application, incl
 Usually, an application saves user data on its own server. The authenticated user can modify the corresponding data through application-defined logic.
 Traditional Internet applications usually use an authentication scheme based on username and password, which is not suitable for blockchain.
 This authentication scheme allows third-party applications to authenticate users based on digital signatures.
-In this way, NEO users can log in to any third-party application that supports this scheme without registering, and allow applications to share user data on the blockchain.
```
igormcoelho: Perhaps, you can keep the first part @erikzhang ...

> In this way, NEO users can log in to any third-party application that supports this scheme without registering.

This is fine.

nep-20.mediawiki
igormcoelho: I see. It's also unlikely to have different NEO-based protocols with same network fields.

nep-20.mediawiki
erikzhang: Signing requires the user's consent, which involves user interaction, so it's unlikely that two signatures can be completed at the same timestamp.

nep-20.mediawiki
erikzhang: Done.

nep-20.mediawiki
roman-khimov: I'd rather have proper account support immediately. While it can be added later, we've done that to NeoFS once (https://github.com/nspcc-dev/neofs-api/pull/324 + a lot of patches to SDK and node, originally NeoFS only supported single-key account, now it can handle anything) and that experience tells me it'd be much easier if it's designed properly from the beginning.

erikzhang: I do not intend to support addresses other than `System.Crypto.CheckSig` here. Doing so would mean that NEP-20-enabled applications would need to run a NEO node, or at least a trusted RPC server. This is currently unacceptable for traditional applications or gaming platforms. We need a fully off-chain authentication scheme.

nep-20.mediawiki
igormcoelho: That's true, it's highly unlikely.

nep-20.mediawiki
igormcoelho: Within the scope of off-chain authentication, indeed some opcodes should be banned. I still see use cases for Neo Scripts other than that, like Multisig, which is perhaps the Absolute Most Common within the 0,000001% which is not CheckSig :joy: At least that is off-chain.
Anything other than that, specially onchain authentication, at this point, exceed my capacity of imagination... it's theoretically useful, but I cannot see how. As long as there's a way to change that later, as Erik already pointed out, that's fine for me at this point.

PR #142, opened 2021-10-28
Add format standard for supportedstandards.

superboyiii:


erikzhang:
I think you should amend [NEP-15](https://github.com/neo-project/proposals/blob/master/nep-15.mediawiki).

PR #145, opened 2021-11-24
Dapi for N3

erikzhang:


lllwvlvwlll:
I'm a proponent of standardizing the protocol as long as it does not impact the transport layer of any solution.  Ultimately, most of this interfacing will be handled by sdks which abstract a lot of this so its primarily a standard for tooling developers.

I would also like to request that we generalize the name of this standard since `dApi` is already a software implementation which also includes a transport layer.

nep-20.mediawiki
```diff
@@ -0,0 +1,419 @@
...
+    type: "FAILED";
+    detail: InvocationResult;
+}
+</pre>
+
+===How to obtain an instance of IDapiProvider interface?===
+
+In order for dapps to have a unified method of obtaining <code>IDapiProvider</code> instances, all providers must trigger the <code>Neo.DapiProvider.ready</code> event of the <code>window</code> object when they are ready.
+
+The front end of the dapp can listen to the event and obtain the <code>IDapiProvider</code> instance from the <code>detail</code> property of the event.
```
devhawk, line 407: This is fine for web based dapp clients, but what about mobile or desktop clients?

nep-20.mediawiki
```diff
@@ -0,0 +1,419 @@
...
+
+<pre>
+window.addEventListener("Neo.DapiProvider.ready", e => {
+    var provider = e.detail.provider;
+});
+</pre>
+
+==Rationale==
+
+This protocol will allow dApp developers to create applications that interact with the NEO blockchain without having to be concerned about managing a full wallet within their application or the details related to handing transaction creation or broadcasting. This will also allow dApps to allow users to transact in a secure fashion that does not require sharing of their private key.
```
devhawk, line 417: Is this intended to be a protocol or an API? the specification above appears to be a Typescript API, but here it's described as a protocol (which would imply the possibility of multiple implementations in different programming languages)

nep-20.mediawiki
erikzhang: I think it is difficult for mobile clients and desktop clients to have a unified method to obtain `IDapiProvider` objects.

nep-20.mediawiki
erikzhang: Currently, the front end of most dapps is Javascript running in browsers.

devhawk: > Currently, the front end of most dapps is Javascript running in browsers.

This doesn't answer my "is this intended to be a protocol or api" question

nep-20.mediawiki
devhawk: It doesn't have to be a unified method, but I'd expect there to be *some* method for mobile and desktop apps to access a DAPI provider

nep-20.mediawiki
erikzhang: The specification describes the interfaces, so it is a protocol. Different wallet providers can have different implementations, but the interaction protocol with dapps is the same.

nep-20.mediawiki
erikzhang: Do you have any ideas?

melanke:
Hello, I just have a few minor considerations regarding this specification:

I've tested Dapi with NeoLine in the past and also by checking it's documentation, I understood that the Chrome Extension doesn't need a proper "connection"
process, the dapp receives the wallet information automatically.

1) This is very different from WalletConnect 2.0: the dapp will need to show a code for the user to input on the wallet, this can be done via QRCode or Deeplinking, so it would be interesting to have such event to allow the dapp to handle such UI interaction.

2) On this specification proposal there is an `authenticate` method. I am not very sure if it was designed with the connection purpose, to continue the discussion I will assume it was. So, here is my first question: What is the purpose of the `AuthenticationChallengePayload.nonce` property?

3) The `Network` enum doesn't allow tests on a private network. And `AuthenticationChallengePayload.network` only accepts a single network, which is fine on most of the cases, but we tought that it would be interesting for the dapp to accept connecting with multiple networks, so the user on it's wallet can control if it will be used for tests, private network or production. This would be great for development purposes, it would make dapp development way easier, testing on the private network before sending to testnet or mainnet.

4) Again, on the connection process, WalletConnect expects the Dapp's meta-data. This is used to show the dapp's name, icon and other information on the wallet. The metadata has this format:
```
export type DappMetadata = {
    name: string
    description: string
    url: string
    icons: string[] // honestly, we could allow a single icon, so it don't need to be an array
}
```
I believe it could be an interesting addition.

5) Both **NeonWallet's WalletConnect** integration and **NeoLine's Dapi** integration have the following methods that I consider important and could be added to the specification:
- [invokeMultiple](https://neoline.io/dapi/N3.html#invokeMultiple). On NW we are accepting an optional parameter on each invocation called `abortOnFail`: it adds `0x39` on the script, so the whole transaction will be aborted if this invocation returns false, which might be very important for transactional operations.
- [signMessage](https://neoline.io/dapi/N3.html#signMessage)

6) The following methods are less important because they could be called directly from the dapp, without the need of the wallet integration, but they are pretty handy, specially to avoid a Neo javascript SDK dependency on the dapp. On NW we are calling those methods without the need for the user authorization, since they are pretty secure, so the dapp and the wallet has a smooth interaction:
- [verifyMessage](https://neoline.io/dapi/N3.html#verifyMessage)
- [invokeRead](https://neoline.io/dapi/N3.html#invokeRead)
- [invokeReadMultiple](https://neoline.io/dapi/N3.html#invokeReadMulti)

7) Being able to easily convert an Address or ScriptHash to hash160 would also be pretty handy. Both NeoLine and NeonWallet handle this differently but we could agree on a single method.

nep-20.mediawiki
melanke: WalletConnect 2.0 handles this using a "connection code": It could open the wallet via Deeplinking or show the code as an QRCode image, and then, on the Wallet the user can confirm the connection. So, for this to work, we would need a new event, so the dapp can handle this UI interaction to show the code.

erikzhang:
> On this specification proposal there is an authenticate method.

Many websites require users to log in to provide services. The `authenticate` method provides an identity authentication scheme for users to log in to websites using blockchain addresses.

> verifyMessage

I think it should be provided in SDK.

> invokeRead

It has been renamed to `call`.

> Being able to easily convert an Address or ScriptHash to hash160 would also be pretty handy.

I think it should be provided in SDK.

devhawk:
What is the relationship of this to WalletConnect? Are they complementary or competitive?

erikzhang:
> What is the relationship of this to WalletConnect? Are they complementary or competitive?

Dapi is a protocol for communication between plug-in wallet and dapp via javascript. And WalletConnect allows communication between the two through a centralized server to relay messages.

nep-20.mediawiki
```diff
@@ -0,0 +1,533 @@
...
+
+// Represents the N3 network.
+// MAINNET: 860833102
+// TESTNET: 877933390
+export type Network = number;
+
+// Represents the name of the event in IDapiProvider.
+export type EventName =
+    "accountschanged" | // Triggered when the accounts in the wallet change.
+    "networkchanged";   // Triggered when the user switches the current network.
```
melanke, line 72: Could we have a new event for `connectionProposal` that could be emitted to deliver a connection code? The dapp could show it as QRCode to stablish a connection. This is very important for mobile Wallets.

nep-20.mediawiki
```diff
@@ -0,0 +1,533 @@
...
+        signatures: Record<ECPoint, Base64Encoded>;
+    }>;
+    network: Network;
+}
+
+export type AuthenticationChallengePayload = {
+    action: "Authentication";
+    grant_type: "Signature";
+    allowed_algorithms: ["ECDSA-P256"];
+    network: Network;
```
melanke, line 348: I think `network` could be an array, so the dapp could specify that it allows connecting with many networks and is up to the wallet to decide which of those it will connect. This gives more flexibility for dapp testing and if the dapp wants to avoid this complexity it could describe a single value.

erikzhang: I'm not sure if this is possible. Because the address format of different networks may be different.

nep-20.mediawiki
erikzhang: I don't quite understand the role of `connectionProposal`. Can you draw a flowchart to explain it?

melanke: - Hey, I just made this, I hope it could be helpful:

wallet-connect-sequence-diagram

- Also, I wrote about it on [this Readme](https://github.com/CityOfZion/wallet-connect-sdk)

---

- And also, I wrote this while I was investingating it:

# Wallet Connect implementation
Wallet Connect is a protocol to allow a safer communication between dApp and Wallet. It works with any blockchain because it simply creates a WebSocket server that handles the communication between dApp and Wallet in a smart way, handling format validation, problems with internet connection, cache, log, etc. It's client-side libraries helps to integrate even further, allowing the user to scan a QR Code to connect the two ends.

## The parts


### Server
A simple middleman that handles the communication between the other parts. At first, we believe we would need to change Wallet Connect relay server so it could work with Neo Blockchain and then send a pull request. But after exploring its code, it looks like we can simply use Wallet Connect server as it is right now.

### dApp
A dApp is an application that uses smart contracts and other technologies as server-side. To securely reach the Smart Contracts, the user must use its own Wallet to sign the operations, and the responsible to delegate such interaction is the dApp's client-side. Firstly, the dApp establishes a connection with Wallet Connect which asks the user to scan the QR Code, by doing this, Wallet Connect knows how to make an end-to-end encryption and transfer information between the dApp and the Wallet. Now, the dApp can request to do an operation on the blockchain using the user's Wallet. By using JsonRPC format, it says which method and arguments it wants to call.

### Wallet
The Wallet is a client-side application that allows the user to handle its assets, sign documents and interact with Smart Contracts. If integrated with Wallet Connect, it can receive messages from dApps that requests the user to procced with a blockchain operation. When receiving such methods, the Wallet should ask for user authorization and then call the blockchain method by itself. A good UX pattern is to show a mobile notification so the user can allow it without having to open the app by himself.

nep-20.mediawiki
melanke: Sure, but the account will be different for each connection. It's not supposed to let the wallet change the network on the fly, only to choose which network it will use on the entire connection, from beginning to end. I've already implemented such interaction between dapp and wallet on my Wallet Connect experiments and it makes sense.

erikzhang: Done.

nep-20.mediawiki
erikzhang: As I said:

> Dapi is a protocol for communication between plug-in wallet and dapp via javascript. And WalletConnect allows communication between the two through a centralized server to relay messages.

They are different.

melanke: Sure, but they focus solving the same problem. Are we going to exclude the possibility for WalletConnect integration on this specification? I mean, if the dapp wants to have multiple dapp-wallet integration providers, WalletConnect couldn't be one of them? Meaning that there will be absolutely no mobile wallet providers.

EdgeDLT:
I think it should be possible for a dapp to provide a signature to a transaction created using the Dapi. From what I understand it should not increase risk for users, only for dApps that need to be careful they don't provide a signature to a malicious transaction.

My imagined use case is for subsidizing transactions as the first signer, but I imagine there are others situations where apps might want to join in a multisig or approve certain contract calls.

erikzhang:
> I think it should be possible for a dapp to provide a signature to a transaction created using the Dapi.

`maketransaction` and `sign` are designed for this situation.

nep-21.mediawiki
```diff
@@ -0,0 +1,535 @@
...
+    sign(context: ContractParametersContext): Promise<ContractParametersContext>;
+    
+    // Signs the message with the specified account.
+    // The algorithm used is ECDsa with SHA256.
+    // If `account` is omitted, the wallet should automatically select an account or prompt the user to select one.
+    // Possible errors: INVALID, NOTFOUND, TIMEOUT, CANCELED.
+    signMessage(message: Base64Encoded, account?: UInt160): Promise<{ signature: Base64Encoded; account: UInt160; pubkey: ECPoint }>;
+
+}
+</pre>
```
vncoelho, line 481: How about ask wallet to add Network and Switch to another Network?

roman-khimov (commented):
I've never used it, so just some nitpicking. Likely we need more input from  providers/consumers of this API.

vncoelho (dismissed):
From my last review I believe that this https://github.com/neo-project/proposals/pull/145#discussion_r1009320733 was pending

nep-21.mediawiki
vncoelho: I would like that the authors of the proposals checked that and solved the issues pending. Maybe it is needed or not.

vncoelho:
> From my last review I believe that this https://github.com/neo-project/proposals/pull/145#discussion_r1009320733 was pending

Ping @erikzhang 

Methods for addnewchain and switchtonetwork,something like this.

nep-20.mediawiki
erikzhang: Yes. WalletConnect 2.0 is indeed well-suited for handling such scenarios.

nep-20.mediawiki
erikzhang: As I understand it, WalletConnect 2.0 has its own protocol, which does not need to be written into this specification.

nep-21.mediawiki
erikzhang: If a dapp requests to switch to a different network, it should call the `authenticate` method again, passing in the network ID to which it wants to switch.

nep-21.mediawiki
```diff
@@ -0,0 +1,535 @@
...
+<pre>
+// Represents a piece of data encoded as a base64 string.
+export type Base64Encoded = string;
+
+// Represents a N3 address.
+// Example: "NSiVJYZej4XsxG5CUpdwn7VRQk8iiiDMPM"
+export type Address = string;
+
+// A 160-bit hash represented by a hexadecimal string.
+// Example: "0x682cca3ebdc66210e5847d7f8115846586079d4a"
```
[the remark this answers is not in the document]
erikzhang: No.

nep-21.mediawiki
```diff
@@ -0,0 +1,535 @@
...
+export type UInt160 = string;
+
+// A 256-bit hash represented by a hexadecimal string.
+// Example: "0x1f4d1defa46faa5e7b9b8d3f79a06bec777d7c26c4aa5f6f5899a291daa87c15"
```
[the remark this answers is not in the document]
erikzhang: No.

nep-21.mediawiki
```diff
@@ -0,0 +1,535 @@
...
+export type UInt256 = string;
+
+// Represents an ECC public key.
+// Example: "03b209fd4f53a7170ea4444e0cb0a6bb6a53c2bd016926989cf85f9b0fba17a70c"
+export type ECPoint = string;
+
+// A large integer of any length expressed as a string.
```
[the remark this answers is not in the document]
erikzhang: No.

nep-21.mediawiki
```diff
@@ -0,0 +1,535 @@
...
+export type Integer = string;
+
+// Represents a piece of data encoded as a hexadecimal string.
+export type HexString = string;
+
+// Represents a version.
+// Example: "1.0.0"
```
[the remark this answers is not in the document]
erikzhang: This can be arbitrary.

nep-21.mediawiki
```diff
@@ -0,0 +1,535 @@
...
+// MAINNET: 860833102
+// TESTNET: 877933390
+export type Network = number;
+
+// Represents the name of the event in IDapiProvider.
+export type EventName =
+    "accountschanged" | // Triggered when the accounts in the wallet change.
+    "networkchanged";   // Triggered when the user switches the current network.
+
+// Represents the type of ContractParameter.
```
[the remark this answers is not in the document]
erikzhang: Done.

nep-21.mediawiki
```diff
@@ -0,0 +1,535 @@
...
+    "InteropInterface" | // Indicates that the parameter is an interoperable interface.
+    "Void";              // It can be only used as the return type of a method, meaning that the method has no return value.
+
+// Represents a parameter of a contract method.
+export type ContractParameter = {
+    name?: string;               // The name of the parameter.
+    type: ContractParameterType; // The type of the parameter.
+}
+
+// Represents an account in a wallet.
```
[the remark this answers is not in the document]
erikzhang: It's similar to the NEP-6 account, but it doesn't have a `key` field and adds a `hash` field. So the two are not the same.

nep-21.mediawiki
```diff
@@ -0,0 +1,535 @@
...
+
+// Provides the necessary arguments for a contract call.
+export type InvocationArguments = {
+    hash: UInt160;         // The hash of the contract to be called.
+    operation: string;     // The operation of the contract to be called.
+    args?: Argument[];     // The arguments for the call.
+    abortOnFail?: boolean; // Indicates whether the entire transaction should fail when the contract returns `false`.
+}
+
+export type InvocationResult = {
```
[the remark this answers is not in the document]
erikzhang: Added.

nep-21.mediawiki
```diff
@@ -0,0 +1,535 @@
...
+export type Argument = {
+    type: StackItemType;
+    value?: any;
+}
+
+export type ApplicationLog = {
+    txid: UInt256;
+    executions: {
+        trigger: TriggerType;
+        vmstate: VMState;
```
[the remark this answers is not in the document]
erikzhang: `getapplicationlog` did not modify this field, so I will keep its current name.

nep-21.mediawiki
```diff
@@ -0,0 +1,535 @@
...
+    // Sends assets to an account and returns the hash of the transaction.
+    // If `from` is omitted, the wallet should automatically select an account or prompt the user to select one.
+    // Possible errors: INVALID, NOTFOUND, FAILED, TIMEOUT, CANCELED, INSUFFICIENT_FUNDS, RPC_ERROR.
+    send(asset: UInt160, amount: Integer, to: UInt160, from?: UInt160): Promise<UInt256>;
+
+    // Signs the transaction with the current wallet. Usually used for multi-signature transactions.
+    // Possible errors: UNSUPPORTED, INVALID, NOTFOUND, TIMEOUT, CANCELED.
+    sign(context: ContractParametersContext): Promise<ContractParametersContext>;
+    
+    // Signs the message with the specified account.
```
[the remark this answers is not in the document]
erikzhang: Currently, there is no standard for how to sign messages. I plan to keep it as a plain message signature for now, and any prefixes or salts can be added to the beginning of the message by the dapp.

nep-21.mediawiki
```diff
@@ -0,0 +1,535 @@
...
+window.addEventListener("Neo.DapiProvider.ready", e => {
+    var provider = e.detail.provider;
+});
+</pre>
+
+==Rationale==
+
+This protocol will allow dApp developers to create applications that interact with the NEO blockchain without having to be concerned about managing a full wallet within their application or the details related to handing transaction creation or broadcasting. This will also allow dApps to allow users to transact in a secure fashion that does not require sharing of their private key.
+
+==Implementation==
```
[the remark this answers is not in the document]
erikzhang: This repository needs to be updated according to this standard.

nep-21.mediawiki
vncoelho: This is not the standard of other chains, as usually we can add a new network as well.

vncoelho:
It has been answered but not resolved or solved

vncoelho:
https://github.com/ethereum/EIPs/issues/3086

https://github.com/MetaMask/metamask-extension/issues/1820

https://ethereum-magicians.org/t/eip-3326-wallet-switchethereumchain/5471

Perhaps future NEPs can also specify that

erikzhang:
> Perhaps future NEPs can also specify that

As I have said, you can use `IDapiProvider.supportedNetworks` to determine which networks the current provider supports. Then use `IDapiProvider.authenticate()` to switch the network.

vncoelho:
> > Perhaps future NEPs can also specify that
> 
> As I have said, you can use `IDapiProvider.supportedNetworks` to determine which networks the current provider supports. Then use `IDapiProvider.authenticate()` to switch the network.

I see it can determine and switch, but how to ask to add a new one?
The switch can pass such parameters?

erikzhang:
> I see it can determine and switch, but how to ask to add a new one?

I think it's hard to add new network. The provider needs to sync blocks.

vncoelho:
> > I see it can determine and switch, but how to ask to add a new one?
> 
> I think it's hard to add new network. The provider needs to sync blocks.

I think I am not in the same page.

This is like MetaMask changing from ETH RPCs to BNB, Polygon, Arbitrium,etc...
It is just a matter of the RPC parameters and network ID.

Like from NEO testnet to NEO mainnet or a privatenet for testing.

erikzhang:
> This is like MetaMask changing from ETH RPCs to BNB, Polygon, Arbitrium,etc... It is just a matter of the RPC parameters and network ID.
> 
> Like from NEO testnet to NEO mainnet or a privatenet for testing.

The provider needs to sync the blocks. Otherwise, how would you respond to requests such as `relay` or `getBlock`?

vncoelho:
MetaMask (the provider), for example, adds network based on the RPC URL, the endpoint (node) is the one which is synced.
So, basically we can add any NEO privatenet and just point to the RPC of that seed node.

I am starting to believe that maybe we are talking about different things.

erikzhang:
> MetaMask (the provider), for example, adds network based on the RPC URL, the endpoint (node) is the one which is synced.
> So, basically we can add any NEO privatenet and just point to the RPC of that seed node.

I got it. To add a new network, we need to provide the `ProtocolSettings` used by that network and an RpcServer endpoint. I'll consider what else we need.

vncoelho:
WhatsApp Image 2026-03-19 at 09 25 06

WhatsApp Image 2026-03-19 at 09 25 05

Something like this for `wallet_switchEthereumChain` :
```json
{
  "id": 1,
  "jsonrpc": "2.0",
  "method": "wallet_switchEthereumChain",
  "params": [
    {
      "chainId": "0x5",
      "fallbackRpc": {
        "chainId": "0x5",
        "chainName": "Goerli",
        "rpcUrl": "https://goerli.infura.io/v3/INSERT_API_KEY_HERE",
        "nativeCurrency": {
          "name": "Goerli ETH",
          "symbol": "gorETH",
          "decimals": 18
        },
        "blockExplorerUrl": "https://goerli.etherscan.io"
      }
    }
  ]

}
```

nep-21.mediawiki
```diff
@@ -0,0 +1,626 @@
...
+export type Signer = {
+    account: UInt160;
+    scopes: WitnessScope;
+    allowedContracts?: UInt160[];
+    allowedGroups?: ECPoint[];
+    rules?: WitnessRule[];
+}
+
+export type TransactionAttributeType =
+    "HighPriority" |
```
Wi1l-B0t, line 226: more TransactionAttributeType now.

erikzhang: I'm not sure exactly how to describe this here, but the number of `TransactionAttributeType` is likely to continue increasing in the future.

erikzhang:
> Something like this for wallet_switchEthereumChain :

To achieve this, the core NEO library needs the capability to read the names of NEO and GAS from `ProtocolSettings` in order to support private chains. Implementing this change is non-trivial, and I am keen to see the dAPI standard finalized as soon as possible. Perhaps we could defer the add-network to version 2.0.

shargon (dismissed):

shargon (approved):

hacfox:
# Clarify active account semantics in `getAccounts()` and `accountchanged` event

## Description

I found a potential ambiguity in the current NEP-21 dAPI specification around `getAccounts()` and the `accountchanged` event.

In the document, `getAccounts()` is defined as:

```ts
// Gets the current connected account in the wallet.
getAccounts(): Promise<Account[]>;
```

The comment says "the current connected account" in singular form, but the return type is `Account[]`.

Also, the `accountchanged` event is described as being triggered when the user changes the connected account in the wallet, and the event payload contains:

```ts
accounts: Account[]; // The new connected accounts in the wallet.
```

This raises a question for dApps:

When multiple accounts have been authorized or connected, how should the dApp determine which account is the currently active or selected one?

## Example

Assume the following case:

- Account A, B, and C have all been authorized by the dApp.
- The user switches the currently selected wallet account to C.
- The wallet emits an `accountchanged` event.
- The event payload returns an account array.

For example:

```ts
accounts: [C, A, B]
```

or:

```ts
accounts: [A, B, C]
```

If the specification does not define the ordering semantics of the `accounts` array, the dApp cannot reliably know which account is the active one.

## Questions

I would like to clarify the expected behavior:

1. Should `accounts[0]` always represent the currently selected / active account?
2. Is the `accounts` array intended to represent all connected accounts, or only the currently active account?
3. Should the `accountchanged` event be triggered when the active account changes, when the connected account list changes, or both?
4. Should the comment for `getAccounts()` be updated to avoid the singular/plural mismatch?

## Suggested clarification

One possible clarification could be:

```ts
// Gets the connected accounts in the wallet.
// The first account in the array MUST be the currently selected / active account.
getAccounts(): Promise<Account[]>;
```

And for `AccountChangedEvent`:

```ts
accounts: Account[]; // Connected accounts. The first account MUST be the currently selected / active account.
```

The `accountchanged` event behavior could also be clarified as:

```ts
// Triggered when the currently selected account changes
// or when the connected account list changes.
```

## Concern

The main concern is that without a clearly defined rule, different wallet implementations may return the same connected accounts in different orders, causing dApps to interpret the active account inconsistently.

Defining `accounts[0]` as the active account would make the behavior deterministic and avoid ambiguity for dApps.

roman-khimov:
There are wallets already using it, it's kinda accepted, but still not merged, why? Do we except any changes? Do we have any feedback from wallets/applications?

nep-21.mediawiki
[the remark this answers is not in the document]
ixje: @erikzhang who is supposed to do that (so I can reach out)? The repo is still archived. COZ has a pending PR for adding dapi support to `neon-js` but not a single dapp listed on ndap.org implements it according to the spec. To quote my colleague

> They all have dedicated "Connect with NeoLine" or "Connect with O3" buttons instead of detecting which dAPI-compatible wallet extensions are installed

We're in need of a reference implementation to validate correctness.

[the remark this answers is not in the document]
erikzhang: https://onegate.gitbook.io/onegate/nep-21-dapi-for-n3/summary

Here is a documentation.

PR #148, opened 2022-04-04
Add source to NEF

shargon:


lllwvlvwlll:
@shargon do you have some examples of how this can be used?  Is there a standard format (like absolute path to the entrypoint) that this is being implemented alongside?

devhawk (changes requested):
This information is already in the debug info. https://github.com/devhawk/proposals/blob/devhawk/debug-info/nep-19.md#documents

I don't see the value of putting it in the NEF file

shargon:
> I don't see the value of putting it in the NEF file

It's already in the nef file

shargon:
@lllwvlvwlll it was introduced in https://github.com/neo-project/neo/pull/2605 , I think that it must be an absolute url.

devhawk:
> It's already in the nef file https://github.com/neo-project/neo/blob/master/src/neo/SmartContract/NefFile.cs#L57

Then I agree it should be in the NEP

devhawk (approved):

nep-16.mediawiki
```diff
@@ -30,6 +30,7 @@ nef_file
 {
   4-byte  magic;
   64-byte compiler;
+  var-byte source;
   2-byte reserve;
```
erikzhang, line 34: Then it becomes 1 byte.

shargon: At least one, but it can be more

erikzhang (approved):

PR #149, opened 2022-06-08
Don't check the caller in NEP17

shargon:
This is not needed, and redundat because it's done during the syscall https://github.com/neo-project/neo/blob/736c346b9d8b1404f10023eeecb3a8e92ae0c542/src/neo/SmartContract/ApplicationEngine.Runtime.cs#L216

### DIFF
--- a/nep-17.mediawiki
+++ b/nep-17.mediawiki
@@ -125,7 +125,7 @@ The function MUST return <code>false</code> if the <code>from</code> account bal
 
 If the method succeeds, it MUST fire the <code>Transfer</code> event, and MUST return <code>true</code>, even if the <code>amount</code> is <code>0</code>, or <code>from</code> and <code>to</code> are the same address.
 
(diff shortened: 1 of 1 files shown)


roman-khimov:
Yep, https://github.com/neo-project/proposals/pull/126#discussion_r522309255

shargon:
> Yep, https://github.com/neo-project/proposals/pull/126#discussion_r522309255

Maybe we can add just a note, *this syscall will compare against the callingScriptHash first..*?

erikzhang:
We need to explain what the contract is supposed to do and give the best practices for doing it in one step using `CheckWitness`.

shargon:
> We need to explain what the contract is supposed to do and give the best practices for doing it in one step using CheckWitness.

We can add a note like this `Take into account that <code>Neo.Runtime.CheckWitness</code> already will check whether the <code>from</code> address equals the caller contract hash.`

Issue #135, opened 2021-04-19
Need a proposal for upgrade method in contract to unify the style


satoshichou:
Currently, Neo don't have a unified style for `upgrade` method writing in SC so that neo-cli or neo-gui can't implement a command or a button for `upgrade` intergration. However, `upgrade` is very important for Dapp devlopment.
@erikzhang

PR #154, opened 2022-08-31
NEP 22: Contract Update Standard

superboyiii:
Close: https://github.com/neo-project/proposals/issues/135

### DIFF
--- a/nep-22.mediawiki
+++ b/nep-22.mediawiki
@@ -1,23 +1,24 @@
 <pre>
   NEP: 22
   Title: Contract Update Function
-  Author: Owen Zhang <zhangzhihong@ngd.neo.org>, Fernando Díaz Toledano <shargon@gmail.com>, Erik Zhang <erik@neo.org>
+  Author: Owen Zhang <zhangzhihong@ngd.neo.org>, Fernando Díaz Toledano <shargon@gmail.com>, Erik Zhang <erik@neo.org>, Roman Khimov <roman@nspcc.ru>
   Type: Standard
   Status: Accepted
   Created: 2021-04-19
 </pre>
 
+
 ==Abstract==
 
-This proposal outlines a basic method guideline for the NEO blockchain that will provide systems with a generalized interaction mechanism for smart contract update and destroy.
+This proposal outlines a standard for the NEO blockchain that will provide systems with a generalized interaction mechanism for smart contract update.
 
 ==Motivation==
 
-As the NEO blockchain scales, Smart Contract Update will become increasingly important.  Without guideline for these methods, systems will be required to maintain different API for different contracts, regardless of their similarity to other contracts which makes application development very inconvenient. This guideline can lead most contracts to implement the same basic methods with the same types of parameters but still accept diversity.
+As the NEO blockchain scales, Smart Contract Update will become increasingly important. Update mechanism is already there from initial N3 release. But we haven't summarized that standard. That's why this standard is necessary.
 
 ==Specification==
 
-In the method definitions below, we provide both the definitions of the functions as they are defined in the contract as well as the invoke parameters.
+Neo N3 has a native <code>ContractManagement</code> contract that is used for contract deployment and update via <code>deploy</code> and <code>update</code> methods of it. After settling with the new NEF and manifest both of them will invoke a special <code>_deploy</code> method if it's implemented by the contract.
 
 ===Methods===
 
@@ -45,7 +46,7 @@ In the method definitions below, we provide both the definitions of the function
 }
 </pre>
 
-This method will be automatically executed by ContractManagement contract when a contract is updated. When <code>update</code> in <code>_deploy</code> is true, ContractManagement contract will update new contract state by this mehod.
+This method will be automatically executed by ContractManagement contract when a contract is updated. When <code>update</code> in <code>_deploy</code> is true, ContractManagement contract will update new contract state by this method.
 
 Updating a smart contract MUST have <code>nefFile</code> and<code>manifest.json</code>. It will be passed to <code>_deploy</code>.
 
--- a/nep-22.mediawiki
+++ b/nep-22.mediawiki
@@ -9,7 +9,7 @@
 
 ==Abstract==
 
-This proposal outlines a basic method guideline for the NEO blockchain that will provide systems with a generalized interaction mechanism for smart contract initial deploy, update and destroy.
+This proposal outlines a basic method guideline for the NEO blockchain that will provide systems with a generalized interaction mechanism for smart contract initial deploy, update, verify and destroy.
 
 ==Motivation==
 
@@ -69,11 +69,13 @@ The parameter <code>data</code> can be any type of supported parameters for cont
 }
 </pre>
 
-If this contract address is included in the transaction signature, it MUST apply this. This method will be triggered as a VerificationTrigger to verify that the signature is correct.. Here's an example:
+If this contract has scenario that its address will be included in the transaction signature, it MUST apply this. A typical scenario is withdraw tokens from a smart contract. The first Verify method in that contract's manifest abi will be triggered as a VerificationTrigger to verify if the signature is correct. Here's an example:
 
 <code>public static bool Verify() => Runtime.CheckWitness(Owner);</code>
 
-This method needs to be called when withdrawing token from the contract.
+This method needs to be called when withdrawing token from the contract. 
+
+Verify method can have multi params as well.
 
 ====update====
 
--- a/nep-22.mediawiki
+++ b/nep-22.mediawiki
@@ -14,11 +14,11 @@ This proposal outlines a standard for the NEO blockchain that will provide syste
 
 ==Motivation==
 
-As the NEO blockchain scales, Smart Contract Update will become increasingly important. Update mechanism is already there from initial N3 release. But we haven't summarized that standard. That's why this standard is necessary.
+As the NEO blockchain scales, Smart Contract Update will become increasingly important. We need standize <code>update</code> method in contract to callback native <code>ContractManagement</code> correctly.
 
 ==Specification==
 
-Neo N3 has a native <code>ContractManagement</code> contract that is used for contract deployment and update via <code>deploy</code> and <code>update</code> methods of it. After settling with the new NEF and manifest both of them will invoke a special <code>_deploy</code> method if it's implemented by the contract.
+Neo N3 has a native <code>ContractManagement</code> contract that is used for contract deployment and update via <code>deploy</code> and <code>update</code> methods of it. After settling with the new NEF and manifest both of them will invoke a special <code>_deploy</code> method if it's implemented by the contract. More details in [NEP-29](https://github.com/neo-project/proposals/blob/master/nep-29.mediawiki)
 
 ===Methods===
 
@@ -46,13 +46,13 @@ Neo N3 has a native <code>ContractManagement</code> contract that is used for co
 }
 </pre>
 
-This method will be automatically executed by <code>ContractManagement</code> contract when a contract is updated. When <code>update</code> in <code>_deploy</code> is true, <code>ContractManagement</code> contract will update new contract state by this method.
+This method will will callback <code>Update</code> method in native <code>ContractManagement</code> contract. It will update contract state if it's exectued successfully.
 
-Updating a smart contract MUST have <code>nefFile</code> and<code>manifest.json</code>. It will be passed to <code>_deploy</code>.
+Updating a smart contract MUST have <code>nefFile</code> or <code>manifest.json</code>. If just update one of them at a time is OK, another SHOULD be set as null. It will be passed to <code>_deploy</code>.
 
 The parameter <code>data</code> can be any type of supported parameters for contract-specific purpose. For example some kinds of json parameters can be put into data for specific check.
 
-This function will check whether the <code>signer</code> address equals the <code>owner</code> hash of contract. This function SHOULD use the SYSCALL <code>Neo.Runtime.CheckWitness</code> to verify the <code>signer</code>. Details has been explained in [NEP-30](https://github.com/superboyiii/proposals/blob/upgrade-standard/nep-30.mediawiki).
+This function SHOULD check whether the <code>signer</code> address equals the <code>owner</code> hash of contract. This function SHOULD use the SYSCALL <code>Neo.Runtime.CheckWitness</code> to verify the <code>signer</code>. Details has been explained in [NEP-30](https://github.com/superboyiii/proposals/blob/upgrade-standard/nep-30.mediawiki).
 
 ===References===
 
--- a/nep-22.mediawiki
+++ b/nep-22.mediawiki
@@ -48,13 +48,13 @@ All contracts SHOULD implement this NEP to be able to fix bugs or otherwise impr
 
 This method MUST call <code>update</code> method in native <code>ContractManagement</code> contract. It will update contract state if it's executed successfully.
 
-Updating a smart contract MUST have <code>nefFile</code> or <code>manifest.json</code>. How do they work is in reference of [https://github.com/neo-project/proposals/blob/master/nep-15.mediawiki NEP-15] and [https://github.com/neo-project/proposals/blob/master/nep-16.mediawiki NEP-16]. Just updating one of them at a time is OK. This means no change on another, so another SHOULD be set as null. It MUST be passed to ContractManagement.Update() in this method.
+Calls to this method MUST have <code>nefFile</code> or <code>manifest</code>. These are serialized representations of NEF ([https://github.com/neo-project/proposals/blob/master/nep-16.mediawiki NEP-16] and NeoContract manifest ([https://github.com/neo-project/proposals/blob/master/nep-15.mediawiki NEP-15]). Omitting one of them (setting to null) is a valid call, this means no changes to this component (old version is kept). These two parameters MUST be passed into <code>ContractManagement</code>'s <code>update</code> method.
 
-The parameter <code>data</code> can be any type of supported parameters for contract-specific purpose. For [https://github.com/nspcc-dev/neofs-contract/blob/99fb86c35a48ed12017423aa4fee13f7d07fa4c0/contracts/proxy/contract.go#L33 example] some kinds of parameters can be put into data for specific check. Also some reference in [https://github.com/neo-project/proposals/blob/master/nep-29.mediawiki NEP-29].
+The parameter <code>data</code> can be any type of supported parameters for contract-specific purpose. For [https://github.com/nspcc-dev/neofs-contract/blob/99fb86c35a48ed12017423aa4fee13f7d07fa4c0/contracts/proxy/contract.go#L33 example] some kinds of parameters can be put into data for specific check. See [https://github.com/neo-project/proposals/blob/master/nep-29.mediawiki NEP-29] for additional reference on contract update callback mechanism that works with this parameter.
 
 This method SHOULD do <code>Checkwitness</code> for contract <code>owner</code> to avoid security risks. This function SHOULD use the SYSCALL <code>Neo.Runtime.CheckWitness</code> to verify the <code>signer</code>. 
 
-It's strongly recommended to execute all codes before calling ContractManagement.Update() to ensure they can be executed correctly.
+It's strongly recommended to execute all codes before calling <code>ContractManagement</code>'s <code>update</code> method to ensure they can be executed correctly.
 
 ===References===
 
--- a/nep-22.mediawiki
+++ b/nep-22.mediawiki
@@ -53,6 +53,12 @@ The parameter <code>data</code> can be any type of supported parameters for cont
 
 This function will check whether the <code>signer</code> address equals the <code>owner</code> hash of contract. This function SHOULD use the SYSCALL <code>Neo.Runtime.CheckWitness</code> to verify the <code>signer</code>. Details has been explained in [NEP-30](https://github.com/superboyiii/proposals/blob/upgrade-standard/nep-30.mediawiki).
 
+===References===
+
+https://github.com/neo-project/proposals/blob/master/nep-29.mediawiki
+
+https://github.com/neo-project/proposals/blob/master/nep-30.mediawiki
+
 ===Implementation===
 
 * C#: https://github.com/neo-project/neo-devpack-dotnet/blob/003a50095efe88f63f518a0c989921be086409f3/examples/Example.SmartContract.NEP17/SampleNep17Token.cs#L130
--- a/nep-22.mediawiki
+++ b/nep-22.mediawiki
@@ -46,7 +46,7 @@ Neo N3 has a native <code>ContractManagement</code> contract that is used for co
 }
 </pre>
 
-This method will be automatically executed by <code>ContractManagement</code> contract when a contract is updated. When <code>update</code> in <code>_deploy</code> is true, ContractManagement contract will update new contract state by this method.
+This method will be automatically executed by <code>ContractManagement</code> contract when a contract is updated. When <code>update</code> in <code>_deploy</code> is true, <code>ContractManagement</code> contract will update new contract state by this method.
 
 Updating a smart contract MUST have <code>nefFile</code> and<code>manifest.json</code>. It will be passed to <code>_deploy</code>.
 
--- a/nep-22.mediawiki
+++ b/nep-22.mediawiki
@@ -18,7 +18,7 @@ As the Neo blockchain scales, Smart Contract Update will become increasingly imp
 
 ==Specification==
 
-All contracts SHOULD implement this NEP in case of any bug hotfix or improvement. If no update method there, incorrect action can never be corrected.
+All contracts SHOULD implement this NEP to be able to fix bugs or otherwise improve functionality. If no update method there, incorrect action can never be corrected.
 
 ===Methods===
 
@@ -46,11 +46,11 @@ All contracts SHOULD implement this NEP in case of any bug hotfix or improvement
 }
 </pre>
 
-This method MUST CALL <code>update</code> method in native <code>ContractManagement</code> contract. It will update contract state if it's executed successfully.
+This method MUST call <code>update</code> method in native <code>ContractManagement</code> contract. It will update contract state if it's executed successfully.
 
-Updating a smart contract MUST have <code>nefFile</code> or <code>manifest.json</code>. How do they work is in reference of [NEP-15](https://github.com/neo-project/proposals/blob/master/nep-15.mediawiki) and [NEP-16](https://github.com/neo-project/proposals/blob/master/nep-16.mediawiki). Just updating one of them at a time is OK. This means no change on another, so another SHOULD be set as null. It MUST be passed to ContractManagement.Update() in this method.
+Updating a smart contract MUST have <code>nefFile</code> or <code>manifest.json</code>. How do they work is in reference of [https://github.com/neo-project/proposals/blob/master/nep-15.mediawiki NEP-15] and [https://github.com/neo-project/proposals/blob/master/nep-16.mediawiki NEP-16]. Just updating one of them at a time is OK. This means no change on another, so another SHOULD be set as null. It MUST be passed to ContractManagement.Update() in this method.
 
-The parameter <code>data</code> can be any type of supported parameters for contract-specific purpose. For [example](https://github.com/nspcc-dev/neofs-contract/blob/99fb86c35a48ed12017423aa4fee13f7d07fa4c0/contracts/proxy/contract.go#L33) some kinds of parameters can be put into data for specific check. Also some reference in [NEP-29](https://github.com/neo-project/proposals/blob/master/nep-29.mediawiki).
+The parameter <code>data</code> can be any type of supported parameters for contract-specific purpose. For [https://github.com/nspcc-dev/neofs-contract/blob/99fb86c35a48ed12017423aa4fee13f7d07fa4c0/contracts/proxy/contract.go#L33 example] some kinds of parameters can be put into data for specific check. Also some reference in [https://github.com/neo-project/proposals/blob/master/nep-29.mediawiki NEP-29].
 
 This method SHOULD do <code>Checkwitness</code> for contract <code>owner</code> to avoid security risks. This function SHOULD use the SYSCALL <code>Neo.Runtime.CheckWitness</code> to verify the <code>signer</code>. 
 
--- a/nep-22.mediawiki
+++ b/nep-22.mediawiki
@@ -18,7 +18,7 @@ As the Neo blockchain scales, Smart Contract Update will become increasingly imp
 
 ==Specification==
 
-All contracts SHOULD implement this NEP in case of any bug hotfix or improvement.
+All contracts SHOULD implement this NEP in case of any bug hotfix or improvement. If no update method there, incorrect action can never be corrected.
 
 ===Methods===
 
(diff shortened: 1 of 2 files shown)


nep-xx.mediawiki
```diff
@@ -0,0 +1,103 @@
...
+|                                                              |      | "type": "Boolean"                                            |
+|                                                              |      | }                                                            |
+|                                                              |      | ],                                                           |
+|                                                              |      | "returntype": "Void"                                         |
+|                                                              |      | }                                                            |
+|                                                              |      | </pre>                                                       |
+|                                                              |      |                                                              |
+|                                                              |      | This is an optional function which will be automatically executed when contract first deployed. |
```
ixje, line 52:
```suggestion
|                                                              |      | This is an optional function which will be automatically executed when a contract is first deployed or updated. |
```

nep-xx.mediawiki
```diff
@@ -0,0 +1,103 @@
...
+|                                                              |      |                                                              |
+|                                                              |      | The parameter <code>data</code> can be any type of supported parameters for additional message purpose. |
+|                                                              |      |                                                              |
+|                                                              |      | <code>update</code> returns if it's the initial deployment of this contract. The function SHOULD check whether <code>update</code> returns true. If not, continue execution. If true, return. |
```
ixje, line 56: > The function SHOULD check whether <code>update</code> returns true. If not, continue execution. If true, return.

I don't think this is correct advice. It seems perfectly reasonable to me to do something like this (pseudo)

```
deploy(data, update) {
   if !update {
      // initial deploy, put default values into storage
   } else {
      // reset part of storage to default
   }
}
```

nep-xx.mediawiki
```diff
@@ -0,0 +1,103 @@
...
+|                                                              |      | "type": "Any"                                                |
+|                                                              |      | }                                                            |
+|                                                              |      | ],                                                           |
+|                                                              |      | "returntype": "Void"                                         |
+|                                                              |      | }                                                            |
+|                                                              |      | </pre>                                                       |
+|                                                              |      |                                                              |
+|                                                              |      | Updating a smart contract MUST have <code>nefFile</code> and <code>manifest.json</code>. |
```
ixje, line 82: Why suggest the requirement that both manifest and NEF must be present? The management contract requires one of the two or both.
https://github.com/neo-project/neo/blob/5f4c14e3801c691ceccb5705c634e9e55dd3bba1/src/Neo/SmartContract/Native/ContractManagement.cs#L233

nep-xx.mediawiki
superboyiii: In fact, either `initial deploy` or `update` will send notification, no need to waste storage.
https://github.com/neo-project/neo/blob/5f4c14e3801c691ceccb5705c634e9e55dd3bba1/src/Neo/SmartContract/Native/ContractManagement.cs#L99

ixje: I'm aware of the existing notification, but that's not the issue I'm trying to point out. The way I'm understanding the text is that it suggests to do the following
```
_deploy(data, update) {
   if update { return }
   // else continue execution of remaining statements
}
```
if that's the case then the whole function should not have a need to have an `update` parameter to begin with. Since it does have this param, it is reasonable that a contract might do something different when updated vs when first deployed. The suggestion would therefore be to remove this text
> The function SHOULD check whether update returns true. If not, continue execution. If true, return.

superboyiii: Done

nep-xx.mediawiki
superboyiii: Done

ixje (dismissed):
LGTM

nep-xx.mediawiki
```diff
@@ -0,0 +1,101 @@
...
+
+
+
+==Abstract==
+
+This proposal outlines a basic method guideline for the NEO blockchain that will provide systems with a generalized interaction mechanism for smart contract initial deploy, update and destroy.
+
+==Motivation==
+
+As the NEO blockchain scales, Smart Contract Initial Deploy, Update and Destroy will become increasingly important.  Without guideline for <code>_deploy</code>, <code>update</code> and <code>destroy</code> method, systems will be required to maintain a unique API for each contract, regardless of their similarity to other contracts which makes application development very inconvenient. This guideline can lead most contracts to implement the same basic methods with the same types of parameters but still accept diversity.
```
roman-khimov, line 22: I'm not entirely sure we should put `_deploy` there on the same line as `update` or `destroy` because it's a system-defined method, not a convention (it's even briefly mentioned on https://docs.neo.org/docs/en-us/reference/scapi/framework/native/ContractManagement/Deploy.html). But maybe this NEP will make it somewhat more visible, probably not all developers even know about it.

nep-xx.mediawiki
```diff
@@ -0,0 +1,101 @@
...
+    {
+      "name": "update",
+      "type": "Boolean"
+    }
+  ],
+  "returntype": "Void"
+}
+</pre>
+
+This is an optional function which will be automatically executed when a contract is first deployed or updated.
```
roman-khimov, line 50: `...executed by ContractManagement contract...`?

nep-xx.mediawiki
```diff
@@ -0,0 +1,101 @@
...
+
+The parameter <code>data</code> can be any type of supported parameters for additional message purpose.
```
roman-khimov, line 52: `contract-specific purpose`? Mention that it's the same `data` that is accepted by ContractManagement's Deploy and Update?

nep-xx.mediawiki
```diff
@@ -0,0 +1,101 @@
...
+
+<code>update</code> returns if it's the initial deployment of this contract.
```
roman-khimov, line 54: I don't think `update` returns anything, it's just a `Boolean`, maybe `update is true when contract is updated (via Update) and it's false on initial deployment (via Deploy)`.

roman-khimov: And BTW, maybe it's important to tell that when contract is updated it's the _new_ contract's `_deploy` that is called, not old one.

nep-xx.mediawiki
```diff
@@ -0,0 +1,101 @@
...
+      "type": "Any"
+    }
+  ],
+  "returntype": "Void"
+}
+</pre>
+
+Updating a smart contract MUST have <code>nefFile</code> or <code>manifest.json</code> or <code>both</code>.
+
+The parameter <code>data</code> can be any type of supported parameters for additional message purpose.
```
roman-khimov, line 82: Mention that it will be passed to `_deploy`?

nep-xx.mediawiki
```diff
@@ -0,0 +1,101 @@
...
+
+The function SHOULD check whether the <code>signer</code> address equals the <code>owner</code> hash of contract. The function SHOULD use the SYSCALL <code>Neo.Runtime.CheckWitness</code> to verify the <code>signer</code>.
```
roman-khimov, line 84: BTW, we don't have any "ownability" NEP that will define what is "owner" and how to deal with this concept, but it's probably not critical for this NEP.

roman-khimov (commented):
It'd be nice to have this, I'd change our examples to use this standard interface.

nep-xx.mediawiki
superboyiii: Done

nep-xx.mediawiki
superboyiii: Done

nep-xx.mediawiki
superboyiii: Done

nep-xx.mediawiki
superboyiii: Done

nep-xx.mediawiki
superboyiii: Done

nep-xx.mediawiki
superboyiii: Done

roman-khimov (approved):

nep-xx.mediawiki
```diff
@@ -0,0 +1,102 @@
...
+
+This is an optional function which will be automatically executed by ContractManagement contract when a contract is first deployed or updated.
+
+The parameter <code>data</code> can be any type of supported parameters for contract-specific purpose. It's the same <code>data</code> that is accepted by Deploy() and Update() in ContractManagement contract.
+
+<code>update</code> is true when contract is updated via <code>update()<code> and it's false on initial deployment via <code>_deploy()<code>.
+
+====update====
+
+<pre>
```
shargon, line 58: In the view https://github.com/neo-project/proposals/pull/154/files?short_path=6caa849#diff-6caa8493f36ba42782e59054f66c45a43f6c170927ce9383eccda13535890ea1 this region is wrong displayed

superboyiii: Fixed.

erikzhang:
Add `_initialize`?

roman-khimov:
`_initialize` is something different to me, not directly related to deploy/update. Maybe it's in some other document (but I wasn't able to find it on https://docs.neo.org/) or other NEP.

superboyiii:
@roman-khimov @erikzhang Added. Please review again.

roman-khimov:
NEP-22:

> provide systems with a generalized interaction mechanism for smart contract initial deploy, update and destroy. 

I don't see `_initialize` being a part of this NEP.

erikzhang:
> I don't see _initialize being a part of this NEP.

I'm not sure if it should be part of this PR, but obviously it should be included in a document.

Also, don't forget the `verify` method.

superboyiii:
> > I don't see _initialize being a part of this NEP.
> 
> I'm not sure if it should be part of this PR, but obviously it should be included in a document.
> 
> Also, don't forget the `verify` method.

Added.

superboyiii:
@erikzhang @roman-khimov Is it OK?

roman-khimov:
It's a method soup now to me. I'd rather have separate NEPs for `_initialize`, `_deploy` and `verify` (three NEPs, seems to be perfectly fine to me, they are different unrelated things) and one for `update`/`destroy`. I want to be able to specify "NEP-XYZ" in `SupportedStandards` and expect all the tooling around to understand that this contract has standardized `update`/`destroy` method pair that can be interacted with. "NEP-22" doesn't give this meaning right now.

nep-xx.mediawiki
roman-khimov: Related to #159, btw.

ixje:
> `_initialize` is something different to me, not directly related to deploy/update. Maybe it's in some other document (but I wasn't able to find it on https://docs.neo.org/) or other NEP.

I can kind of see the relationship. `update` can call `_deploy` which in turn can call `_initialize` again. Ideally there would be a flow diagram included to make it clear. `verify` on the other hand seems totally unrelated and worthy of a separate document if we're splitting anyway

roman-khimov:
> I can kind of see the relationship

Is it related? Probably so. Can be it be used _without_ all the other things? Sure it is. Can all the other things in this proposal be used without `_initialize`? Sure they are. We can have as many standards as we need and they can be related, but different standards for different things. `NEP-XYZ` in `SupportedStandards` should have some specific meaning. And `SupportedStandards` can have as many `NEP-X, NEP-Y, NEP-Z` as needed.

EdgeDLT:
> And SupportedStandards can have as many NEP-X, NEP-Y, NEP-Z as needed.

I agree in principle with the idea of having independent standards. But I suspect there are many others like myself who also have a mental block at the future thought of these long lists of numbers with no meaning. Maybe we could package them together under a standard group, but at that point, is it that much different from what we have now?

cschuchardt88:
> > I can kind of see the relationship
> 
> Is it related? Probably so. Can be it be used _without_ all the other things? Sure it is. Can all the other things in this proposal be used without `_initialize`? Sure they are. We can have as many standards as we need and they can be related, but different standards for different things. `NEP-XYZ` in `SupportedStandards` should have some specific meaning. And `SupportedStandards` can have as many `NEP-X, NEP-Y, NEP-Z` as needed.

_initialize is used when verifying a witness and/or in invocationScripts & verificationScript.  It gets initialized before invoking those scripts. I don't know if there is a standard for that function may need to create one. I know neo uses it internal for the above mentioned. also anytime contract is invoked. I do believe the attribute "InitializeValue" creates this function, don't recall. Would have to check

nep-22.mediawiki
```diff
@@ -0,0 +1,123 @@
...
+  "name": "update",
+  "safe": false,
+  "parameters": [
+    {
+      "name": "nefFile",
+      "type": "ByteArray"
+    },
+    {
+      "name": "manifest",
+      "type": "String"
```
roman-khimov, line 91: How come it's a `String` btw? ContractManagement itself expects a `ByteArray`.

cschuchardt88: It doesn't matter for arguments. In the vm its already byte array; to be specific `ByteString`; can be anything for the most part. It all get compiled down to byte array.

roman-khimov: That's not the point, these are different things as per NEP-14 and one can expect some type symmetry between `ContractManagement` definition of `update` and this one. Manifests matter, especially if you try to generate some code using them.

cschuchardt88: Then why not make a `ByteString`?

superboyiii: > That's not the point, these are different things as per NEP-14 and one can expect some type symmetry between `ContractManagement` definition of `update` and this one. Manifests matter, especially if you try to generate some code using them.

Fixed

roman-khimov:
My take on it is still the same, three NEPs here would be much better than a single one:

> It's a method soup now to me. I'd rather have separate NEPs for `_initialize`, `_deploy` and `verify` (three NEPs, seems to be perfectly fine to me, they are different unrelated things) and one for `update`/`destroy`. I want to be able to specify "NEP-XYZ" in `SupportedStandards` and expect all the tooling around to understand that this contract has standardized `update`/`destroy` method pair that can be interacted with. "NEP-22" doesn't give this meaning right now.

This one was supposed to be about update/destroy initially.

nep-22.mediawiki
```diff
@@ -0,0 +1,123 @@
...
+
+<code>update</code> is true when contract is updated via <code>update()</code> and it's false on initial deployment via <code>_deploy()</code>.
+
+====verify====
+
+<pre>
+{
+  "name": "verify",
+  "safe": false,
+  "parameters": [],
```
shargon, line 67: In fact we accept verify with args too

nep-22.mediawiki
```diff
@@ -0,0 +1,123 @@
...
+  Title: Contract Basic Methods Guideline
+  Author: Owen Zhang <zhangzhihong@ngd.neo.org>, Fernando Díaz Toledano <shargon@gmail.com>, Erik Zhang <erik@neo.org>
+  Type: Informational
+  Status: Accepted
+  Created: 2021-04-19
+</pre>
+
+==Abstract==
+
+This proposal outlines a basic method guideline for the NEO blockchain that will provide systems with a generalized interaction mechanism for smart contract initial deploy, update and destroy.
```
shargon, line 12: and verify 
```suggestion
This proposal outlines a basic method guideline for the NEO blockchain that will provide systems with a generalized interaction mechanism for smart contract initial deploy, update, verify and destroy.
```

superboyiii:
@shargon done

roman-khimov (changes requested):
https://github.com/neo-project/proposals/pull/154#issuecomment-1988307672

I can't imagine how to use this NEP in its current form. It needs a split.

Jim8y:
> [#154 (comment)](https://github.com/neo-project/proposals/pull/154#issuecomment-1988307672)
> 
> I can't imagine how to use this NEP in its current form. It needs a split.

@roman-khimov this one can be the most general description of base methods, as for specific method that should have a seperate NEP, we can create another NEPs for them, like Verify.

roman-khimov:
Let's get back to #135 that started all of this for a moment. What people really wanted there is a _convention_ for methods that contract authors can use to have some common way to update/destroy contracts. In this sense it's somewhat similar to NEP-11/17/24. And this is only about two methods: `update` and `destroy`, so that I wouldn't want to do
```
func obnovlyashka(data Array)
```
to update my contract (because 3rd-party tools couldn't easily create an update transaction for me in this case, unlike when they deal with a compliant `update`).

But what we also have here now are:
 * `verify`, which has nothing to do with update/destroy, it's just a different mechanism
 * `_deploy`, which _is_ related to `update`, but is still somewhat different in that it can be omitted completely, it can be implemented in a contract that doesn't have `update` and, most importantly, it's not something you can call externally yourself, it's a native contract mechanism

At least we need to move `verify` out. It'd be perfect to have a NEP for it, but it's a separate matter. I would also move `_deploy` out since it has a broader definition (not just updates, but deployments as well, works irrespective of how your exposed-from-contract update methods are called), then this NEP could just refer to it.

AnnaShaleva:
Totally agree with Roman, we should update this NEP to standardize a single update method. The rest of methods standards should be moved to a separate granular NEPs.

superboyiii:
@roman-khimov @shargon @cschuchardt88 @Jim8y Now I changed it to `Contract Update Standard`. I will submit `destroy standard` as well.

nep-22.mediawiki
```diff
@@ -0,0 +1,67 @@
...
+    {
+      "name": "data",
+      "type": "Any"
+    }
+  ],
+  "returntype": "Void"
+}
+</pre>
+
+This method will be automatically executed by ContractManagement contract when a contract is updated. When <code>update</code> in <code>_deploy</code> is true, ContractManagement contract will update new contract state by this method.
```
shargon, line 49:
```suggestion
This method will be automatically executed by <code>ContractManagement</code> contract when a contract is updated. When <code>update</code> in <code>_deploy</code> is true, ContractManagement contract will update new contract state by this method.
```

nep-22.mediawiki
```diff
@@ -0,0 +1,67 @@
...
+      "type": "Any"
+    }
+  ],
+  "returntype": "Void"
+}
+</pre>
+
+This method will be automatically executed by <code>ContractManagement</code> contract when a contract is updated. When <code>update</code> in <code>_deploy</code> is true, <code>ContractManagement</code> contract will update new contract state by this method.
+
+Updating a smart contract MUST have <code>nefFile</code> and<code>manifest.json</code>. It will be passed to <code>_deploy</code>.
```
shargon, line 51: In fact, you can send one of them as NULL, but not both

https://github.com/neo-project/neo/blob/978828005474dce455fe368e6e275f30293a5411/src/Neo/SmartContract/Native/ContractManagement.cs#L274

you can update only the manifest

nep-22.mediawiki
```diff
@@ -0,0 +1,67 @@
...
+</pre>
+
+
+==Abstract==
+
+This proposal outlines a standard for the NEO blockchain that will provide systems with a generalized interaction mechanism for smart contract update.
+
+==Motivation==
+
+As the NEO blockchain scales, Smart Contract Update will become increasingly important. Update mechanism is already there from initial N3 release. But we haven't summarized that standard. That's why this standard is necessary.
```
roman-khimov, line 17: It's more like we didn't have an agreement on how deployed contracts expose this functionality to their admins. Because ContractManagement itself doesn't care about code that calls it.

nep-22.mediawiki
```diff
@@ -0,0 +1,67 @@
...
+
+==Specification==
+
+Neo N3 has a native <code>ContractManagement</code> contract that is used for contract deployment and update via <code>deploy</code> and <code>update</code> methods of it. After settling with the new NEF and manifest both of them will invoke a special <code>_deploy</code> method if it's implemented by the contract.
```
roman-khimov, line 21: Reference NEP-29 for details here.

nep-22.mediawiki
```diff
@@ -0,0 +1,67 @@
...
+    {
+      "name": "data",
+      "type": "Any"
+    }
+  ],
+  "returntype": "Void"
+}
+</pre>
+
+This method will be automatically executed by <code>ContractManagement</code> contract when a contract is updated. When <code>update</code> in <code>_deploy</code> is true, <code>ContractManagement</code> contract will update new contract state by this method.
```
roman-khimov, line 49: That's not how it works. It's an externally (by transaction) called method that calls ContractManagement internally. Then it does the magic.

nep-22.mediawiki
```diff
@@ -0,0 +1,67 @@
...
+
+Updating a smart contract MUST have <code>nefFile</code> and<code>manifest.json</code>. It will be passed to <code>_deploy</code>.
+
+The parameter <code>data</code> can be any type of supported parameters for contract-specific purpose. For example some kinds of json parameters can be put into data for specific check.
+
+This function will check whether the <code>signer</code> address equals the <code>owner</code> hash of contract. This function SHOULD use the SYSCALL <code>Neo.Runtime.CheckWitness</code> to verify the <code>signer</code>. Details has been explained in [NEP-30](https://github.com/superboyiii/proposals/blob/upgrade-standard/nep-30.mediawiki).
```
roman-khimov, line 55: SHOULD check, this can be done (or not done) in various ways.

superboyiii:
@shargon @roman-khimov  Done.

nep-22.mediawiki
```diff
@@ -0,0 +1,67 @@
...
+    {
+      "name": "data",
+      "type": "Any"
+    }
+  ],
+  "returntype": "Void"
+}
+</pre>
+
+This method will will callback <code>Update</code> method in native <code>ContractManagement</code> contract. It will update contract state if it's exectued successfully.
```
vncoelho, line 49: will will

nep-22.mediawiki
```diff
@@ -0,0 +1,67 @@
...
+
+Updating a smart contract MUST have <code>nefFile</code> or <code>manifest.json</code>. If just update one of them at a time is OK, another SHOULD be set as null. It will be passed to <code>_deploy</code>.
+
+The parameter <code>data</code> can be any type of supported parameters for contract-specific purpose. For example some kinds of json parameters can be put into data for specific check.
```
vncoelho, line 53: "For example some kinds of json parameters can be put into data for specific check."
Some examples here?

nep-22.mediawiki
```diff
@@ -0,0 +1,67 @@
...
+
+This function SHOULD check whether the <code>signer</code> address equals the <code>owner</code> hash of contract. This function SHOULD use the SYSCALL <code>Neo.Runtime.CheckWitness</code> to verify the <code>signer</code>. Details has been explained in [NEP-30](https://github.com/superboyiii/proposals/blob/upgrade-standard/nep-30.mediawiki).
```
vncoelho, line 55: This function? Which function?

nep-22.mediawiki
```diff
@@ -0,0 +1,67 @@
...
+</pre>
+
+
+==Abstract==
+
+This proposal outlines a standard for the NEO blockchain that will provide systems with a generalized interaction mechanism for smart contract update.
+
+==Motivation==
+
+As the NEO blockchain scales, Smart Contract Update will become increasingly important. We need standize <code>update</code> method in contract to callback native <code>ContractManagement</code> correctly.
```
roman-khimov, line 17: `s/NEO/Neo`. `s/standize/a standard/`? I'm also not sure we need to say anything about callbacks. It's about standard API to update contracts, what it does internally is not motivation probably.

nep-22.mediawiki
```diff
@@ -0,0 +1,67 @@
...
+
+==Specification==
+
+Neo N3 has a native <code>ContractManagement</code> contract that is used for contract deployment and update via <code>deploy</code> and <code>update</code> methods of it. After settling with the new NEF and manifest both of them will invoke a special <code>_deploy</code> method if it's implemented by the contract. More details in [NEP-29](https://github.com/neo-project/proposals/blob/master/nep-29.mediawiki)
```
roman-khimov, line 21: I don't think it's a part of specification. We're specifying something different here.

nep-22.mediawiki
roman-khimov: Probably just `MUST call`

nep-22.mediawiki
```diff
@@ -0,0 +1,67 @@
...
+      "type": "Any"
+    }
+  ],
+  "returntype": "Void"
+}
+</pre>
+
+This method will will callback <code>Update</code> method in native <code>ContractManagement</code> contract. It will update contract state if it's exectued successfully.
+
+Updating a smart contract MUST have <code>nefFile</code> or <code>manifest.json</code>. If just update one of them at a time is OK, another SHOULD be set as null. It will be passed to <code>_deploy</code>.
```
roman-khimov, line 51: Maybe refer to NEP-16 and NEP-15 here. I also don't understand 

> It will be passed to <code>_deploy</code>.

here. Only `data` is going to be passed and that's exactly where a reference to NEP-29 is needed.

nep-22.mediawiki
roman-khimov: Maybe better as `this method`, since we're talking about `update` _method_ here. Also, I don't understand what is to `check whether the <code>signer</code> address equals the <code>owner</code> hash of contract`. Transaction can have a lot of signers. We should care more about owner allowing this action (`CheckWitness`). And NEP-30 is a different thing in general, _most_ of contracts will not have contract-based owners, they're usually just multisig.

superboyiii:
All done @roman-khimov @vncoelho

nep-22.mediawiki
```diff
@@ -0,0 +1,73 @@
...
+
+This proposal outlines a standard for the NEO blockchain that will provide systems with a generalized interaction mechanism for smart contract update.
+
+==Motivation==
+
+As the Neo blockchain scales, Smart Contract Update will become increasingly important. We need a standard for <code>update</code> method in contract to callback native <code>ContractManagement</code> correctly.
+
+==Specification==
+
+All contracts SHOULD implement this NEP in case of any bug hotfix or improvement.
```
vncoelho, line 21: Maybe explain current consequences if not implemented in the first version.

nep-22.mediawiki
```diff
@@ -0,0 +1,73 @@
...
+    {
+      "name": "data",
+      "type": "Any"
+    }
+  ],
+  "returntype": "Void"
+}
+</pre>
+
+This method MUST callback <code>Update</code> method in native <code>ContractManagement</code> contract. It will update contract state if it's exectued successfully.
```
vncoelho, line 49: exectued

nep-22.mediawiki
```diff
@@ -0,0 +1,73 @@
...
+
+Updating a smart contract MUST have <code>nefFile</code> or <code>manifest.json</code>, see reference of [NEP-15](https://github.com/neo-project/proposals/blob/master/nep-15.mediawiki) and [NEP-16](https://github.com/neo-project/proposals/blob/master/nep-16.mediawiki). If just update one of them at a time is OK, another SHOULD be set as null. It MUST be passed to ContractManagement.Update() in this method.
```
vncoelho, line 51: I think maybe more details on the consequences here. I mean, the differences that can happen if updated in two calls.

nep-22.mediawiki
Jim8y: executed？

nep-22.mediawiki
superboyiii: We can guide them to read the relevant NEP standards listed below.

nep-22.mediawiki
superboyiii: Done

nep-22.mediawiki
superboyiii: Done

nep-22.mediawiki
```diff
@@ -18,7 +18,7 @@ As the Neo blockchain scales, Smart Contract Update will become increasingly imp
 
 ==Specification==
 
-All contracts SHOULD implement this NEP in case of any bug hotfix or improvement.
+All contracts SHOULD implement this NEP in case of any bug hotfix or improvement. If no update method there, incorrect action can never be corrected.
```
vncoelho, line 21: good

vncoelho (dismissed):

nep-22.mediawiki
roman-khimov: It can be implemented in a different way, I mean non-NEP-22 method that still does the update. Also, the way it's worded now is like "if you have a bug, please implement this method", while in fact it's too late if you have a bug and no way to update the contract.

superboyiii:
@roman-khimov Done

nep-22.mediawiki
```diff
@@ -46,13 +46,13 @@ All contracts SHOULD implement this NEP in case of any bug hotfix or improvement
 }
 </pre>
 
-This method MUST callback <code>Update</code> method in native <code>ContractManagement</code> contract. It will update contract state if it's executed successfully.
+This method MUST CALL <code>update</code> method in native <code>ContractManagement</code> contract. It will update contract state if it's executed successfully.
```
roman-khimov, line 49: s/CALL/call/

nep-22.mediawiki
```diff
@@ -0,0 +1,71 @@
...
+https://github.com/neo-project/proposals/blob/master/nep-15.mediawiki
+
+https://github.com/neo-project/proposals/blob/master/nep-16.mediawiki
+
+https://github.com/neo-project/proposals/blob/master/nep-29.mediawiki
+
+===Implementation===
+
+* C#: https://github.com/neo-project/neo-devpack-dotnet/blob/003a50095efe88f63f518a0c989921be086409f3/examples/Example.SmartContract.NEP17/SampleNep17Token.cs#L130
+
```
roman-khimov, line 70: I think we don't need a space in this list, although it renders fine in the end.

nep-22.mediawiki
```diff
@@ -0,0 +1,71 @@
...
+      "type": "Any"
+    }
+  ],
+  "returntype": "Void"
+}
+</pre>
+
+This method MUST CALL <code>update</code> method in native <code>ContractManagement</code> contract. It will update contract state if it's executed successfully.
+
+Updating a smart contract MUST have <code>nefFile</code> or <code>manifest.json</code>. How do they work is in reference of [NEP-15](https://github.com/neo-project/proposals/blob/master/nep-15.mediawiki) and [NEP-16](https://github.com/neo-project/proposals/blob/master/nep-16.mediawiki). Just updating one of them at a time is OK. This means no change on another, so another SHOULD be set as null. It MUST be passed to ContractManagement.Update() in this method.
```
roman-khimov, line 51: Unfortunately, it's not markdown, see how it renders with "view file" on GH, links need to be fixed (see https://github.com/neo-project/proposals/blob/a88f56086f8ca607af2c0be3c0d87dec2b3ca4ca/nep-25.mediawiki?plain=1#L252 and other NEPs).

nep-22.mediawiki
```diff
@@ -0,0 +1,71 @@
...
+
+This proposal outlines a standard for the NEO blockchain that will provide systems with a generalized interaction mechanism for smart contract update.
+
+==Motivation==
+
+As the Neo blockchain scales, Smart Contract Update will become increasingly important. We need a standard for <code>update</code> method in contract to callback native <code>ContractManagement</code> correctly.
+
+==Specification==
+
+All contracts SHOULD implement this NEP in case of any bug hotfix or improvement. If no update method there, incorrect action can never be corrected.
```
roman-khimov, line 21: Maybe `... implement this NEP to be able to fix bugs or otherwise improve functionality`. Just to not be misinterpreted for "implement after you have a bug", no, it's "think about it before you have bugs".

roman-khimov (commented):
Editorial nitpicks, but OK otherwise.

nep-22.mediawiki
superboyiii: C# example is necessary as well as GO.

superboyiii:
@roman-khimov Done

roman-khimov (approved):

Jim8y (approved):

nep-22.mediawiki
```diff
@@ -0,0 +1,71 @@
+<pre>
+  NEP: 22
+  Title: Contract Update Function
+  Author: Owen Zhang <zhangzhihong@ngd.neo.org>, Fernando Díaz Toledano <shargon@gmail.com>, Erik Zhang <erik@neo.org>, Roman Khimov <roman@nspcc.ru>
+  Type: Standard
+  Status: Accepted
```
AnnaShaleva, line 6: This standard is accepted, shall we update the README.md?

nep-22.mediawiki
```diff
@@ -0,0 +1,71 @@
...
+  Created: 2021-04-19
+</pre>
+
+
+==Abstract==
+
+This proposal outlines a standard for the NEO blockchain that will provide systems with a generalized interaction mechanism for smart contract update.
```
AnnaShaleva, line 13:
```suggestion
This proposal outlines a standard for the Neo blockchain that will provide systems with a generalized interaction mechanism for smart contract update.
```

AnnaShaleva (approved):

AnnaShaleva (approved):
Sorry for this minor fix, I couldn't resist, and now all previous approvals are gone.  @Jim8y, @roman-khimov, need your approvals one more time.

nep-22.mediawiki
roman-khimov: I think we better merge this, #182 and then update the README, because there are some omissions there anyway.

roman-khimov (approved):

nep-22.mediawiki
roman-khimov: The scheme is exactly to avoid endless reapprovals, the document is fine now and that's the main thing here.

Jim8y:
> Sorry for this minor fix, I couldn't resist, and now all previous approvals are gone. @Jim8y, @roman-khimov, need your approvals one more time.

That is totally ok to comment on minor fixes, you are the strongest sheld we have to ensure eveything works properly, your concern definately will be addressed after the merge. Just approve an NEP takes too long while fixing minor issues can be in another track, along with file name/nep number/readme.

Issue #152, opened 2022-07-18
NEP-17 extension discussion


vincentgeneste:
**Summary:**
I would like to open a discussion to add some kind of extension to NEP-17 standard, maybe NEP-17.1 or any name that may fit, which would allow/replicate some features from ERC-20 standard and the allowance mechanism, which i think are lacking in NEP-17 standard: `allowance`, `transferFrom`, and `approve`, as well as one event related to that: `approval`. The whole idea is to allow a **spender** to spend x **amount** of funds on behalf of an **owner**.

**Goal:**
The goal is neither to replace NEP-17, nor to create a fully separate standard, but more to add some kind of extension to the current one, which project could signal their intention to support or not.

**Usecase:**
Our own usecase is simple: this allows us to process NFT trades on behalf of users, without them having to lock their tokens in advance, what we call open offers. Without this, if someone wanted to place 10 offers on 10 different NFT for 10 NEO each, he would have to lock 100 NEO without even being sure those would be accepted. With this extension, the NEO only move from the buyer's wallet if and when a trade is made (when an offer is accepted), it is effectively way more convenient.
There are of course lots more usecase than that, most of them for defi applications.

**Risks:**
Obviously the concept of approval and allowance is complex and prone to security risks, as people giving a too high / unlimited allowance could end up with their NEP-17 drained, but that's solved with user education, learning how to revoke and check allowance etc, and I personally think that the upside out-weights this risk.

**Feedback:**
This is a feature we already have implemented in our project and which work very well for us and our own NEP-17, and the goal is that this feature would be opt-in for any project that want to implement it on their NEP-17 token contract. 

**Technical details:**
If a project decide to implement it, it would be as easy as adding a `supported_standards = ["NEP-17", "NEP-17-1"]` which then any project could check on ContractManagement Manifest
```
                var contractDetails = ContractManagement.GetContract(contract);
                string stringToCheck = "NEP-17-1";
                bool isCompatible = false;
                foreach (var s in contractDetails.Manifest.SupportedStandards)
                {
                    if (s == stringToCheck)
                    {
                        isCompatible = true;
                        break;
                    }
                }
                return isCompatible;
```

Regarding the implementation itself, we currently use it like this (in python) and tried to make it very similar to ERC-20 to help with adoption. We only changed a few NEO-specific things obviously (passing spender as an arg instead of using tx.Sender, etc).

```
@public(safe=True)
def allowance(from_address: UInt160, spender: UInt160) -> int:
    """
[...]

PR #155, opened 2022-09-02
Add Royalty standard

superboyiii:


roman-khimov:
Standards work when they're standards, that is you say "NEP-11" and everyone has the same understanding of what it is. It's impossible to have this if the definition of "NEP-11" changes. Please, do not change NEP-11 or any other standard. We can have as many standards as needed, this method itself can be a separate standard for example. Or we can obsolete NEP-11 in favor of new NEP-X including this method. It'd be much easier to handle for everyone then.

erikzhang:
But it doesn't break anything and is fully compatible with previous standards. In this case I don't think it's necessary to create a new standard.

superboyiii:
> But it doesn't break anything and is fully compatible with previous standards. In this case I don't think it's necessary to create a new standard.

I agree with Roman, we may add more and more standards on NFT related issues, but we should not add them all in NEP-11 since NEP-11 already has completed. We can't iterate other features on it, NEP-11 doesn't have v1.0, 2.0, 3.0... We should make it clear, different standard should apply on different features, they shouldn't be modified when it's completed. It's also better for developers to know which is doing what.

EdgeDLT:
Agree with the above. NEP-11 is the base class. Some NFTs might want to extend that with additional methods for loyalties, but others may have no reason to implement them. It makes sense to keep the standards separate.

The same is true for https://github.com/neo-project/proposals/issues/152 and NEP-17 tokens, also! We shouldn't want to change NEP-17 to add approve/transferFrom/allowance, but adding a new NEP to standardize these methods for NEP-17 tokens that do want to use that functionality makes complete sense.

superboyiii:
@roman-khimov @erikzhang @EdgeDLT Done.

vincentgeneste:
Great idea to try and make this a standard feature :) we've been actively trying to get this done since so many months!

Couple questions / comments :

Based on the naming of methods / args, i guess you're trying to replicate EIP-2981, but in their case its a per NFT royalty settings, while in your proposal, there's no args, so i assume you want to make it a per contract setting? I would definitely think a per NFT is what's required for this proposal (so the tokenId should have to be passed).

Any reason not to add contractHash in `ReceivedRoyalties` event?

Most importantly, in most places, whether NEO or EVM, all fees are expressed in BPS (1000 -> 10%) directly. I think for convenience purposes and to help onboarding other devs, it should be the same here.

superboyiii:
@merl111 @vincentgeneste  Done.

vincentgeneste:
> @merl111 @vincentgeneste Done.

Couple more comments from me... :)

- your second `royaltyInfo` implementation has safe : false, should be safe : true
- typo `founction`
- we should incorporate multi recipients. While EIP2981 does not support it, it's actually one of the main drawbacks of using it, being limited to only one recipient. So we should allow an array of recipient/values, as well as emit one event per royalty paid
- i do not think `isGlobalRoyalty` is usefull at all. at the end of the day what we want is to know what the royalty is for one NFT. whether it is a global value or not is irrelevent we have to call the contract. So for simplification, i would get rid of it completely and just use your second implementation with `royaltyInfo / tokenId / salePrice`. It is up to the NEP11 contract to define if its global or not, the standard should not really care about it.

superboyiii:
> your second `royaltyInfo` implementation has safe : false, should be safe : true

In second situation, we have a dynamic `royaltyInfo` based on `salePrice`, creator could add his own logic to define how much `royaltyAmount` could be in different `salePrice`, so it should be not safe.

> typo `founction`

Fixed.

> i do not think isGlobalRoyalty is usefull at all. whether it is a global value or not is irrelevent we have to call the contract.

Yes, we have to call the contract, but not necessary to call it for each tokenId. I mean there should be two scenarios: Simple enough or flexible enough. In simple scenario: creators or artists have no requirements for different royalties in different tokenId in one collection, so I'd like to give a simple solution for marketplace to get a clear flag if the royalty is global. It can greatly reduce unnecessary invocation on nodes. If a contract have global royalty, still have to input `tokenId`,  `salePrice` but return the same storage, it's too strange. These params are unnecessary.
For the second scenario, we give a more flexible solution which EIP-2981 already done. (Specific royalty for different tokenId or even dynamic for different salePrice).

vincentgeneste:
> > your second `royaltyInfo` implementation has safe : false, should be safe : true
> 
> In second situation, we have a dynamic `royaltyInfo` based on `salePrice`, creator could add his own logic to define how much `royaltyAmount` could be in different `salePrice`, so it should be not safe.
> 
> > typo `founction`
> 
> Fixed.
> 
> > i do not think isGlobalRoyalty is usefull at all. whether it is a global value or not is irrelevent we have to call the contract.
> 
> Yes, we have to call the contract, but not necessary to call it for each tokenId. I mean there should be two scenarios: Simple enough or flexible enough. In simple scenario: creators or artists have no requirements for different royalties in different tokenId in one collection, so I'd like to give a simple solution for marketplace to get a clear flag if the royalty is global. It can greatly reduce unnecessary invocation on nodes. If a collection(1 contract, different tokenId) have global royalty, still have to input `tokenId`, `salePrice` but return the same storage, it's too strange. These params are unnecessary. For the second scenario, we give a more flexible solution which EIP-2981 already done. (Specific royalty for different tokenId or even dynamic for different salePrice).

Regarding the safe tag, i still think it should be set to true. Even if someone adds his own logic, it should not require write storage access, as we re just trying to get read only data here,i think it's more "safe" to have it set to safe: true :D 

For the `isGlobalRoyalty` i understand the idea behind it personally, it is to propose two scenarios, i just thought personally that this complexified the standard, for very low optimization, but maybe it is just me, i will let others comment on this!

superboyiii:
> Regarding the safe tag, i still think it should be set to true. Even if someone adds his own logic, it should not require write storage access

Yes, you're right, fixed.

> For the `isGlobalRoyalty` i understand the idea behind it personally, it is to propose two scenarios, i just thought personally that this complexified the standard, for very low optimization, but maybe it is just me, i will let others comment on this!

Yes, maybe we can get more feedback to see if it's necessary.

nep-x1.mediawiki
```diff
@@ -0,0 +1,137 @@
...
+    {
+      "name": "salePrice",
+      "type": "Integer"
+    }
+  ],
+  "returntype": "Array"
+}
+</pre>
+
+Returns a JSON Array which includes <code>royaltyRecipient</code> and <code>royaltyAmount</code>.
```
roman-khimov, line 72: Why is it a `JSON` if it's clearly a NeoVM array stack item with two elements inside?

nep-x1.mediawiki
```diff
@@ -0,0 +1,137 @@
...
+
+If return <code>false</code>, then it has specific royalty. MUST store specific royalty amount for each tokenId, static or dynamic, either is OK .
+
+====royaltyInfo====
+
+If <code>isGlobalRoyalty</code> return true, this function should be:
+
+<pre>
+{
+  "name": "royaltyInfo",
```
roman-khimov, line 45: To me, dropping `isGlobalRoyalty` and making just one `royaltyInfo` would make this simpler, global royalty will just have one line implementation of `royaltyInfo`, OK, but without a need for an additional method.

nep-x1.mediawiki
superboyiii: Done.

nep-x1.mediawiki
superboyiii: I removed `isGlobalRoyalty` for better compatiblity with EIP-2981

vincentgeneste:
@superboyiii what about supporting multiple recipients ? Is this something we want ? Personally I would love to have it as I see many different use cases for this. It’s not here on EIP2981 , but it’s highly likely because of limitations with multiple transactions in one transaction on EVM, which is not an issue at all on NeoVM.

superboyiii:
@erikzhang @shargon Could you have a review?

shargon (commented):
Any method to transfer the royalty to other account?

superboyiii:
> Any method to transfer the royalty to other account?

Yes, this could exist. But since it's customized person by person, we don't need a standard for `set`, we just do it for `get`.

superboyiii:
@erikzhang It's time to move forward.

superboyiii:
> @dotnet-policy-service agree [company="{your company}"]

@dotnet-policy-service agree [company="{Neo}"]

superboyiii:
@erikzhang @shargon Let's move on.

nep-x1.mediawiki
```diff
@@ -0,0 +1,161 @@
...
+        }
+    ]
+]
+</pre>
+
+<code>royaltyToken</code> is the token hash for royalty payment.
+
+<code>salePrice</code> MUST be amount without decimals.
+
+<code>royaltyRecipient</code> is the address of who should be sent the royalty payment, SHOULD be a 20-byte address.
```
shargon, line 109:
```suggestion
<code>royaltyRecipient</code> is the address of who will be sent the royalty payment, MUST be a 20-byte address.
```

superboyiii: Done @shargon

shargon (approved):

superboyiii:
Need votes for going on or not @roman-khimov @shargon @Jim8y @AnnaShaleva @cschuchardt88 @vncoelho

Jim8y (approved):

vncoelho:
I agree as well, i will do a final read this week.

shargon:
It must change also the readme, isn't it?

superboyiii:
I'm not sure which NEP NO. should pick. If you guys think it's OK, let's decide the NO. And then I will modify readme as well. @shargon

cschuchardt88:
> I'm not sure which NEP No. should pick. If you guys think it's OK, let's decide the No. And then I will modify readme as well. @shargon

It normally is the `github` issue number `id` of the issue that was created. In, this case its would be `NEP-155`.

superboyiii:
> @superboyiii
> 
> > I'm not sure which NEP No. should pick. If you guys think it's OK, let's decide the No. And then I will modify readme as well. @shargon
> 
> It normally is the `github` issue number `id` of the issue that was created. In, this case its would be `NEP-155`.

NEP is a bit special, current standards are a series of number 1 by 1. But I like your idea, this could avoid conflict. @roman-khimov @shargon @Jim8y @AnnaShaleva @cschuchardt88 @vncoelho Is `NEP-155` OK?

roman-khimov:
It doesn't fit current NEP-1, so I doubt it's an improvement. However, this can be a solution to what @Jim8y desperately wants --- reference to unfinished NEP. Can be like NEPR-155 (NEP request 155).

superboyiii:
I think NEO's proposals currently are not frequently created, so we can still follow NEP-1. If more PRs there and iterated quickly, let's make a short proposal like [this](https://github.com/neo-project/proposals/pull/155#issuecomment-1999234326). Now I need `[accepted]` and `[NEP-Standard]` label, so I could ensure it's passed and quickly name it as NEP-23 (now the latest is NEP-22). @roman-khimov @shargon @Jim8y @AnnaShaleva @cschuchardt88 @vncoelho

nep-x1.mediawiki
```diff
@@ -0,0 +1,161 @@
...
+<pre>
+{
+  "name": "RoyaltiesTransferred",
+  "parameters": [
+    {
+      "name": "royaltyToken",
+      "type": "Hash160"
+    },
+    {
+      "name": "royaltyRecipient",
```
vncoelho, line 136: Here, how is the event for "Multi royaltyRecipient"?

vncoelho (commented):
Looks good and precise, just some minor question on the event.

nep-x1.mediawiki
```diff
@@ -0,0 +1,161 @@
...
+  Created: 2022-09-05
+</pre>
+
+==Abstract==
+
+This NEP defines a global standard to get royalty payment information for non-fungible tokens (NFTs) to enable support for royalty payments across all NFT marketplaces in the NEO Smart Economy.
+
+==Motivation==
+
+Need support royalty information which is not only for royalty payments across NFT marketplaces, but also good to protect creative copyright. Currently, no standard for creators to declare royalty to all marketplaces in their NFTs. They have to discuss the royalty with different marketplaces, otherwise these marketplaces which haven't touched with the creators can't get royalty details and have to define the royalty in their marketplace contract by themselves. This standard is compatible with [https://github.com/neo-project/proposals/blob/master/nep-11.mediawiki NEP-11].
```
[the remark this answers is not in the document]
vncoelho: If you want to use this alternative version of the motivation

nep-x1.mediawiki
```diff
@@ -0,0 +1,161 @@
...
+  Title: NFT Royalty Standard
+  Author: Owen Zhang<zhangzhihong@ngd.neo.org>, Vincent Geneste<vincentgeneste@gmail.com>
+  Type: Standard
+  Status: Draft
+  Created: 2022-09-05
+</pre>
+
+==Abstract==
+
+This NEP defines a global standard to get royalty payment information for non-fungible tokens (NFTs) to enable support for royalty payments across all NFT marketplaces in the NEO Smart Economy.
```
vncoelho, line 12: This NEP defines a global standard to get royalty payment information for Non-Fungible Tokens (NFTs) in order to enable support for royalty payments across all NFT marketplaces in the NEO Smart Economy.

vncoelho: minor changes here as well, fell free to apply or not

nep-x1.mediawiki
vincentgeneste: It’s one event emitted per recipient

vncoelho: So fine, thanks

vncoelho: Let's add this in the description

nep-x1.mediawiki
```diff
@@ -0,0 +1,161 @@
...
+
+Marketplaces that support this standard MUST emit the event, <code>RoyaltiesTransferred</code>, after sending a payment.
+
+===Events===
+
+====RoyaltiesTransferred====
+
+<pre>
+{
+  "name": "RoyaltiesTransferred",
```
vncoelho, line 129: RoyaltiesTransferred_i

vncoelho: or RoyaltiesTransferredi
RoyaltiesTransferred1
RoyaltiesTransferred2

nep-x1.mediawiki
```diff
@@ -0,0 +1,161 @@
...
+
+This function is not for initial sale or mint, it's only for secondary marketplaces.
+
+Marketplaces that support this method MAY implement any method of calculating or transferring royalties to the royalty recipient.
+
+Marketplaces MUST pay the royalty in the same unit of exchange as that of the <code>salePrice</code> passed to royaltyInfo(). For example, if the sale price is in NEO, then the royalty payment must also be paid in NEO, and if the sale price is in GAS, then the royalty payment must also be paid in GAS.
+
+The royaltyInfo() function MUST NOT ignore the <code>salePrice</code>, <code>royaltyAmount</code> MAY be dynamic due to <code>salePrice</code> and time. 
+
+Marketplaces that support this standard MUST emit the event, <code>RoyaltiesTransferred</code>, after sending a payment.
```
vncoelho, line 121: Marketplaces that support this standard MUST emit the event for each recipient, <code>RoyaltiesTransferred</code>, after sending a payment.

nep-x1.mediawiki
superboyiii: It's not good for adding a count on 1 event. It's just like `Transfer`, we can have multi `Transfer` in 1 transaction.

superboyiii:
@vncoelho Thanks for your advice, I made some improvement. Please check again.

nep-x1.mediawiki
vncoelho: sure

vncoelho:
> @vncoelho Thanks for your advice, I made some improvement. Please check again.

Looks good to me, fell free to move forward

vncoelho (approved):

vncoelho (approved):

vincentgeneste (approved):

shargon:
Do we change to accept or merge this as a draft and make a different pr for number assignment?

vncoelho:
> Do we change to accept or merge this as a draft and make a different pr for number assignment?

Any is good to me

Jim8y (approved):

roman-khimov:
> name it as NEP-23 

It's already in https://github.com/neo-project/proposals/pull/156, let's take 24 for it.

superboyiii:
I will update the Number soon!

nep-x1.mediawiki
```diff
@@ -0,0 +1,161 @@
...
+}
+</pre>
+
+<code>royaltyToken</code> MUST be the same address as that in <code>royaltyInfo</code> method.
+
+<code>royaltyRecipient</code> MUST be the same address as that in <code>royaltyInfo</code> method.
+
+MUST trigger after marketplaces transferring royalties to the royalty recipient if <code>royaltyInfo</code> method is implemented.
+
+Each <code>royaltyRecipient</code> has 1 events. That means if 5 <code>royaltyRecipient</code> in <code>royaltyInfo</code> method, 5 <code>RoyaltiesTransferred</code> events MUST be triggered.
```
AnnaShaleva, line 161: s/`has 1 events`/`has 1 event`

AnnaShaleva, line 161: s/`if 5 <code>royaltyRecipient</code>`/`if there are 5 <code>royaltyRecipient</code>`

nep-x1.mediawiki
```diff
@@ -0,0 +1,161 @@
+<pre>
+  NEP: Undefined
+  Title: NFT Royalty Standard
+  Author: Owen Zhang<zhangzhihong@ngd.neo.org>, Vincent Geneste<vincentgeneste@gmail.com>
+  Type: Standard
+  Status: Draft
+  Created: 2022-09-05
+</pre>
+
+==Abstract==
```
AnnaShaleva, line 10: We need to add Table of Contents before the Abstract to match the rest of NEPs.

nep-x1.mediawiki
```diff
@@ -0,0 +1,161 @@
...
+
+This NEP defines a global standard to get royalty payment information for Non-Fungible Tokens (NFTs) in order to enable support for royalty payments across all NFT marketplaces in the NEO Smart Economy.
```
AnnaShaleva, line 12: Need to add a link to https://github.com/neo-project/proposals/blob/master/nep-11.mediawiki.

nep-x1.mediawiki
```diff
@@ -0,0 +1,161 @@
...
+      "name": "royaltyToken",
+      "type": "Hash160"
+    },
+    }
+    {
+      "name": "salePrice",
+      "type": "Integer"
+    }
+  ],
+  "returntype": "Array"
```
AnnaShaleva, line 45: Should this method always return the same result given the same input data or it's allowed to change royalty amount or receipts through time? Maybe a note should be added. Ref. https://github.com/neo-project/proposals/blob/90686c5fae075541285f1200a584223504878b6d/nep-11.mediawiki?plain=1#L40

nep-x1.mediawiki
AnnaShaleva, line 161: Do we have any implementation example? Without implementation this standard can't be moved to the Final stage:

> Implementations -- The implementations must be completed before any NEP is given status "Final", but it need not be completed before the NEP is accepted. It is better to finish the specification and rationale first and reach consensus on it before writing code.

AnnaShaleva (commented):
The standard itself looks useful to me.

nep-x1.mediawiki
vincentgeneste: GhostMarket already supports it yes, both our nep11 and our trading contract implemented it. what’s needed to show the implementation? Just link to repo?

superboyiii: > Do we have any implementation example? Without implementation this standard can't be moved to the Final stage:
> 
> > Implementations -- The implementations must be completed before any NEP is given status "Final", but it need not be completed before the NEP is accepted. It is better to finish the specification and rationale first and reach consensus on it before writing code.

[MegaOasis.io](https://megaoasis.io/) has also implemented this as well.

nep-x1.mediawiki
superboyiii: `royaltyAmount` maybe dynamic. `royaltyRecipient` can be tranferred to another address.

superboyiii:
Updated @AnnaShaleva

shargon (dismissed):

Jim8y (approved):

Jim8y:
I think this one is good to go @shargon

nep-x1.mediawiki
shargon: > GhostMarket already supports it yes, both our nep11 and our trading contract implemented it. what’s needed to show the implementation? Just link to repo?

Yes, we could add the link to our contract

vincentgeneste: here's our NEP11 implementation https://github.com/OnBlockIO/n3-tokens-contracts/blob/6b01ac3f46b18629f880dcac4e2f5bb4e7f94a37/contracts/NEP11/GhostMarket.NFT.py#L475

Jim8y: @superboyiii

README.mediawiki
```diff
@@ -99,7 +99,18 @@ First review [[nep-1.mediawiki|NEP-1]]. Then clone the repository and add your N
 | Informational
 | Accepted
 |-
-| 
+| [https://github.com/neo-project/proposals/pull/156 23]
+| RPC Errors Proposal
+| Anna Shaleva, Roman Khimov
+| Standard
+| Accepted
```
AnnaShaleva, line 106: I don't think it should be here. It's not related to this PR. If #156 is merged, then we just rebase onto master.

README.mediawiki
```diff
@@ -140,4 +151,4 @@ First review [[nep-1.mediawiki|NEP-1]]. Then clone the repository and add your N
 | 
 | Standard
 | Stub
-|}
+|}
```
AnnaShaleva, line 154: Useless change.

nep-24.mediawiki
```diff
@@ -0,0 +1,161 @@
...
+}
+</pre>
+
+<code>royaltyToken</code> MUST be the same address as that in <code>royaltyInfo</code> method.
+
+<code>royaltyRecipient</code> MUST be the same address as that in <code>royaltyInfo</code> method.
+
+MUST trigger after marketplaces transferring royalties to the royalty recipient if <code>royaltyInfo</code> method is implemented.
+
+Each <code>royaltyRecipient</code> has 1 event. That means if there are 5 <code>royaltyRecipient</code> in <code>royaltyInfo</code> method, 5 <code>RoyaltiesTransferred</code> events MUST be triggered.
```
AnnaShaleva, line 161: Then we need to add links to the Royalty standard implementation to the end of this NEP, exactly like in https://github.com/neo-project/proposals/blob/master/nep-11.mediawiki#implementation.

nep-24.mediawiki
```diff
@@ -0,0 +1,161 @@
+<pre>
+  NEP: 24
+  Title: NFT Royalty Standard
+  Author: Owen Zhang<zhangzhihong@ngd.neo.org>, Vincent Geneste<vincentgeneste@gmail.com>
+  Type: Standard
+  Status: Accepted
+  Created: 2022-09-05
+</pre>
+
```
AnnaShaleva, line 9: Table of contents is still not added. We need it exactly like in all NEPs, see https://github.com/neo-project/proposals/blob/master/nep-11.mediawiki#table-of-contents.

Jim8y: @superboyiii can make the final format update

superboyiii: > Table of contents is still not added. We need it exactly like in all NEPs, see https://github.com/neo-project/proposals/blob/master/nep-11.mediawiki#table-of-contents.

It's automatically generated. See: https://github.com/neo-project/proposals/blob/8a8b8c229262d46e553b00f9110eae3ccad6c1a3/nep-24.mediawiki
c9ade7d951b311b9529c0c2b6a3c089

superboyiii: Will add another link to C# demo, please don't merge.

nep-24.mediawiki
superboyiii: Will add another link to C# demo, please don't merge.

shargon: When i'ts ready, please change the label

shargon (dismissed):

nep-24.mediawiki
AnnaShaleva: Got it, my bad.

AnnaShaleva (commented):
LGTM, but if we're going to merge it, then I think we need to start with those NEPs that were approved earlier (#156) to avoid conflicts in README.

nep-24.mediawiki
AnnaShaleva: Resolved.

nep-x1.mediawiki
AnnaShaleva: Resolved.

superboyiii:
I think it's not necessary to add count after `royaltyRecipient` and `royaltyAmount` in Multi royaltyRecipient

shargon (approved):

superboyiii:
@shargon @AnnaShaleva Good to merge!

shargon (approved):

AnnaShaleva (approved):

Jim8y (approved):

PR #151, opened 2022-06-26
Debug Info v2

devhawk:
This is a draft PR for the updated Debug Info used by the [Neo Debugger Storage Schema Preview](https://github.com/neo-project/neo-debugger/blob/master/docs/storage-schema-overview.md).

Changes are not expected, but until the storage schema work is out of preview, this PR will stay in draft status.

PR #160, opened 2022-12-09
Draft NEP 25: Extended types for NEP-14

roman-khimov:
This is a proposal for a new ABI standard replacing NEP-14. It solves the problem of opaque Array, Map and InteropInterface types allowing to provide more specific details about them. A somewhat similar thing is also provided in #151, but it's detached from the ABI and introduces some new notations that should be parsed by clients. This NEP is based on NeoGo [ExtendedType data](https://pkg.go.dev/github.com/nspcc-dev/neo-go/pkg/smartcontract/binding#ExtendedType) that is used for contract-specific RPC SDK generation, but that implementation is also detached from the ABI (an additional YAML file is used).

While this problem can be solved with additional (non-ABI) data, it might add some additional barriers affecting contract interoperability (starting with the basics, where to store this data). The intention here was to create a 100% backwards-compatible extension (if new fields are ignored) that can be easily provided by compilers and can be easily consumed by any other tools, irrespective of the implementation language. Having this data in ABI means every contract will have it and it'll be easily accessible. In fact this data is a part of the ABI in its essence, that's what contracts accept/return.

Things intentionally omitted for now (I'd like to get more feedback on this before doing any of those):
 * JSON schemas would be more appropriate for specifications
 * "length" parameter for integers (8/16/32/64/128), byte arrays and strings (1-1M) can be useful for more precision if it's known exactly
 * "unsigned" for integers
 * maybe even "min" and "max" for integers
 * "nullable" to resolve "Null" handling ambiguity, can it be provided in a parameter, can it be returned from the method?
 * the problem of `Any` parameter in the specification (like `data` for `transfer` in NEP-17 or NEP-11) with the contract using some specific type (if it uses `data`, it expects some particular type of `data`)


ixje:
> * "nullable" to resolve "Null" handling ambiguity, can it be provided in a parameter, can it be returned from the method?

:1st_place_medal:

devhawk (changes requested):
I am very much in favor of this proposal, but it needs to be combined with the changes proposed in #151. In particular, I think this proposal needs a syntax-independent description of the extended type semantics. Something similar to the [ContractType Type Model](https://github.com/neo-project/proposals/blob/7ef78e595c7a67db71744acbd10f73803b09ae6d/nep-19.mediawiki#contracttype-type-model-version-2) section of #151.

Note, I don't care if the eventual proposal uses the string encoding from #151 (`Array<Hash160>`) or JSON encoding from this proposal (`{"type": "Array", "value": {"type": "Hash160"}}`). Whatever the group at large decides to support for *this* proposal, debug info v2 will adopt.

roman-khimov:
> Something similar to the [ContractType Type Model](https://github.com/neo-project/proposals/blob/7ef78e595c7a67db71744acbd10f73803b09ae6d/nep-19.mediawiki#contracttype-type-model-version-2) section of https://github.com/neo-project/proposals/pull/151.

I only see `ContractType` extensions there like `Struct` or `Address`, but these can't be backwards compatible with NEP-14 and I'd like to keep that given that the additional type data here allows to differentiate between structures and arrays. `Address` is `Hash160` to me, any `Hash160` can be an address, it's just a matter of representation. `Iterator` is also included here as one (and only, technically the only other we have is storage context, but of course this can be changed in future) of `InteropInterface` subtypes. Named structured types are also possible here. Am I missing something?

devhawk:
A few nitpicks:

* I think we should use the term `struct` instead of `namedtype`
* Why is `value` a JSON object with a `type` property? Seems needlessly complex. Can we change this to a string value?

devhawk:
> I only see `ContractType` extensions there like `Struct` or `Address`, but these can't be backwards compatible with NEP-14 and I'd like to keep that given that the additional type data here allows to differentiate between structures and arrays. `Address` is `Hash160` to me, any `Hash160` can be an address, it's just a matter of representation. `Iterator` is also included here as one (and only, technically the only other we have is storage context, but of course this can be changed in future) of `InteropInterface` subtypes. Named structured types are also possible here. Am I missing something?

It is true that the way #151 *encodes* Struct types would be a NEP-14 breaking change. But the semantics of #151 structs is not a breaking change. The semantics of encoding a struct param as `{type: package.Structure}` vs `{type: Array, namedtype: package.Structure}` are equivalent. The only difference is back compatibility. Arguably, the biggest issue with #151 is the breaking change to how type info is encoded. Further, I agree that the syntax approach defined in this PR preferable to #151 since it adds the additional info without breaking back compat. This has huge benefits for all language projects. 

I asked for this PR to include a syntax-independent description of the extended type system to ensure that the extended type system works for both contract client code generation *and* better debugger experience. Once we nail down the semantics of the extended type system, we can choose a syntax that is backwards compatible. 

Your point about Iterator is spot on. #151 defines semantics for Interop ContractType, but didn't originally include any specific semantics for Iterators. Given their importance, this is an oversight in #151 that I recently addressed but haven't implemented yet. Having a syntax-independent description of the type system allows us to have discussions like "What information do we need about Interop parameters for code gen + debugger purposes?" without getting bogged down on how this stuff is encoded. 

BTW, Address is just an alias for Hash160 so it wouldn't break NEP-14 back compat. As you point out, "any Hash160 can be an address, it's just a matter of representation". Hash160 representation may not affect contract client code gen but it *definitely* affects debugger representation. The same way this proposal adds extended fields to more fully describe `Array` and `Map`, it could also add extended field to better describe `Hash160` in a back compatible way such as `{type: Hash160, address: true}` or something like that.

roman-khimov:
> I think we should use the term struct instead of namedtype

Technically, we can have non-struct things in `namedtypes`, maybe they could be used for something in future. But we have only structures there now, that's true, can be renamed.

> Why is value a JSON object with a type property? Seems needlessly complex. Can we change this to a string value?

A string will need to be parsed, while we can have the same JSON structures representing all of this directly (maybe some depth limit is necessary though, but I've tested things like `map[int][]map[string][]interop.Hash160` (map of int to array of maps of string to array of hash160) with NeoGo `ExtendedType`). To me unified representation is easier to work with. #151 is more compact here, I agree, but it requires additional handling.

roman-khimov:
> Once we nail down the semantics of the extended type system, we can choose a syntax that is backwards compatible.

True, the main thing of course is to settle on the exact details we want/need to express.

> `{type: Hash160, address: true}` or something like that.

Do we have any `Hash160` that is not an address? But it can be added if this helps in any way.

devhawk:
> To me unified representation is easier to work with. #151 is more compact here, I agree, but it requires additional handling.

When we first built the debugger, I was concerned about debug info size. Hence we chose a compact CSV-style string representation instead of separate JSON fields. But that concern has proven to be somewhat overblown.

> True, the main thing of course is to settle on the exact details we want/need to express.

If you add the semantic type system details to this document, then I will be able to leverage that for the advanced debug information. As I said, I'd be happy to update the debug info proposal to support JSON serialization for type information in order to bring it more inline with how the manifest is serialized. That should make producing debug info easier for language teams.

> Do we have any `Hash160` that is not an address? But it can be added if this helps in any way.

Contract Hashes are Hash160s values but are typically *not* treated as an address.

devhawk:
@roman-khimov Any progress on including the type system info from #151 into this NEP?

roman-khimov:
I'm waiting for more feedback on "things intentionally omitted for now" and other general topics, "Address" (as a representation hint for Hash160) can be added as well if it's OK for us to have representation details at this level (it's OK for me, but maybe there are some objections).

BTW, as I've said earlier, we have JSON-RPC SDK generator that uses similar data to create (Go) client-side code that will then work via regular Neo JSON-RPC. It works fine for backend code, but we also have something for frontend in the queue. It's a different experimental thing that automatically provides server-side contract-specific RESTful APIs using similar ABI data, this then makes it trivial for JS clients to access contract methods.

devhawk:
> I'm waiting for more feedback on "things intentionally omitted for now" and other general topics, "Address" (as a representation hint for Hash160) can be added as well if it's OK for us to have representation details at this level (it's OK for me, but maybe there are some objections).

Looking forward to the next revision after you get more feedback.

> BTW, as I've said earlier, we have JSON-RPC SDK generator that uses similar data to create (Go) client-side code that will then work via regular Neo JSON-RPC. It works fine for backend code, but we also have something for frontend in the queue. It's a different experimental thing that automatically provides server-side contract-specific RESTful APIs using similar ABI data, this then makes it trivial for JS clients to access contract methods.

There's a similar tool in Neo.BuildTools that generates a C# interface from the current manifest file. THis interface is currently only used in C# test methods, but could also be used for generating client APIs as well. I've had similar thoughts to generating these APIs from the extended type info in the prototype v2 debug info. However, I would *MUCH* rather this extended type info in the manifest.

igormcoelho:
Interesting proposal to have all types checked. Regarding syntax, do you think some shorter "template-based" version like:

Original:
```
         "returntype" : "InteropInterface",
         ...
         "extendedreturntype" : {
            "value" : {
               "type" : "Array",
               "namedtype" : "package.Structure"
            },
            "type" : "InteropInterface",
            "interface" : "IIterator"
         },
```

Possibility with template-based syntax embedded directly on the string:
```
         "returntype" : "InteropInterface",
         ...
         "extendedreturntype" : "InteropInterface<IIterator<Array<package.Structure>>>" 
```

Is this the intention, right? It's smaller, but do you think it makes it much harder to parse? (some extra code will be needed for angle brackets)

roman-khimov:
It's more like `InteropInterface<IIterator<package.Structure>>` (in that we have an iterator and each value it returns is a `package.Structure`). But I'd keep it as is because:
 * parsing complicates things and goes a bit against the spirit of NEP-14 to me
 * we have some symmetry between `extendedreturntype` and `Parameter` type data
 * any NEP-14 type definition is also a valid extended type definition, a property that is nice to have

devhawk:
Neo Debug Info v2 uses a single string like @igormcoelho asked about. But I agree with @roman-khimov that a mechanism that is easier to parse is probably easier to get adopted across the community

melanke:
This is a very interesting proposal, in fact, before knowing about this proposal, @meevee98, @luc10921 and I were working on something similar to this. We implemented a few "hints" generated by **neo3-boa** ([PR link](https://github.com/CityOfZion/neo3-boa/pull/991)) that can be used by **neo3-parser** ([PR link](https://github.com/CityOfZion/neo3-parser/pull/1/files?short_path=774cc6e#diff-774cc6e8d51969a78ed81b2861ea51b7e2e026d6165d9b43f61f24dbcbb15d88)).

For instance, if a method is implemented like this:
```py
@public
def Main(var: Dict[str, List[bool]]) -> List[Union[Dict[str, int], str, bool]]:
    if var:
        return []
    return []
```

Neo3-boa will compile the manifest to something like this:
```json
{
    "name": "ManifestTypeHintMapsArraysUnionHint",
    "groups": [],
    "abi": {
        "methods": [
            {
                "name": "Main",
                "offset": 0,
                "parameters": [
                    {
                        "type": "Map",
                        "generickey": {
                            "type": "String"
                        },
                        "genericitem": {
                            "type": "Array",
                            "generic": {
                                "type": "Boolean"
                            }
                        },
                        "name": "var"
                    }
                ],
                "safe": false,
                "returntype": "Array",
                "returngeneric": {
                    "type": "Any",
                    "union": [
                        {
                            "type": "String"
                        },
                        {
                            "type": "Map",
                            "generickey": {
                                "type": "String"
                            },
                            "genericitem": {
                                "type": "Integer"
                            }
                        },
                        {
                            "type": "Boolean"
                        }
                    ]
                }
            }
        ],
        "events": []
    },
    "permissions": [
        {
            "contract": "*",
            "methods": "*"
        }
    ],
    "trusts": [],
    "features": {},
    "supportedstandards": [],
    "extra": null
}
```

There are only 5 new keys to indicate the extended types:
hint, generic, genericKey, genericItem and union.

Today this is saved on the manifest but when deploying to the blockchain this is lost. We have to manually copy this content from the manifest to neo3-parser as a `ParseConfig`.

So I am very excited by this proposal, even if it's decided a very different format than what we implemented.
It would makes developer's life way easier :)

devhawk:
I'm back from vacation. Can we start making progress on this?

roman-khimov:
As for "things intentionally ommitted for now", I don't see any specific opinions on that which to me means that there is no immediate need for any of them. They can be added in a future NEP update, so it shouldn't be a blocker.

The only relevant unsolved feedback here is the address representation from @devhawk. I don't see any objection to it, so I'll update the proposal to include this data as well and consider it to be done for now.

nep-Y.mediawiki
```diff
@@ -0,0 +1,328 @@
...
+
+An Application Binary Interface (ABI) is the interface between two program modules, one of which is often a library and/or operating system and the other one is usually an application created by a regular programmer.
+
+This NEP describes the ABI standards for Neo smart contracts.
+
+==Motivation==
+
+Neo smart contract system is designed to be mutually invocable between contracts. To achieve this, we need a mechanism for exposing the interface of smart contracts. With NeoContract ABI, developers can easily create programs to invoke smart contracts or write clients that automatically access contract functionalities.
+
+This NEP is a compatible extension of the previous one, NEP-14. Some complex
```
vncoelho, line 21: add link to https://github.com/neo-project/proposals/blob/master/nep-14.mediawiki?

vncoelho:
I will soon review and vote.

nep-Y.mediawiki
```diff
@@ -0,0 +1,331 @@
+<pre>
+  NEP: <to be assigned>
```
AnnaShaleva, line 2: NEP: 24? Right after https://github.com/neo-project/proposals/pull/156.

Also need to update README.

AnnaShaleva (commented):
Definitely should be accepted. We've tried extended types with our compiler, these types help a lot in RPC bindings code generation, they are as informative as needed for fully automated structures/iterators/nested types code generation.

nep-Y.mediawiki
superboyiii: > NEP: 24? Right after #156.
> 
> Also need to update README.

Maybe 25, https://github.com/neo-project/proposals/pull/155 is ready for 24

roman-khimov: Done.

README.mediawiki
```diff
@@ -45,12 +45,6 @@ First review [[nep-1.mediawiki|NEP-1]]. Then clone the repository and add your N
 | Standard
 | Final
 |-
-| [[nep-14.mediawiki|14]]
-| NeoContract ABI
-| Erik Zhang
-| Standard
-| Final
```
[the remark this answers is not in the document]
roman-khimov: The standard itself is moved to `obsolete`, but it's removed from the list just like old NEP-3.

shargon (commented):
@roman-khimov Why not set `"useExtendedTypes": true` in the abi, and use the same fields for types?

roman-khimov:
Backwards compatibility. Valid NEP-14 manifest remains valid NEP-25 manifest which simplifies things a lot to me, you don't have to treat these manifest types separately.

roman-khimov:
BTW, the other way around is also true. Valid NEP-25 is a valid NEP-14 if you're to ignore new fields. So all the tooling we have this NEP-14 won't break immediately.

shargon (commented):
Can we add more types, like byte, short, int, long?

roman-khimov:
That'd be 

> "length" parameter for integers (8/16/32/64/128), byte arrays and strings (1-1M) can be useful for more precision if it's known exactly

from https://github.com/neo-project/proposals/pull/160#issue-1486551106. In general, I like it because it allows to generate more specific code from the manifest, but it complicates things at the same time (especially given the fact that VM doesn't distinguish them). Vote below, :+1: :-1:.

AnnaShaleva:
>but it complicates things at the same time (especially given the fact that VM doesn't distinguish them)

Vote up if we make this `length` parameter optional. Shall we? I don't have anything against this `length` parameter, it may be helpful for some manifest-level checks and is definitely useful for contract bindings, but I'm a little bit worried about the manifest generation part, it may be tricky to generate proper manifest with this additional info enabled.

Jim8y (dismissed):

shargon (commented):
Can we change the parameter to be:

```
{
  "name": "from",
  "type": "Hash160"
   "extendedtype": { extended type object } 
}
```

?

I think that is more backward compatible, and easy to deal with the current code, take a look to the implementation (https://github.com/neo-project/neo/pull/3272/files) If we use a parameter other than the return, we cannot have a clean hierarchy.

roman-khimov:
I _really_ wanted to avoid additional nesting and `extendedtype` fields, so it's embedded (except for `returntype`!). There are some constraints on how different fields can be used depending on the base `type` and having it at the same level simplifies these checks, but then if we have it both in `parameter` and `extendedtype` it's duplication, a bit less effective (and needs to be checked as well).

But then again, we inevitably have this problem already with `returntype`, so your proposal adds some symmetry here. Let's see if there are any opinions on this, it can be changed if it's eaiser to implement in the end.

AnnaShaleva:
I don't have any special preferences on this topic, Fernando's option is a bit easier to implement and follow, Roman's one is more compact. Maybe if we have this problem with return type, then it's better to have a symmetric structure for simple parameters.

nep-25.mediawiki
```diff
@@ -0,0 +1,331 @@
...
+data types required more data for ABI to be useful for automatic SDK/wrapper
+code generation. For example, parameter of "Array" type could be a structure
+with three fields of "ByteArray", "Hash160" and "Integer" types or an array of
+"Integer" values. These cases couldn't be differentiated both for method
+parameters and return values with
+[https://github.com/neo-project/proposals/blob/06a6b550987a74025009cf6841757f2426462697/nep-14.mediawiki NEP-14], while this proposal fixes it.
+
+==Rationale==
+
+We assume the Application Binary Interface (ABI) is strongly typed, known at compilation time and static. No introspection mechanism will be provided. We assert that all contracts will have the interface definitions of any contracts they call available at compile-time.
```
Jim8y, line 32:
```suggestion
We assume the ABI is strongly typed, known at compilation time and static. No introspection mechanism will be provided. We assert that all contracts will have the interface definitions of any contracts they call available at compile-time.
```

Jim8y (dismissed):

AnnaShaleva:
@Jim8y, we can't approve it yet, we firstly need to decide on https://github.com/neo-project/proposals/pull/160#pullrequestreview-2074421878.

@neo-project/core, need your opinions on this topic.

roman-khimov:
@AnnaShaleva, I will update it and address @shargon's feedback. If it's easier to implement and we don't have any strong arguments against it (other than duplication, but you we already have some of it) then this is the way to go.

Jim8y:
> @Jim8y, we can't approve it yet, we firstly need to decide on [#160 (review)](https://github.com/neo-project/proposals/pull/160#pullrequestreview-2074421878).

I dont like 2-year late suggestion. And but since @roman-khimov decides to update, so be it.

cschuchardt88 (dismissed):

roman-khimov:
@shargon, take a look at the latest commit, especially at the examples. I think it sucks for `namedtypes`, much more complex structure, it'd be harder to use. `extendedreturntype` on the other hand is not used often, most of the time functions return simple bool/hash things, more rarely iterators.

shargon (dismissed):
Now I think that it's easier to implement

Jim8y (dismissed):

AnnaShaleva (dismissed):

igormcoelho:
For me, the standard is clear, just a few comments.
The first example given in standard for extended types is very weird:
```
{
  "type": "Integer",
  "namedtype": "name",
  "interface": "IIterator",
  "key": "Integer",
  "value": {},
  "fields": []
}
```
I took some time to realize that this object is impossible... as how could an Integer have an extended type? Could a better example be given here please?
And for the last json: 
```
{
  "methods": [
    {
      "returntype": "InteropInterface",
     ...
```
I put it on a JSON checker and beautifier, and it is broken. I think it's just an extra comma, but anyway, please ensure that all json given in standard are properly checked and beautified.
My beautifier did not put the fields in a particular order, but it would be easier to understand if the fields are ordered... not in alphabetic order, but in some logic order at least.
For example, this is confusing:
```
        {
          "type": "String",
          "name": "StringField"
        },
        {
          "name": "ByteArrayField",
          "type": "ByteArray"
        }
```
It seems to me that the `type` field should always come first... it's easier to understand what object is that. If you agree, I can change it.

roman-khimov:
> I took some time to realize that this object is impossible... as how could an Integer have an extended type? Could a better example be given here please?

As written there:
> (the object below is not a valid ExtendedType specification, rather it represents all possible fields).

_Any_ example that has all fields would be incorrect in some ways, so I doubt it can be improved.

> I put it on a JSON checker and beautifier, and it is broken.

Fixed!

> My beautifier did not put the fields in a particular order, but it would be easier to understand if the fields are ordered...

Well, JSON is order-agnostic for object elements, so we can have it in any way, but to make it easier to read I've reordered some things.

cschuchardt88:
>name - is the name of the method, which can be any valid identifier.

Can we address this as well as others in this `NEP`. The above is from `NEP-3` and never was addressed in `NEP-14` that is `any valid identifier`? See https://github.com/neo-project/neo/issues/2930 for more details. Now properties need to have restrictions of what they `SHOULD BE` or `MUST BE`. We need to be more `strict` in `NEP`s. They need to be written like `RFC` `Standards`, leave no wiggle room for error or guessing. Otherwise it just a guideline.

roman-khimov:
Likely what was originally meant there is https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/identifier-names. But I don't know if there are any contracts now that go beyond this definition of identifier (which would raise compatibility question). We can fix it (to valid UTF-8 letters?), but only after mainnet/testnet check.

`safe` definition can also be fixed, it's very vague at the moment.

shargon (dismissed):

AnnaShaleva (dismissed):

cschuchardt88 (commented):
All the `json` should be `json syntax highlighted`, no sense in using legacy markdown commands.

\`\`\`json
Text here
\`\`\`

nep-25.mediawiki
```diff
@@ -0,0 +1,362 @@
...
+
+Extended type data is an object can be used for <code>extendedtype</code>
+and <code>extendedreturntype</code> fields or named types
+(the object below is not a valid ExtendedType specification, rather it
+represents all possible fields).
+
+<pre>
+{
+  "type": "Integer",
+  "namedtype": "name",
```
shargon, line 174: Add `AllowNull: boolean` ?

roman-khimov: Made it the other way around with default to allow null and a field to forbid them. I think it's safer as a default given the behavior of NEP-14, but this can be discussed.

shargon (approved):

AnnaShaleva (approved):

Jim8y (approved):

Jim8y:
@shargon please merge if others still not review, this proposal has being extensively discussed and reviewed and updated and apporoved.

vncoelho (commented):
Good it moved forward

shargon:
@roman-khimov value  and fields what kind of object are? It's not very clear to me in the description

roman-khimov:
`value` is an object (in JSON terms), `fields` is an array. Like in example:
```
            {
               "extendedtype" : {
                  "value" : {
                     "type" : "Hash160"
                  },
                  "type" : "Array"
               },
               "name" : "arrayOfHash160",
               "type" : "Array"
            }
```
for `[]Hash160` or (shortened):
```
      "local.Structure" : {
         "fields" : [
            {
               "name" : "IntField",
               "type" : "Integer"
            },
            {
               "name" : "Hash256Field",
               "type" : "Hash256"
            },
         ]
         "type" : "Array"
      }
```
for `Structure{IntField int, Hash256Field Hash256}`.

shargon:
> `value` is an object (in JSON terms), `fields` is an array. Like in example:
> 
> ```
>             {
>                "extendedtype" : {
>                   "value" : {
>                      "type" : "Hash160"
>                   },
>                   "type" : "Array"
>                },
>                "name" : "arrayOfHash160",
>                "type" : "Array"
>             }
> ```
> 
> for `[]Hash160` or (shortened):
> 
> ```
>       "local.Structure" : {
>          "fields" : [
>             {
>                "name" : "IntField",
>                "type" : "Integer"
>             },
>             {
>                "name" : "Hash256Field",
>                "type" : "Hash256"
>             },
>          ]
>          "type" : "Array"
>       }
> ```
> 
> for `Structure{IntField int, Hash256Field Hash256}`.

Free object or the same as `extendedType`?

roman-khimov:
ExtendedType, we can clarify minor details in an update, if needed. It's not `Final` yet.

PR #167, opened 2024-03-02
onNEP17Payment Callback Method Standard for NEO Smart Contracts

Jim8y:
This pr focus on making the `onNEP17Payment` method an NEP standard, such that user can declare their contract as payable and we can verify this method in our tools.

This nep is backwards compatible, not a single smart contract or standards need to be updated.

### DIFF
--- a/nep-17-x.mediawiki
+++ b/nep-17-x.mediawiki
@@ -1,0 +1,57 @@
+  NEP: <to be assigned>
+  Title: onNEP17Payment Callback Method Standard for NEO Smart Contracts
+  Author: Jimmy Liao <jimmy@r3e.network>
+  Type: Standard
+  Status: Draft
+  Created: <March 2, 2024>
+  
+==Abstract==
+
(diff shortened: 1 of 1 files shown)


roman-khimov (commented):
Parameters are not described. Implementation details (caller check, error handling) are not clear.

Seems like you're rushing for something.

Jim8y:
@roman-khimov told ya I can't write formal document, need you to do it. Open this pr to get a place for further discussion. Will close this one when you decide to write a formal one.

Jim8y:
BTW, I am not rushing, I just don't know how. LOL.

roman-khimov:
> Will close this one when you decide to write a formal one.

You've opened a PR. To me this means "I'm working on it with the goal to have it accepted eventually" (which includes fixing any problems and making it a real good standard). If that's the case, I'm more than happy to not do this because I have a pretty long TODO list. If that's not the case then I don't really understand why we have this PR. Issues are better suited for discussions.

Jim8y:
> > Will close this one when you decide to write a formal one.
> 
> You've opened a PR. To me this means "I'm working on it with the goal to have it accepted eventually" (which includes fixing any problems and making it a real good standard). If that's the case, I'm more than happy to not do this because I have a pretty long TODO list (I will review/comment on it of course to help you). If that's not the case then I don't really understand why we have this PR. Issues are better suited for discussions.

Sure, I will open an issue and close this pr. Thank you very much for your help.

PR #172, opened 2024-05-21
Draft NEP29: Standardization of NEP Number Assignment and Usage

Jim8y:
The current process for managing NEP numbers is cumbersome and inefficient, causing confusion and delays in the drafting and reviewing phases. Proposers often need to reference NEP numbers in their drafts and related documentation, but without a permanent number, this becomes a moving target until adoption. This NEP seeks to streamline the process by ensuring each draft receives a unique and final NEP number upon submission.

1. **NEP Number Assignment**:
   - Upon submission of a draft NEP, a unique NEP number will be assigned immediately.
   - This number will be considered used, regardless of the draft’s final status (adopted or not).

2. **Usage of NEP Numbers**:
   - Drafts must use their assigned NEP number in all references within the document.
   - When submitting a draft as a pull request (PR), the PR title must include the NEP number in the format: `Draft NEP XX: [Title]`.
   - If a PR includes multiple drafts, the title format should be: `Draft NEP XX and Draft NEP YY: [Title]`.

### DIFF
--- a/nep-29.mediawiki
+++ b/nep-29.mediawiki
@@ -1,0 +1,41 @@
+NEP: 29
+Title: Standardization of NEP Number Assignment and Usage
+Author: Jimmy Liao <jimmy@r3e.network>
+Type: Meta
+Status: Draft
+Created: 2024-05-22
+
+==Abstract==
+
+NEP 29 proposes a standardized method for assigning and managing Neo Enhancement Proposal (NEP) numbers. This proposal addresses the logistical challenges involved with the naming and referencing of NEP drafts, which currently cannot be assigned a permanent NEP number until the proposal is officially adopted. The solution is to uniquely assign NEP numbers at the draft stage, which will remain permanently associated with the draft, whether or not it is adopted.
+
+==Motivation==
+
+The current process for managing NEP numbers is cumbersome and inefficient, causing confusion and delays in the drafting and reviewing phases. Proposers often need to reference NEP numbers in their drafts and related documentation, but without a permanent number, this becomes a moving target until adoption. This NEP seeks to streamline the process by ensuring each draft receives a unique and final NEP number upon submission.
+
+==Specification==
+
+1. **NEP Number Assignment**:
+   - Upon submission of a draft NEP, a unique NEP number will be assigned immediately.
+   - This number will be considered used, regardless of the draft’s final status (adopted or not).
+
+2. **Usage of NEP Numbers**:
+   - Drafts must use their assigned NEP number in all references within the document.
+   - When submitting a draft as a pull request (PR), the PR title must include the NEP number in the format: `Draft NEP XX: [Title]`.
+   - If a PR includes multiple drafts, the title format should be: `Draft NEP XX and Draft NEP YY: [Title]`.
+
+==Rationale==
+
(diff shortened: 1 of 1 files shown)


roman-khimov:
1. This doesn't require an additional NEP, NEP-1 is exactly for this purpose.
2. Drafts can be NEPR-$PR_NUMBER, like NEPR-172 here. Much easier to separate from real NEPs, separate numbering.

Jim8y:
> 1. This doesn't require an additional NEP, NEP-1 is exactly for this purpose.
> 2. Drafts can be NEPR-$PR_NUMBER, like NEPR-172 here. Much easier to separate from real NEPs, separate numbering.

1. NEP-1 has no definition and regulation on managing the NEP number, otherwise there should not be a problem at all.
2. What is the benifit of making it slightly more complex as using PR number? Cause you will not know your PR number until you open one. 
3. you still need to be assigned a NEP number that no one knows if that is already assigned but not yet merged. 
4. You still need to keep updating the file names and contents.

So, why?

superboyiii:
At least I agree with @Jim8y at 2 points:
1. Don't use NEPR number as NEP number, it breaks the way previous and make people confused when NEP-29 jump to NEP-123.
2. NEP number needn't pick serial numbers, so every PR +1 makes sense.

Jim8y:
and, using pr number means we have to update the pr again after we get sufficient approve, but that will dismiss all existing approve......while collecting approve can take multiple weeks or months.

shargon:
> and, using pr number means we have to update the pr again after we get sufficient approve, but that will dismiss all existing approve......while collecting approve can take multiple weeks or months.

That's right

nep-29.mediawiki
```diff
@@ -0,0 +1,41 @@
+NEP: 29
+Title: Standardization of NEP Number Assignment and Usage
+Author: Jimmy Liao <jimmy@r3e.network>
+Type: Meta
+Status: Draft
+Created: 2024-05-22
+
+==Abstract==
+
+NEP 29 proposes a standardized method for assigning and managing Neo Enhancement Proposal (NEP) numbers. This proposal addresses the logistical challenges involved with the naming and referencing of NEP drafts, which currently cannot be assigned a permanent NEP number until the proposal is officially adopted. The solution is to uniquely assign NEP numbers at the draft stage, which will remain permanently associated with the draft, whether or not it is adopted.
```
AnnaShaleva, line 10: But what happens if some NEP is declined? Right now we have consecutive numbers of the accepted NEPs, and it would be nice to preserve it.

Jim8y: we have relatively `consecutive` numbers, some of them are already deprecated, no issue at all. We just list those active NEPs. Declined draft can drop that NEP number. No one will be interested why there is an NEP missing number, people will only focus on NEPs that are of their use.

roman-khimov (changes requested):
1. We have NEP-1 for these matters.
2. I don't see it solving any real problem for us.

Jim8y:
> 1. We have NEP-1 for these matters.
> 2. I don't see it solving any real problem for us.

@roman-khimov Its ok, you can ignor it. There have being many cases i have to convince you hard, cause you are always like believe there is no problem at all.

Even if NEPs being there for many years can not be finished, even if NEP numbers are assigned but we dont know, even if without this pr, new NEPs will have to keep updating the file name, the file content and dismiss all existing approvals, just because we dont know the NEP number or we have to keep updating the NEP number.

Its a mess of maintaining multiple NEP proposals already, can't you see it already?

PR #177, opened 2024-07-19
Update nep-6.mediawiki

### LINKED ISSUE #176: Broken link in NEP-6
The link to one of the implementation examples in the NEP-6 standard is broken. Likely due to some dir name changes in the `/neo` repo recently.

Current link:
https://github.com/neo-project/neo/blob/master/src/neo/Wallets/NEP6/NEP6Wallet.cs

Should be the following:
https://github.com/neo-project/neo/blob/master/src/Neo/Wallets/NEP6/NEP6Wallet.cs

shargon:
Close #176

### DIFF
--- a/nep-6.mediawiki
+++ b/nep-6.mediawiki
@@ -118,5 +118,5 @@ All old-format wallets should be able to be converted to this new JSON format ea
 
 ==Implementation==
 
-*neo-project/neo: https://github.com/neo-project/neo/blob/master/src/neo/Wallets/NEP6/NEP6Wallet.cs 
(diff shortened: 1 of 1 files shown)


Jim8y (approved):

cschuchardt88 (approved):

Jim8y:
Again, i propose to reduce the number of approvals. We can wait for longer for people to review even if we have collected sufficient review, but not necessary to require that many number of approvals.

People who dont review at all are blocking.

shargon:
I will merge it because it fix a broken link

PR #179, opened 2024-07-24
nep-Y: add _deploy NEP

roman-khimov:
NEP-22 can reference it. This mostly follows the implementation, there is nothing really new here.

### DIFF
--- a/nep-29.mediawiki
+++ b/nep-29.mediawiki
@@ -41,9 +41,9 @@ Neo N3 has a native <code>ContractManagement</code> contract that is used for co
 }
 </pre>
 
-This method will be automatically executed by ContractManagement contract when a contract is first deployed or updated. At this point new contract state is already saved (and it can be called from other code), but <code>Deploy</code> or <code>Update</code> notification is not yet emitted by ContractManagement contract.
+This method will be automatically executed by <code>ContractManagement</code> contract when a contract is first deployed or updated. At this point new contract state is already saved (and it can be called from other code), but <code>Deploy</code> or <code>Update</code> notification is not yet emitted by <code>ContractManagement</code> contract.
 
-The <code>data</code> parameter is completely contract-specific, contracts can accept any valid [[nep-14.mediawiki|NEP-14]] parameter types and handle them as required. It's the same <code>data</code> that is accepted by <code>deploy</code> and <code>update</code> methods of ContractManagement contract.
+The <code>data</code> parameter is completely contract-specific, contracts can accept any valid [[nep-14.mediawiki|NEP-14]] parameter types and handle them as required. It's the same <code>data</code> that is accepted by <code>deploy</code> and <code>update</code> methods of <code>ContractManagement</code> contract.
 
 <code>update</code> is true when contract is updated via <code>update</code> and it's false on initial deployment via <code>deploy</code>.
 
@@ -53,7 +53,7 @@ This callback is designed to guarantee the first code to be executed in a new/up
 
 <code>_deploy</code> parameters are made to be generic enough to cover all contract state management cases, it is possible to get additional per-network or per-instance contract data for initialization and updates can be easily differentiated from deployments via <code>update</code> parameter.
 
-Notice that if there are any notifications emitted during <code>_deploy</code> execution they will precede <code>Deploy</code> or <code>Update</code> notifications from ContractManagement. This behavior is historic, it mostly means that we're considering contract to be completely ready after succesful <code>_deploy</code> invocation (it can throw an exception or abort execution as well).
+Notice that if there are any notifications emitted during <code>_deploy</code> execution they will precede <code>Deploy</code> or <code>Update</code> notifications from <code>ContractManagement</code>. This behavior is historic, it mostly means that we're considering contract to be completely ready after succesful <code>_deploy</code> invocation (it can throw an exception or abort execution as well).
 
 ==Implementation==
 
--- a/nep-29.mediawiki
+++ b/nep-29.mediawiki
@@ -58,3 +58,4 @@ Notice that if there are any notifications emitted during <code>_deploy</code> e
 ==Implementation==
 
 * https://github.com/nspcc-dev/neofs-contract/blob/99fb86c35a48ed12017423aa4fee13f7d07fa4c0/contracts/netmap/contract.go#L79
+* https://github.com/neo-project/non-native-contracts/blob/8d72b92e5e5705d763232bcc24784ced0fb8fc87/src/NameService/NameService.cs#L439
--- a/nep-29.mediawiki
+++ b/nep-29.mediawiki
@@ -57,5 +57,5 @@ Notice that if there are any notifications emitted during <code>_deploy</code> e
 
 ==Implementation==
 
-* https://github.com/nspcc-dev/neofs-contract/blob/99fb86c35a48ed12017423aa4fee13f7d07fa4c0/contracts/netmap/contract.go#L79
-* https://github.com/neo-project/non-native-contracts/blob/8d72b92e5e5705d763232bcc24784ced0fb8fc87/src/NameService/NameService.cs#L439
+* C#: https://github.com/neo-project/non-native-contracts/blob/8d72b92e5e5705d763232bcc24784ced0fb8fc87/src/NameService/NameService.cs#L439
+* Go: https://github.com/nspcc-dev/neofs-contract/blob/99fb86c35a48ed12017423aa4fee13f7d07fa4c0/contracts/netmap/contract.go#L79
--- a/nep-29.mediawiki
+++ b/nep-29.mediawiki
@@ -17,7 +17,7 @@ Contracts may need to run some initialization/update code right after their depl
 
 ==Specification==
 
-Neo N3 has a native <code>ContractManagement</code> contract that is used for contract deployment and update via <code>deploy</code> and <code>update</code> methods of it. After settling with the new NEF and manifest both of them can invoke a special <code>_deploy</code> method if it's implemented by the contract.
+Neo N3 has a native <code>ContractManagement</code> contract that is used for contract deployment and update via <code>deploy</code> and <code>update</code> methods of it. After settling with the new NEF and manifest both of them will invoke a special <code>_deploy</code> method if it's implemented by the contract.
 
 ===Methods===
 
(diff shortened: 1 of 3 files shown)


roman-khimov:
Refs. #154.

Jim8y (approved):

Jim8y:
See, man, here is the problem, you dont have an NEP number,,,,, and we have to approve it without the number,,, then when its acctepted, you will need to update it with a number,,,, then all of us have to reapprove it again...... you know how hard it is to collect review and approves.

Jim8y (approved):

nep-29.mediawiki
```diff
@@ -0,0 +1,60 @@
...
+
+This proposal standardizes the callback function that can be implemented by contracts to have some code executed right after initial deployment or update.
+
+==Motivation==
+
+Contracts may need to run some initialization/update code right after their deployment or upgrade. Neo N3 has provided a mechanism for that since version 3.0, but it was never properly standardized, so this proposal fixes it.
+
+==Specification==
+
+Neo N3 has a native <code>ContractManagement</code> contract that is used for contract deployment and update via <code>deploy</code> and <code>update</code> methods of it. After settling with the new NEF and manifest both of them can invoke a special <code>_deploy</code> method if it's implemented by the contract.
```
shargon, line 20: We always said only N3, I think that we can remove Neo from the name

shargon (approved):

nep-29.mediawiki
roman-khimov: Well, I've seen "Neo N3" used a lot and maybe it's even more important now with "Neo X".

nep-Y.mediawiki
```diff
@@ -0,0 +1,60 @@
...
+{
+  "name": "_deploy",
+  "safe": false,
+  "parameters": [
+    {
+      "name": "data",
+      "type": "Any"
+    },
+    {
+      "name": "update",
```
AnnaShaleva: Just a minor note: NeoGo has a lot of examples with `isUpdate` parameter name instead of `update` (including the native Management itself). I know it's not important for the method itself, but what is the common standard for C# contracts?

AnnaShaleva (commented):
LGTM.

nep-29.mediawiki
```diff
@@ -0,0 +1,60 @@
...
+      "type": "Any"
+    },
+    {
+      "name": "update",
+      "type": "Boolean"
+    }
+  ],
+  "returntype": "Void"
+}
+</pre>
```
cschuchardt88, line 42:
```json
{
  "name": "_deploy",
  "safe": false,
  "parameters": [
    {
      "name": "data",
      "type": "Any"
    },
    {
      "name": "update",
      "type": "Boolean"
    }
  ],
  "returntype": "Void"
}
```

cschuchardt88: syntax highlighting would be nice.

roman-khimov: See https://github.com/neo-project/proposals/pull/180#discussion_r1704911920

nep-Y.mediawiki
roman-khimov: It was this way in #154 and likely that's because of https://github.com/neo-project/neo/blob/1488958d8b39057009f645a777ab4876029fecff/src/Neo/SmartContract/Native/ContractManagement.cs#L65, so `update` is a bit more canonical and we can just change to this name. Also, see
 * https://github.com/neo-project/non-native-contracts/blob/8d72b92e5e5705d763232bcc24784ced0fb8fc87/src/NameService/NameService.cs#L440
 * https://github.com/neo-project/neo-devpack-dotnet/blob/8f4cd2ba3de4e54946a563d4a1e5e937a31235d8/src/Neo.SmartContract.Template/templates/neocontractowner/Ownable.cs#L59
 * https://github.com/neo-project/neo-devpack-dotnet/blob/8f4cd2ba3de4e54946a563d4a1e5e937a31235d8/src/Neo.SmartContract.Template/templates/neocontractnep17/Nep17Contract.cs#L90

nep-29.mediawiki
vncoelho, line 20: Neo N3 has a native <code>ContractManagement</code> contract that is used for contract deployment (<code>deploy</code>) and update (<code>update</code>) methods. After settling with the new NEF and Manifest, both of them can invoke a special <code>_deploy</code> method (if it's implemented by the contract).

vncoelho: @roman-khimov, consider this new sentence.

One question.
"both of them can invoke a special¨, you mean, if the developer implements the _deploy method, it is still optional to call it from Deploy/Update,right?
It is not automatically and compulsory, correct?

roman-khimov: > consider this new sentence.

It was phrased like that in one of the earlier drafts, but I wanted to avoid "deployment (deploy)/update (update)". Pure stylistic, I like the way it's in the PR, but I can change it as well if there is a wider agreement that your version is better.

> if the developer implements the _deploy method

Then it will _always_ be called. Maybe `s/can/will/` makes sense here, a bit stronger causality (if implemented -> will invoke).

AnnaShaleva (dismissed):

nep-29.mediawiki
roman-khimov: Fixed. Now we need more approvals.

AnnaShaleva (dismissed):

Jim8y (approved):

shargon (dismissed):

igormcoelho:
My only recommendation here is to include an example of of the `data` on `_deploy`, something that could demonstrate its usefulness, and if possible, add a C# reference to the method implementation as well. Minor comments, author can decide to follow them or not. For me it's accepted.

roman-khimov:
> an example of of the data on _deploy

This better be in implementations, some of them do use it.

> C# reference to the method

Added! It's not used often, so it takes some time to find any implementation.

nep-29.mediawiki
```diff
@@ -0,0 +1,61 @@
...
+
+This method will be automatically executed by ContractManagement contract when a contract is first deployed or updated. At this point new contract state is already saved (and it can be called from other code), but <code>Deploy</code> or <code>Update</code> notification is not yet emitted by ContractManagement contract.
```
cschuchardt88, line 44:
```suggestion
This method will be automatically executed by <code>ContractManagement</code> contract when a contract is first deployed or updated. At this point new contract state is already saved (and it can be called from other code), but <code>Deploy</code> or <code>Update</code> notification is not yet emitted by <code>ContractManagement</code> contract.
```

wrap `ContractManagement`

nep-29.mediawiki
```diff
@@ -0,0 +1,61 @@
...
+
+==Rationale==
+
+This callback is designed to guarantee the first code to be executed in a new/updated contract. Previously contracts could provide methods of their own for doing this, but this either required an additional code in deployment/update transaction script or an additional transaction and some special contract state management (to disable actions before some initialization).
+
+<code>_deploy</code> parameters are made to be generic enough to cover all contract state management cases, it is possible to get additional per-network or per-instance contract data for initialization and updates can be easily differentiated from deployments via <code>update</code> parameter.
+
+Notice that if there are any notifications emitted during <code>_deploy</code> execution they will precede <code>Deploy</code> or <code>Update</code> notifications from ContractManagement. This behavior is historic, it mostly means that we're considering contract to be completely ready after succesful <code>_deploy</code> invocation (it can throw an exception or abort execution as well).
```
cschuchardt88, line 56:
```suggestion
Notice that if there are any notifications emitted during <code>_deploy</code> execution they will precede <code>Deploy</code> or <code>Update</code> notifications from <code>ContractManagement</code>. This behavior is historic, it mostly means that we're considering contract to be completely ready after succesful <code>_deploy</code> invocation (it can throw an exception or abort execution as well).
```

wrap `ContractManagement`

nep-29.mediawiki
```diff
@@ -0,0 +1,61 @@
...
+
+==Implementation==
+
+* https://github.com/nspcc-dev/neofs-contract/blob/99fb86c35a48ed12017423aa4fee13f7d07fa4c0/contracts/netmap/contract.go#L79
```
cschuchardt88, line 60:
```suggestion
* Google GO - https://github.com/nspcc-dev/neofs-contract/blob/99fb86c35a48ed12017423aa4fee13f7d07fa4c0/contracts/netmap/contract.go#L79
```

nep-29.mediawiki
```diff
@@ -0,0 +1,61 @@
...
+* https://github.com/neo-project/non-native-contracts/blob/8d72b92e5e5705d763232bcc24784ced0fb8fc87/src/NameService/NameService.cs#L439
```
cschuchardt88, line 61:
```suggestion
* C# .NET - https://github.com/neo-project/non-native-contracts/blob/8d72b92e5e5705d763232bcc24784ced0fb8fc87/src/NameService/NameService.cs#L439
```

nep-29.mediawiki
roman-khimov: Fixed.

nep-29.mediawiki
roman-khimov: Fixed.

cschuchardt88 (approved):

shargon (approved):

nep-29.mediawiki
```diff
@@ -0,0 +1,61 @@
...
+
+<code>_deploy</code> parameters are made to be generic enough to cover all contract state management cases, it is possible to get additional per-network or per-instance contract data for initialization and updates can be easily differentiated from deployments via <code>update</code> parameter.
+
+Notice that if there are any notifications emitted during <code>_deploy</code> execution they will precede <code>Deploy</code> or <code>Update</code> notifications from <code>ContractManagement</code>. This behavior is historic, it mostly means that we're considering contract to be completely ready after succesful <code>_deploy</code> invocation (it can throw an exception or abort execution as well).
+
+==Implementation==
+
+* C#: https://github.com/neo-project/non-native-contracts/blob/8d72b92e5e5705d763232bcc24784ced0fb8fc87/src/NameService/NameService.cs#L439
+* Go: https://github.com/nspcc-dev/neofs-contract/blob/99fb86c35a48ed12017423aa4fee13f7d07fa4c0/contracts/netmap/contract.go#L79
```
AnnaShaleva, line 61: Language keywords are missing in several other standards, may be then we need to add them everywhere.

AnnaShaleva (approved):

superboyiii:
Seems good to merge.

superboyiii (approved):

Jim8y (approved):

vncoelho (approved):

PR #183, opened 2024-09-12
fix status

Jim8y:
this pr fix the status of nep28 and add it to readme.

### DIFF
--- a/README.mediawiki
+++ b/README.mediawiki
@@ -123,6 +123,12 @@ First review [[nep-1.mediawiki|NEP-1]]. Then clone the repository and add your N
 | Standard
 | Final
 |-
+| [[nep-28.mediawiki|28]]
+| CStandardization of the Smallest Unit of GAS
+| Jimmy Liao
+| Standard
+| Final
+|-
 | [[nep-29.mediawiki|29]]
 | Contract deployment/update callback function
 | Roman Khimov
--- a/README.mediawiki
+++ b/README.mediawiki
@@ -124,7 +124,7 @@ First review [[nep-1.mediawiki|NEP-1]]. Then clone the repository and add your N
 | Final
 |-
 | [[nep-28.mediawiki|28]]
-| CStandardization of the Smallest Unit of GAS
+| Standardization of the Smallest Unit of GAS
 | Jimmy Liao
 | Standard
 | Final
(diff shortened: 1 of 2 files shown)


README.mediawiki
```diff
@@ -123,6 +123,12 @@ First review [[nep-1.mediawiki|NEP-1]]. Then clone the repository and add your N
 | Standard
 | Final
 |-
+| [[nep-28.mediawiki|28]]
+| CStandardization of the Smallest Unit of GAS
```
roman-khimov, line 127:
```suggestion
| Standardization of the Smallest Unit of GAS
```

roman-khimov (changes requested):

shargon (dismissed):

roman-khimov (dismissed):

AnnaShaleva (approved):

nep-28.mediawiki
```diff
@@ -2,7 +2,7 @@ NEP: 28
 Title: Standardization of the Smallest Unit of GAS
 Author: Jimmy Liao <jimmy@r3e.network>
 Type: Standard
-Status: Draft
+Status: Final
```
AnnaShaleva, line 5: BTW, while we're here, let's fix the formatting. Standard header should be sorrounded by `<pre>` tag with corresponding indentation for proper MD rendering. Here's how does it looks like right now:
image
And here's how should it be:
image

vncoelho (approved):

nep-28.mediawiki
```diff
@@ -1,9 +1,11 @@
-NEP: 28
-Title: Standardization of the Smallest Unit of GAS
-Author: Jimmy Liao <jimmy@r3e.network>
-Type: Standard
-Status: Draft
-Created: 2024-05-20
+<pre>
+    NEP: 28
+    Title: Standardization of the Smallest Unit of GAS
+    Author: Jimmy Liao <jimmy@r3e.network>
```
shargon, line 4: If it is HTML, this should be escaped?

roman-khimov: All the others are formatted like this and it seems to be OK: https://github.com/neo-project/proposals/blob/master/nep-16.mediawiki

roman-khimov (approved):

AnnaShaleva (approved):

shargon (approved):

PR #189, opened 2025-02-08
[DBFT] Optimize dBFT Consensus Timing Mechanism

Jim8y:
As we plan to reduce the block time in future updates, network latency issues will have a more pronounced impact on block generation speed. This could potentially slow down the block production rate and, consequently, affect the <code>GAS</code> generation rate. Therefore, we need to implement a mechanism to minimize these latency effects and maintain a stable gas generation rate across the network.

### DIFF
--- a/nep-dbft.mediawiki
+++ b/nep-dbft.mediawiki
@@ -1,0 +1,138 @@
+<NEP: xxx>
+  Title: Optimize dBFT Consensus Timing Mechanism
+  Author: Jimmy Liao <vvvincentvan@gmail.com>
+  Status: Draft
+  Created: 2025-02-08
+  Requires: N/A
+  Replaces: N/A
+  Hardfork: HF_Echidna
+
+==Abstract==
+
+This NEP proposes an optimization to the dBFT consensus timing mechanism in Neo N3, to be implemented as part of the HF_Echidna hardfork. The change modifies how consensus rounds are scheduled by integrating the consensus time within the block time, transforming from a sequential approach to an overlapping timeline approach.
+
+==Terminology==
+
+===Time Parameters===
+* Block Time (Tb): The fixed time interval between two consecutive blocks
+* Consensus Time (Tc): The actual time required to complete one round of consensus
+* Previous Implementation: Tb + Tc was the effective time between blocks
+* New Implementation: Tc occurs within Tb, making the effective time between blocks equal to Tb
+
+==Motivation==
+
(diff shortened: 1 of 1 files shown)


roman-khimov (commented):
1. This has nothing to do with hardforks. At all.
2. This doesn't make any sense as a NEP to me. dBFT stays the same, timer management is deeply internal to it and 3.7 will be compatible with 3.8 wrt this even if they slightly deviate.
3. The description is insufficient for any kind of implementation. What is `T(n)` really? It will be different on different machines. What is `Tc`, when consensus is considered to be started and when is it considered to be finished wrt dBFT cycle?

Jim8y:
> 1. This has nothing to do with hardforks. At all.

It needs hardfork to ensure all consensus node works the same.

> 2. This doesn't make any sense as a NEP to me. dBFT stays the same, timer management is deeply internal to it and 3.7 will be compatible with 3.8 wrt this even if they slightly deviate.

Its OK if you dont think it should be this way, its not required to merge a draft. 

> 3. The description is insufficient for any kind of implementation. What is `T(n)` really? It will be different on different machines. What is `Tc`, when consensus is considered to be started and when is it considered to be finished wrt dBFT cycle?

I can update for better description.

Jim8y:
I am trying to make everything standardized, may make many mistakes, not intentional mistake, but appologize for those. 

I have no idea what should be an NEP, what should not be an NEP, its up to the team to decide and i accept the team decision,

roman-khimov:
> It needs hardfork to ensure all consensus node works the same.

Nope. This adjustment is backwards-compatible, for most networks the difference won't affect correctness.

> I am trying to make everything standardized

@Jim8y, don't get me wrong, this is what I totally support and I will help with any protocol formalization efforts. It's just that this particular case is arguable for two reasons: we don't have dBFT NEP and this doesn't change the algorithm itself in any way. It's an implementation detail which has some importance, but maybe not requiring NEP. At the same time, I agree that

> its up to the team to decide

And if the majority find this NEP useful then it should be explained in dBFT terms, so that it could be understood and implemented.

shargon (commented):
This is already merged in `HF_Echidna` should we move on?

Jim8y:
Closing this draft because the timing mechanism is already covered by the HF_Echidna work and the proposal does not define a protocol-level dBFT change clearly enough for a NEP. If this needs to return, it should be rewritten around concrete dBFT semantics rather than hardfork framing.

PR #190, opened 2025-02-11
[Protocol] 3 seconds dbft consensus

Jim8y:
This pr tries to formalize our so called 3-seconds consensus. 

It requires hardfork and:

- reduce the block time from 15 seconds to 3 seconds
- reduce the gasperblock from 5 GAS to 1 GAS

Affected Configurations:

* MillisecondsPerBlock: 15000 => 3000
* MaxTransactionsPerBlock: 512 => 256
* GasPerBlock: 
  * Current: 5 GAS
  * At hardfork: Automatically reduced to 1 GAS
  * Post-hardfork: Configurable by council
* MaxValidUntilBlockIncrement: 100 => 500?

### DIFF
--- a/nep-3-seconds.mediawiki
+++ b/nep-3-seconds.mediawiki
@@ -10,8 +10,11 @@
 ==Abstract==
 
 This NEP proposes two significant changes to the Neo N3 network parameters:
-1. Reducing the block time from 15 seconds to 3 seconds
-2. Reducing the GAS generation rate per block from 5 GAS to 1 GAS
+1. Enabling governance mechanism for block time adjustment via the Policy contract
+2. Preparing for future reduction of block time from 15 seconds to 3 seconds
+3. Preparing for future reduction of GAS generation rate per block from 5 GAS to 1 GAS
+
+The hardfork will introduce the governance methods while maintaining current values. After the hardfork, the Neo Council will be responsible for executing the parameter changes through the governance mechanism.
 
 ==Motivation==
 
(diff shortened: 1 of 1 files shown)


roman-khimov:
Isn't it a _policy_ change, rather than a protocol change? Do we track policy changes in NEPs?

roman-khimov:
Maybe

> An Informational NEP describes a NEO design issue, or provides general guidelines or information to the NEO community, but does not propose a new feature. Informational NEPs do not necessarily represent a NEO community consensus or recommendation, so users and implementors are free to ignore Informational NEPs or follow their advice.

But then this means that any mainnet policy changes (like fees) should be described in NEPs. Which would be nice, but are we ready for it?

AnnaShaleva:
> Which would be nice, but are we ready for it?

It seems that we're not yet ready. We'd better start with NEPs for protocol changes only, because otherwise we'll end up in a very large amount of pending NEPs describing all Neo policies that need to be created/reviewed/approved.

So vote up for not including this change into NEPs.

cschuchardt88:
`PolicyContract` doesn't hold this information. It's a configuration in a file. How else you going to change either setting for privatenet?

shargon (dismissed):

Wi1l-B0t (dismissed):

superboyiii:
Yes, we have set then we should have get.

nep-3-seconds.mediawiki
```diff
@@ -0,0 +1,177 @@
...
+===GAS Generation Rate Reduction===
+* Current value: 5 GAS per block
+* Proposed value: 1 GAS per block
+* Implementation: 
+  * Retain the existing governance model where council controls GasPerBlock
+  * Council will need to adjust GasPerBlock separately after block time changes
+  * Coordination between the two changes will be managed through governance
+
+===Policy Contract Enhancement===
+* New native method: `SetBlockTime(BigInteger milliseconds)`
```
superboyiii, line 42:
```suggestion
* New native method: `SetBlockGenTime(BigInteger milliseconds)`
```
I prefer a more clearer name since we already have a similar name of method `GetBlock` in `native Ledger`

nep-3-seconds.mediawiki
```diff
@@ -0,0 +1,177 @@
...
+1. Network Stability
+* Issue: Faster block times may increase network overhead
+* Mitigation: Implement improved network optimization and node requirements
+
+2. Smart Contract Timing
+* Issue: Some contracts may assume 15-second blocks
+* Mitigation: Provide migration guidelines and testing framework
+
+3. Network Synchronization
+* Issue: Faster blocks may impact node sync speed
```
superboyiii, line 149:
```suggestion
* Issue: Faster blocks may impact node sync speed since more block data will be generated.
```

nep-3-seconds.mediawiki
```diff
@@ -0,0 +1,177 @@
...
+   * Consensus nodes MUST use the new block time for the next consensus round
+
+4. Return Value:
+   * Returns true if the operation is successful
+   * Returns false if any validation fails
+
+5. Effect on Contracts:
+   * Contract executions MUST adapt to the new block time for all time-related calculations
+   * The consensus mechanism MUST read the block time from the Policy contract for each consensus round
+
```
superboyiii, line 118:
```suggestion

===Events===

====Transfer====

<pre>
{
  "name": "SetBlockGenTime",
  "parameters": [
    {
      "name": "milliseconds",
      "type": "Integer"
    }
  ]
}
</pre>

```
We'd better add a eventlog in the end of the method.

Jim8y:
Closing this for now. The discussion shows this is better treated as a policy/governance decision than a protocol NEP unless the project decides to standardize policy changes through NEPs.

PR #191, opened 2025-02-14
proposal: Unified Neo chain dump format

AliceInHunterland:


ixje:
Adding the network magic as part of the header would allow to improve on https://github.com/nspcc-dev/neo-go/issues/3499#issuecomment-2199495544

Maybe even a file magic to start with? `4e454f3344554d50` (`NEO3DUMP`) fits nicely

AliceInHunterland:
We have an official .acc format that has been in use for years, compresses more efficiently than multiple-file ZIP archives (which compress each file individually), and supports streaming for parallel reading. Additional metadata like network or checksums can be added without discarding .acc. The .acc file already meets current needs and can be zipped more efficiently if desired. A manifest is unnecessary; but if required, it can be addressed in a future NEP (for example, to include network ID and avoid mixing dumps) as well as adding support of zip format by CLI.

roman-khimov:
> Adding the network magic as part of the header

Nice one, but that'd be a _new_ file format. Not sure we're ready for it given all the infrastructure built around current one.

cschuchardt88:
I have had a lot of time building and testing which data is necessary for backup at a minimum for `csharp` node, based off how it works at importing data.

### Zip Entries
#### Manifest
- filename: `manifest` or `MANIFEST`
- contents: `Network`, `Version` (_of manifest format_), `Checksum` (_of manifest header_) and `BlockChecksumTable` (_ checksums of all the blocks in this zip file_).

### Blocks
- filename: `{index}`
  - format of `{index}` - was be just `Block.Index` with `ToString("d")`
- contents: `Block.ToArray()`

### Process
1. Read the `manifest` file in the zip.
2. Verify all `manifest` header, but not `BlockChecksumTable`.
3. Get all filename entries in zip
4. Order filenames by ascending.
5. Get entry data from zip
6. compare `manifest.BlockChecksumTable[index]` with an checksum of entry data
7. Load into blockchain

### Options
1. Be able to `disable` verification on blocks
2. Be able to `disable` verification on checksum of blocks.

### Checksum fails
- Stop the process of loading

### Zip Filename formats
- nonincremental: `{name}.arc`
- incremental: `{name}.{index}.arc` - start from `0`

#### Example
- `MyBackup.arc`
- `MyBackup.0.arc`
- `MyBackup.1.arc`

#### Example Entry Filenames in (`MyBackup.arc`)
- `MANIFEST`
- `1`
- `2`
- `3`
- `4`
- `6`

#### Example Entry Filenames in (`MyBackup.0.arc`)
- `7`
- `8`

#### Example Entry Filenames in (`MyBackup.1.arc`)
- `9`
- ... etc

roman-khimov:
> PS: something like this WILL be add in csharp node.

I don't think so.

1. We already have a format that is widely available and supported.
2. We can discuss some enhancements for it, but zip files are absolutely unacceptable to me for the purpose. They're _not_ faster in any way, in fact our acc format is easier to create and process than your proposal. Zip requires more data to be transmitted compared to compressed acc (and compression is just a separate matter, not something for this format to worry about) and our good old acc can be perfectly streamed (processed as you get it from elsewhere, NeoGo in fact supports this mode of operation, you can `curl https://example.com/mainnet.acc | neo-go db restore --config-file config/protocol.mainnet.yml`), which zip can't do.
3. You do not need any checksums for blocks, they're all hashed and signed and checked already.

cschuchardt88:
> 1. We already have a format that is widely available and supported.

So? This second options and old one can still be used.

> 2. We can discuss some enhancements for it, but zip files are absolutely unacceptable to me for the purpose. They're _not_ faster in any way, in fact our acc format is easier to create and process than your proposal. Zip requires more data to be transmitted compared to compressed acc (and compression is just a separate matter, not something for this format to worry about) and our good old acc can be perfectly streamed (processed as you get it from elsewhere, NeoGo in fact supports this mode of operation, you can `curl https://example.com/mainnet.acc | neo-go db restore --config-file config/protocol.mainnet.yml`), which zip can't do.

1. This isn't a problem for csharp. 
2. Don't know the network.
3. Don't know how many blocks.
4. Easy to corrupt.
5. Already uses zip format
6. No integrity checks
7. No Protection
8. Can Repair
9. Can Recover
10. You can do `curl http:/example.com/main.0.arc | some-command` Use `application/octet-stream` or https://www.iana.org/assignments/media-types/application/zip which is already widely supported.

> 3. You do not need any checksums for blocks, they're all hashed and signed and checked already.

No sense in trying to process corrupt data. If the data is corrupted it could request the block from the network. But seeing how we don't know; we could use `ProcolSettings` for that. But, if you are syncing a different network than the one you're on. Because it only verifies locally until sync is completed than it will connect to the network and request new blocks. OOps I started at Genesis and synced the wrong network's blocks. All that time wasted.

roman-khimov:
> So? This second options and old one can still be used.

All of the community will definitely love having two incompatible solutions for the same (very simple in its essence) problem. Or not? What this PR tries to do is exactly to avoid this ever happening.

> Don't know the network.

That's true. As said above already, this PR doesn't even try proposing any _new_ format. Because even though technically there is no network data in this dump, what happens in real life when you're trying to use an inappropriate one is https://github.com/nspcc-dev/neo-go/issues/3499. It just doesn't work.

> Don't know how many blocks.

Not true. It's in the header if you haven't noticed.

> Easy to corrupt.

Not true. Just to remind, we're talking about blockchain, each block can be verified on its own. And just to remind, when you're synchronizing over P2P any peer can send you any kind of garbage instead of a proper block and this is never a problem. All of your points about corrupt data are incorrect, the chain won't process an invalid block.

> Already uses zip format

Not true. It's just that sync.ngd.network zips these files for your convenience (and to minimize traffic obviously). I'd rather gzip them. Someone might like xz a bit more. But it's irrelevant, C# happily accepts raw acc files already. But it also supports zipped ones transparently, OK. NeoGo doesn't, for example, we deliberately don't care about the way you archive your dumps.

But as already noticed above this acc+zip combination is far more efficient in compression than "one block --- one file" approach.

> Use application/octet-stream or https://www.iana.org/assignments/media-types/application/zip which is already widely supported.

This problem has nothing to do with `Content-Type`. We're talking about streams here. What I mean by 
```
$ curl https://example.com/mainnet.acc | neo-go db restore --config-file config/protocol.mainnet.yml
```
(which can even be more like
```
$ curl https://example.com/mainnet.acc.xz | unxz | neo-go db restore --config-file config/protocol.mainnet.yml
```
) is that there is no need to download the whole dump in advance and store it locally. The last process in this chains reads block from input one by one (exactly as they're stored in .acc) and processes them. ZIP files do not allow for that easily, you will have to ensure metadata (Central Directory, right?) is stored in the beginning of the file and then to ensure that blocks are laid out consecutively in the archive. While technically possible, it's not something you get out of box with ZIPs. And as mentioned already numerous times it's horribly inefficient compression-wise.

> What if you are syncing a different network than the one you're on.

It doesn't work (and C# node will react in a similar fashion). https://github.com/nspcc-dev/neo-go/issues/3499

cschuchardt88:
>Not true. It's in the header if you haven't noticed.

Filename only has starting block index. Nothing else.

>Not true. Just to remind, we're talking about blockchain, each block can be verified on its own. And just to remind, when you're >synchronizing over P2P any peer can send you any kind of garbage instead of a proper block and this is never a problem. All of >your points about corrupt data are incorrect, the chain won't process an invalid block.

We not be using the same format than. File is an array of data of only Prefix data length and than block data. If length is incorrect how you going to find a block? there is no magic code to search for.

>This problem has nothing to do with Content-Type. We're talking about streams here. What I mean by

Like I said csharp doesn't have this problem and can stream zip files.

> is that there is no need to download the whole dump in advance and store it locally.

Content-Type plays a big roles in how data is chunked, read and processed by an http server. I spend more than a decade developing on http servers, you have to enable the feature (_at least in csharp_) or maybe your http server doesnt support this feature.

>ZIP files do not allow for that easily, you will have to ensure metadata (Central Directory, right?) is stored in the beginning of the file and then to ensure that blocks are laid out consecutively in the archive.

a zip file on your machine is different than http `Content-Type` of `application/zip` I suggest read about that `Content-Type` and how that protocol works.

>And as mentioned already numerous times it's horribly inefficient compression-wise.

You don't have to use compression. Plus we are not zipping one file, like it is now that is 2GB+ big. We are talking small data mostly empty blocks.

Not having standards or having proprietary things will stop people from innovating with `neo` (_current state now_). So your opinion is noted.

roman-khimov:
> Filename only has starting block index. Nothing else.

But that's just not true! Please read this PR, Neo-CLI code handling acc and https://github.com/neo-project/neo/issues/2998.

roman-khimov:
Regarding zips and streaming I'll just leave this link here, btw:
https://github.com/dotnet/runtime/blob/1d1bf92fcf43aa6981804dc53c5174445069c9e4/src/libraries/System.IO.Compression/src/System/IO/Compression/ZipArchive.cs#L141-L147

dump-format.mediawiki
```diff
@@ -0,0 +1,66 @@
...
+  Author: Ekaterina Pavlova <ekt@nspcc.ru>, Roman Khimov <roman@nspcc.ru>
+  Type: Informational
+  Status: Draft
+  Created: 2025-02-14
+</pre>
+
+==Abstract==
+
+This proposal standardizes a single Neo chain dump format that unifies the “full” and “incremental”
+approaches into one consistent structure. It enables different Neo node implementations to remain
```
AnnaShaleva, line 13:
```suggestion
approaches into one consistent structure. It enforces different Neo node implementations to remain
```

dump-format.mediawiki
```diff
@@ -0,0 +1,66 @@
+<pre>
+  NEP: TBD
+  Title: Unified Neo chain dump format
```
AnnaShaleva, line 3: Just a concern: we're going to have state dump format as far, so may be it's better to use "block dump" format instead of "chain dump" in order to avoid confusion? Although "chain dump" is how we name it from the beginning of Neo.

@roman-khimov

dump-format.mediawiki
```diff
@@ -0,0 +1,66 @@
...
+and an incremental dump from an arbitrary block (e.g., chain.X.acc). This dual approach complicates
+tooling and confuses operators (e.g., chain.0.acc is actually incremental). This NEP unifies both
+formats into one, making dumps interchangeable, simplifying import/restore, and reducing confusion.
+
+==Specification==
+
+===Dump Structure===
+
+==== Dump Header ====
+* '''Start (uint32, LE)''': The first block index included in the dump.
```
AnnaShaleva, line 29:
```suggestion
* '''Start (uint32, LE)''': Index of the first block included in the dump.
```

dump-format.mediawiki
```diff
@@ -0,0 +1,66 @@
...
+* '''Count (uint32, LE)''': The total number of blocks contained in the dump.
+
+==== Blocks ====
```
AnnaShaleva, line 32: I'd expect the format to be somehow similar to https://github.com/neo-project/proposals/blob/master/nep-16.mediawiki#neffile-structure. It's just an idea to keep proposals format unified.

dump-format.mediawiki
```diff
@@ -0,0 +1,66 @@
...
+All blocks in the specified range <code>[start, start + count - 1]</code> MUST be
+serialized in ascending order immediately after the header.
+
+
+For each block in <code>[Start, Start + Count - 1]</code>:
```
AnnaShaleva, line 37: It's either `start` and `count` or `Start` and `Count`.

dump-format.mediawiki
```diff
@@ -0,0 +1,66 @@
...
+* A 4-byte length prefix (<code>uint32</code>, LE).
+* The block’s serialized data (length bytes).
```
AnnaShaleva, line 39: What does `(length bytes)` mean in this context?

dump-format.mediawiki
```diff
@@ -0,0 +1,66 @@
...
+
+
+Nodes MUST parse the first two integers to determine the dump’s block range.
+No assumptions about file names (<code>chain.acc</code>, <code>chain.X.acc</code>)
+are allowed; the header is the sole source of truth.
```
AnnaShaleva, line 44: I love that, we need this proposal.

dump-format.mediawiki
```diff
@@ -0,0 +1,66 @@
...
+Nodes treated “full” vs. “incremental” dumps differently, confusing — especially with
+“official” dumps from [https://sync.ngd.network/ sync.ngd.network].
+Unifying on one incremental format (using start=0 for a “full” dump) removes
+filename dependencies, ensures consistent tooling, and delivers a
+straightforward “download/restore/run” experience.
+
+== Backwards Compatibility ==
+Nodes relying on <code>chain.acc</code> or <code>chain.X.acc</code> can still import
+dumps by reading their header. Legacy tools may need a transition period, but moving
+to a single incremental format does not break existing usage and simplifies future interoperability.
```
AnnaShaleva, line 56: Line width.

dump-format.mediawiki
```diff
@@ -0,0 +1,66 @@
...
+
+== References ==
+* [https://sync.ngd.network/ Neo Official Dumps]
+* [https://github.com/neo-project/neo/issues/2998 Neo Issue#2998]
+* [https://github.com/neo-project/neo-node/issues/536 Neo Issue#536]
+* [https://github.com/nspcc-dev/neo-go/issues/2849 Neo Go Issue#2849]
```
AnnaShaleva, line 62: It's NeoGo.

dump-format.mediawiki
```diff
@@ -0,0 +1,66 @@
...
+
+== Implementation ==
+* C#: https://github.com/neo-project/neo/blob/2cb647614db6bcdb90d276ac809e78b277408305/src/Neo.CLI/CLI/MainService.Blockchain.cs#L30
+* Go: https://github.com/nspcc-dev/neo-go/blob/ebb71497a531edd10fc72fddd50c18c8e41c8356/cli/server/server.go#L164
```
AnnaShaleva, line 66: Add links to blocks import as far.

AnnaShaleva (commented):
I love this proposal, we need it to be unified, because otherwise there's an inambiguity between two different formats.

dump-format.mediawiki
roman-khimov: `enables` seems to be more correct to me

dump-format.mediawiki
roman-khimov: Check it against CLI and docs. It's known as "chain dump", "block dump", "offline package", but likely some term is prevailing. "State dump" is different, but that doesn't intersect with others to me. We never call this one "state dump".

dump-format.mediawiki
cschuchardt88: The size of the block in bytes (_`Block.ToArray().length`_). The prefix is the type, in this case it's `uint32`

dump-format.mediawiki
cschuchardt88: Localcation is here 

https://github.com/neo-project/neo/blob/2e2fbe9552110c4f0c34468722e9900ca3ab0bf1/src/Neo.CLI/CLI/MainService.cs#L187-L235

Jim8y (dismissed):

vncoelho (commented):
I will review and approve asap

dump-format.mediawiki
```diff
@@ -0,0 +1,90 @@
...
+}
+</pre>
+
+====start====
+The <code>start</code> item represents the index of the first block
+included in the dump. It is stored as a <code>uint32</code>.
+
+====count====
+The <code>count</code> item denotes the total number of blocks
+contained in the dump. It is stored as a <code>uint32</code>.
```
AnnaShaleva, line 46: We need to specify the endianness for all integers, it's a part of the standard. For example, NEF standard also contains endianness specification:
https://github.com/neo-project/proposals/blob/ff39b57eb67981b1ec7f0b244ec053aacdb7139a/nep-16.mediawiki?plain=1#L26

dump-format.mediawiki
```diff
@@ -0,0 +1,90 @@
...
+and an incremental dump from an arbitrary block (e.g., chain.X.acc). This dual approach complicates
+tooling and confuses operators (e.g., chain.0.acc is actually incremental). This NEP unifies both
+formats into one, making dumps interchangeable, simplifying import/restore, and reducing confusion.
+
+==Specification==
+
+===Dump Structure===
+
+<pre>
+dump_file
```
AnnaShaleva, line 29: We need at least short description, like it is done in
https://github.com/neo-project/proposals/blob/ff39b57eb67981b1ec7f0b244ec053aacdb7139a/nep-16.mediawiki?plain=1#L26

dump-format.mediawiki
```diff
@@ -0,0 +1,90 @@
...
+{
+  uint32 start;
```
AnnaShaleva, line 31:
```suggestion
  4-byte start;
```
and etc. for the fields below, to be unified with
https://github.com/neo-project/proposals/blob/ff39b57eb67981b1ec7f0b244ec053aacdb7139a/nep-16.mediawiki?plain=1#L31C3-L31C9

dump-format.mediawiki
AnnaShaleva: I know what it is. But it is not clear from the description.

dump-format.mediawiki
```diff
@@ -0,0 +1,90 @@
...
+The <code>length</code> item is a <code>uint32</code> indicating the
+length of the serialized block data.
+
+=====data=====
+The <code>data</code> item contains the serialized block data, with a
+length specified by the <code>length</code> field.
+
+
+Nodes MUST parse the first two integers to determine the dump’s block range.
+ Multibyte data items are always stored in little-endian order.
```
AnnaShaleva, line 63: Useless space in the start of the line.

dump-format.mediawiki
cschuchardt88: Also it encode `integer` for example has a `prefix` for `type` of `integer` and its `compressed`

dump-format.mediawiki
```diff
@@ -0,0 +1,94 @@
...
+Neo nodes have historically used two chain dump formats: a full dump from block 0 (e.g., chain.acc)
+and an incremental dump from an arbitrary block (e.g., chain.X.acc). This dual approach complicates
+tooling and confuses operators (e.g., chain.0.acc is actually incremental). This NEP unifies both
+formats into one, making dumps interchangeable, simplifying import/restore, and reducing confusion.
+
+==Specification==
+
+===Dump Structure===
+
+Multibyte data items are always stored in little-endian order, where the low bytes come first.
```
AnnaShaleva, line 28: Add `A dump file consists of a stream of 8-bit bytes.` at the start of the line to follow https://github.com/neo-project/proposals/blob/master/nep-16.mediawiki#neffile-structure ?

AnnaShaleva (commented):
LGTM otherwise.

dump-format.mediawiki
roman-khimov: We can end up with "octets" this way.

cschuchardt88 (approved):

dump-format.mediawiki
```diff
@@ -0,0 +1,94 @@
...
+  4-byte count;
+  block{
+    4-byte length;
+    length-byte data;
+  } blocks[];
+}
+</pre>
+
+====start====
+The <code>start</code> item represents the index of the first block
```
shargon, line 44: `VarSize` could save some space

dump-format.mediawiki
```diff
@@ -0,0 +1,94 @@
...
+===Dump Structure===
+
+Multibyte data items are always stored in little-endian order, where the low bytes come first.
+Its structure is described further below.
+
+<pre>
+dump_file
+{
+  4-byte start;
+  4-byte count;
```
shargon, line 35: if we stream it, this can be -1?

dump-format.mediawiki
AnnaShaleva: It's only 4 bytes, and this format was used before, so we don't think that it's worth changing it right now.

dump-format.mediawiki
AnnaShaleva: This file contains the restricted number of blocks anyway.

AnnaShaleva (approved):

dump-format.mediawiki
```diff
@@ -0,0 +1,94 @@
...
+are allowed; the header is the sole source of truth.
+
+== Rationale ==
+
+Nodes treated “full” vs. “incremental” dumps differently, confusing — especially with
+“official” dumps from [https://sync.ngd.network/ sync.ngd.network].
+Unifying on one incremental format (using start=0 for a “full” dump) removes
+filename dependencies, ensures consistent tooling, and delivers a
+straightforward “download/restore/run” experience.
+
```
vncoelho, line 76: I think that some real examples of names are missing here to fully clarify everything

AliceInHunterland: We do have filename examples in the Motivation section above.

dump-format.mediawiki
shargon: Agree, `start` is only 4, but block length is 4 for each block (6m now), a full chain could save some `mb`.

Jim8y (approved):

Jim8y:
@shargon this has being discussed and agreed to process one month ago, please consider to take a look, if you are cool with it, consider to merge after your approval. Minor typo issuss or any other non-critical issues can be done afterwards. Lets us not block it for non-criticial things.

PR #194, opened 2025-04-08
proposal: NeoFS block storage format

AliceInHunterland:


neofs-blocks.mediawiki
```diff
@@ -0,0 +1,58 @@
...
+ - second-precision block uploading timestamp (`Timestamp:1627894840`)
+
+These attributes are stored as NeoFS object metadata alongside the binary block data.
+Multibyte data items are always stored in little-endian order, where the low bytes come first.
+Its structure is described further below.
+
+<pre>
+block_binary
+{
+  var-byte header;            // Fixed-size header including witnesses
```
shargon, line 43:
```suggestion
  var-byte header;             // Fixed-size header including witnesses
```

neofs-blocks.mediawiki
```diff
@@ -0,0 +1,58 @@
...
+ - previous block hash in the LE form (`PrevHash:3654a054d82a8178c7dfacecc2c57282e23468a42ee407f14506368afe22d929`)
+ - millisecond-precision block creation timestamp (`BlockTime:1627894840919`)
+ - second-precision block uploading timestamp (`Timestamp:1627894840`)
+
+These attributes are stored as NeoFS object metadata alongside the binary block data.
+Multibyte data items are always stored in little-endian order, where the low bytes come first.
+Its structure is described further below.
+
+<pre>
+block_binary
```
roman-khimov, line 41: I think you can just refer to "standard Neo block serialization", that's what it is in its essence.

neofs-blocks.mediawiki
```diff
@@ -0,0 +1,58 @@
...
+  Title: NeoFS block storage format
+  Author: Ekaterina Pavlova <ekt@nspcc.ru>
+  Type: Informational
+  Status: Draft
+  Created: 2025-04-07
+</pre>
+
+==Abstract==
+
+This proposal outlines the methodology and specifications for storing Neo blockchain blocks within
```
roman-khimov, line 12: `methodology` is too strong word for this.

neofs-blocks.mediawiki
```diff
@@ -0,0 +1,58 @@
...
+the NeoFS container.
+
+==Motivation==
+
+Neo node synchronization via P2P requires all block headers (~700 bytes each, over 1 GB per year with
```
roman-khimov, line 17: headers and blocks

neofs-blocks.mediawiki
```diff
@@ -0,0 +1,58 @@
...
+15-second blocks) to remain available, limiting scalability as chains grow. Storing blocks in NeoFS
+provides a distributed alternative, reducing local storage, enabling on-demand fetches, and speeding
+up sync with snapshots or checkpoints. This proposal defines a NeoFS storage schema, building on
```
roman-khimov, line 20: snapshots and checkpoints are for a different proposal.

neofs-blocks.mediawiki
```diff
@@ -0,0 +1,58 @@
...
+experiments (e.g., nspcc-dev/neo-go#3496) and future enhancements (#1526).
```
roman-khimov, line 21: Experiments are for a different proposal section (rationale).

neofs-blocks.mediawiki
```diff
@@ -0,0 +1,58 @@
...
+==Abstract==
+
+This proposal outlines the methodology and specifications for storing Neo blockchain blocks within
+the NeoFS container.
+
+==Motivation==
+
+Neo node synchronization via P2P requires all block headers (~700 bytes each, over 1 GB per year with
+15-second blocks) to remain available, limiting scalability as chains grow. Storing blocks in NeoFS
+provides a distributed alternative, reducing local storage, enabling on-demand fetches, and speeding
```
roman-khimov, line 19: allowing to drop blocks and headers from storage for most nodes

neofs-blocks.mediawiki
```diff
@@ -0,0 +1,58 @@
...
+up sync with snapshots or checkpoints. This proposal defines a NeoFS storage schema, building on
+experiments (e.g., nspcc-dev/neo-go#3496) and future enhancements (#1526).
+
+==Specification==
+
+===Block Storage Schema===
+A single NeoFS container is used to store blocks. Each container
```
roman-khimov, line 26: per network

neofs-blocks.mediawiki
```diff
@@ -0,0 +1,58 @@
...
+has network magic attribute (`Magic:56753`). Each block is stored in a binary
+form as a separate object with a unique OID and a set of attributes:
+ - block object identifier with block index value (`Block:1`)
+ - primary node index (`Primary:0`)
```
roman-khimov, line 30: We have it now, but I'm still thinking if its useful. I can imagine searches for block hashes/timestamps, but primary index is not so obvious (especially without a verification script hash).

neofs-blocks.mediawiki
```diff
@@ -0,0 +1,58 @@
...
+up sync with snapshots or checkpoints. This proposal defines a NeoFS storage schema, building on
+experiments (e.g., nspcc-dev/neo-go#3496) and future enhancements (#1526).
+
+==Specification==
+
+===Block Storage Schema===
+A single NeoFS container is used to store blocks. Each container
+has network magic attribute (`Magic:56753`). Each block is stored in a binary
+form as a separate object with a unique OID and a set of attributes:
+ - block object identifier with block index value (`Block:1`)
```
roman-khimov, line 29: Define field names first, then describe them. Examples are for a different section.

neofs-blocks.mediawiki
```diff
@@ -0,0 +1,58 @@
...
+ - primary node index (`Primary:0`)
+ - block hash in the LE form (`Hash:5412a781caf278c0736556c0e544c7cfdbb6e3c62ae221ef53646be89364566b`)
```
roman-khimov, line 31: LE/BE are a bit different in C#, we can say "in hex hash representation"

neofs-blocks.mediawiki
```diff
@@ -0,0 +1,58 @@
...
+ - previous block hash in the LE form (`PrevHash:3654a054d82a8178c7dfacecc2c57282e23468a42ee407f14506368afe22d929`)
+ - millisecond-precision block creation timestamp (`BlockTime:1627894840919`)
+ - second-precision block uploading timestamp (`Timestamp:1627894840`)
+
+These attributes are stored as NeoFS object metadata alongside the binary block data.
+Multibyte data items are always stored in little-endian order, where the low bytes come first.
+Its structure is described further below.
```
roman-khimov, line 38: Add a not that the object MAY have additional data after the block, it can be useful for future extensions.

neofs-blocks.mediawiki
```diff
@@ -0,0 +1,58 @@
...
+
+<pre>
+block_binary
+{
+  var-byte header;             // Fixed-size header including witnesses
+  var-byte txcount;            // Transaction count
+  var-byte transactions[];     // Array of serialized transactions
+}
+</pre>
+
```
roman-khimov, line 48: We need an example of object.

roman-khimov (dismissed):
Index files? Synchronization algorithm?

Jim8y:
@AliceInHunterland please update accordingly, this is not likely can be reviewed with any useful suggestions from other coredevs since we know little about neofs, please just follow roman's suggestion

neofs-blocks.mediawiki
```diff
@@ -0,0 +1,115 @@
...
+Neo node synchronization via P2P requires all headers and blocks (~700 bytes each, over 1 GB per year with
+15-second blocks) to remain available, limiting scalability as chains grow. Storing blocks in NeoFS
+provides a distributed alternative, reducing local storage, enabling on-demand fetches and allowing to drop
+blocks and headers from storage for most nodes.
+
+==Specification==
+
+===Block Storage Schema===
+A single NeoFS container is used to store blocks and index files per network. Each container
+has network magic attribute (`Magic:56753`). Each block is stored in a binary
```
AnnaShaleva: Replace direct example of `Magic:56753` with description of attribute name and the form of attribute value, because this section is a specification description, examples shouldn't be placed here.

neofs-blocks.mediawiki
```diff
@@ -0,0 +1,115 @@
...
+form as a separate object with a unique OID and a set of attributes:
```
AnnaShaleva: > in a binary form

in a binary form following standard Neo block serialization rules

neofs-blocks.mediawiki
```diff
@@ -0,0 +1,115 @@
...
+* Block: block object identifier with block index value.
+* Hash: block hash in the in hex hash representation.
+* PrevHash: previous block hash in the in hex hash representation.
+* BlockTime: millisecond-precision block creation timestamp.
+* Timestamp: second-precision block uploading timestamp.
```
AnnaShaleva: Ref. https://github.com/neo-project/proposals/pull/194#discussion_r2033607310, from this description it's not quite clear that `Block`, `Hash`, etc. are attribute names and they are strictly fixed. I'd say that we either need additional note about it or may be even a table, something like that:

| Attribute name | Attribute value |
| --- | --- |
| Block | String representation of block index in the base of 10. |
...


The same stile can be applied to container attributes.

neofs-blocks.mediawiki
```diff
@@ -0,0 +1,115 @@
...
+These attributes are stored as NeoFS object metadata alongside the binary block data and can be extended.
+Multibyte data items are always stored in little-endian order, where the low bytes come first.
+Its structure is standard Neo block serialization and the object may have additional data after the block.
+
+An [https://rest.fs.neo.org/v1/objects/3RCdP3ZubyKyo8qFeo7EJPryidTZaGCMdUjqFJaaEKBV/by_id/Fu7yQzspvLJwSGJNK64xeeyMdWXtU5B5b1es6KSxUag1 example]
+of a block object is as follows:
+<pre>
+ID: Fu7yQzspvLJwSGJNK64xeeyMdWXtU5B5b1es6KSxUag1
+CID: 3RCdP3ZubyKyo8qFeo7EJPryidTZaGCMdUjqFJaaEKBV
+Owner: NVvY1FF67XJ2GTVhy9FqiZGC4jEQtvjmHt
```
AnnaShaleva: Is it correct to attach object link as an example? This object/container may be removed in future whereas we expect NEP to be consistent. @roman-khimov?

neofs-blocks.mediawiki
```diff
@@ -0,0 +1,115 @@
...
+A single NeoFS container is used to store blocks and index files per network. Each container
+has network magic attribute (`Magic:56753`). Each block is stored in a binary
+form as a separate object with a unique OID and a set of attributes:
+* Block: block object identifier with block index value.
+* Hash: block hash in the in hex hash representation.
+* PrevHash: previous block hash in the in hex hash representation.
+* BlockTime: millisecond-precision block creation timestamp.
+* Timestamp: second-precision block uploading timestamp.
+
+These attributes are stored as NeoFS object metadata alongside the binary block data and can be extended.
```
AnnaShaleva:
```suggestion
These attributes are stored as NeoFS object metadata; the list of attributes MAY be extended.
```

neofs-blocks.mediawiki
```diff
@@ -0,0 +1,115 @@
...
+Multibyte data items are always stored in little-endian order, where the low bytes come first.
+Its structure is standard Neo block serialization and the object may have additional data after the block.
```
AnnaShaleva:
```suggestion
The object consists of block serialized following the standard Neo block serialization format. The object MAY include additional data after the block.
```

neofs-blocks.mediawiki
```diff
@@ -0,0 +1,115 @@
...
+has network magic attribute (`Magic:56753`). Each block is stored in a binary
+form as a separate object with a unique OID and a set of attributes:
+* Block: block object identifier with block index value.
+* Hash: block hash in the in hex hash representation.
+* PrevHash: previous block hash in the in hex hash representation.
+* BlockTime: millisecond-precision block creation timestamp.
+* Timestamp: second-precision block uploading timestamp.
+
+These attributes are stored as NeoFS object metadata alongside the binary block data and can be extended.
+Multibyte data items are always stored in little-endian order, where the low bytes come first.
```
AnnaShaleva: s/low/least significant

AnnaShaleva: And I'm not sure we need this sentence here, @roman-khimov?

neofs-blocks.mediawiki
```diff
@@ -0,0 +1,115 @@
...
+</pre>
+
+Each index file is an object containing a constant-sized batch of raw block object
+IDs in binary form ordered by block index. Each index file is marked with the
+following attributes:
+* Index: index file identifier with consecutive file index value.
+* IndexSize: the number of OIDs included into index file.
+* Timestamp: second-precision index file uploading timestamp.
+
+An [https://rest.fs.neo.org/v1/objects/3RCdP3ZubyKyo8qFeo7EJPryidTZaGCMdUjqFJaaEKBV/by_id/4nxxFWRFbNbcfwf5EuftboHQVnnsH6vDw5Dp6mPTuwYN example]
```
AnnaShaleva: We don't have index files anymore.

neofs-blocks.mediawiki
```diff
@@ -0,0 +1,115 @@
...
+</pre>
+
+===Synchronization Algorithm===
+Blocks stored in NeoFS and those synchronized via P2P both use the standard Neo block
+serialization format, ensuring compatibility across different synchronization methods.
+The NeoFS container serves as verified storage for downloading blocks, with two methods
+for block retrieval:
+ - Index File Search: Searches for index files by index file attribute and reads block
+ OIDs from index file object-by-object.
+ - Direct Block Search: Searches blocks one by one directly by block attribute.
```
AnnaShaleva: Index files support is removed.

neofs-blocks.mediawiki
```diff
@@ -0,0 +1,115 @@
...
+
+Downloaded blocks are inserted into the blockchain using the same logic as in the
+P2P synchronization protocol.
+
+==Rationale==
+This proposal defines a NeoFS storage schema, building on experiments (e.g., nspcc-dev/neo-go#3496)
+and future enhancements (#1526).
```
AnnaShaleva: It's not the proper section for the rationale. Also, the text itself may be improved, see as an example https://github.com/neo-project/proposals/blob/master/nep-25.mediawiki#user-content-Rationale.

neofs-blocks.mediawiki
```diff
@@ -0,0 +1,115 @@
...
+== References ==
+* [https://github.com/neo-project/neo/issues/3463 Neo Issue#3463]
+* [https://github.com/neo-project/neo/issues/2373 Neo Issue#2373]
+* [https://github.com/neo-project/neo/issues/1526 Neo Issue#1526]
+* [https://github.com/nspcc-dev/neo-go/issues/3496 NeoGo Issue#3496]
+* [https://github.com/nspcc-dev/neofs-api/issues/285 NeoFS API Issue#285]
+
+
+== Implementation ==
+* Go: https://github.com/nspcc-dev/neo-go/blob/61992405983ddfb5547ae3afab75c22186dc966f/cli/util/uploader.go
```
AnnaShaleva: Do we need to attach an example of NeoFSBlockFetcher as far? I think it will be useful since synchronisation algorythm is also a part of this proposal. @roman-khimov?

neofs-blocks.mediawiki
```diff
@@ -0,0 +1,115 @@
...
+==Rationale==
+This proposal defines a NeoFS storage schema, building on experiments (e.g., nspcc-dev/neo-go#3496)
+and future enhancements (#1526).
+
+== References ==
+* [https://github.com/neo-project/neo/issues/3463 Neo Issue#3463]
+* [https://github.com/neo-project/neo/issues/2373 Neo Issue#2373]
+* [https://github.com/neo-project/neo/issues/1526 Neo Issue#1526]
+* [https://github.com/nspcc-dev/neo-go/issues/3496 NeoGo Issue#3496]
+* [https://github.com/nspcc-dev/neofs-api/issues/285 NeoFS API Issue#285]
```
AnnaShaleva: I don't think that this issue is directly related to the proposal. @roman-khimov?

neofs-blocks.mediawiki
```diff
@@ -0,0 +1,115 @@
...
+
+==Rationale==
+This proposal defines a NeoFS storage schema, building on experiments (e.g., nspcc-dev/neo-go#3496)
+and future enhancements (#1526).
+
+== References ==
+* [https://github.com/neo-project/neo/issues/3463 Neo Issue#3463]
+* [https://github.com/neo-project/neo/issues/2373 Neo Issue#2373]
+* [https://github.com/neo-project/neo/issues/1526 Neo Issue#1526]
+* [https://github.com/nspcc-dev/neo-go/issues/3496 NeoGo Issue#3496]
```
AnnaShaleva: These issues should have readable description instead of just `Issue#...`. See https://github.com/neo-project/proposals/blob/master/nep-23.mediawiki#references as an example.

neofs-blocks.mediawiki
```diff
@@ -0,0 +1,115 @@
+<pre>
+  NEP: TBD
+  Title: NeoFS block storage format
+  Author: Ekaterina Pavlova <ekt@nspcc.ru>
+  Type: Informational
```
AnnaShaleva, line 5: To me the type of this standard looks more like `Standard` rather than `Informational` because we directly standardize NeoFS storage schema and building node synchronization mechanism over it. Or may be even `Meta` standard, since it's not directly related to the Neo protocol, ref https://github.com/neo-project/proposals/blob/master/nep-1.mediawiki#nep-types. @roman-khimov?

neofs-blocks.mediawiki
AliceInHunterland: > The same stile can be applied to container attributes.

We have, for now, only one. Would you like a separate table for it?

neofs-blocks.mediawiki
AliceInHunterland: removed

neofs-blocks.mediawiki
AliceInHunterland: added

neofs-blocks.mediawiki
AliceInHunterland: removed

neofs-blocks.mediawiki
AnnaShaleva: It's nice to have things unified, so table also suits even for one attribute. If we won't use table, then we need at least to adjust description wrt https://github.com/neo-project/proposals/pull/194#discussion_r2163367424.

neofs-blocks.mediawiki
roman-khimov: Example can be a proper example with this data if needed. It doesn't imply any of the IDs are relevant for mainnet or whatever else.

neofs-blocks.mediawiki
roman-khimov: Numbers are stringified for attributes in fact, so this doesn't make sense.

neofs-blocks.mediawiki
roman-khimov: Can be referenced.

neofs-blocks.mediawiki
roman-khimov: Yeah, a different story. Even quite old one.

neofs-blocks.mediawiki
roman-khimov: To me it's `Informational` since there is no obligation for N3 protocol to store blocks in NeoFS and be able to synchronize from it. It's for interoperability if anyone wants to fetch blocks this way, but it's far from P2P `getblocks` for example.

AnnaShaleva:
@roman-khimov, let's review one more time.

neofs-blocks.mediawiki
AnnaShaleva: Reverted.

neofs-blocks.mediawiki
```diff
@@ -0,0 +1,97 @@
+<pre>
+  NEP: TBD
+  Title: NeoFS block storage format
+  Author: Ekaterina Pavlova <ekt@nspcc.ru>
+  Type: Standard
```
roman-khimov: https://github.com/neo-project/proposals/pull/194#discussion_r2169136585

neofs-blocks.mediawiki
```diff
@@ -0,0 +1,97 @@
...
+  Status: Draft
+  Created: 2025-04-07
+</pre>
+
+==Abstract==
+
+This proposal outlines the specification for storing Neo blockchain blocks within
+the NeoFS container.
```
roman-khimov: s/the//

neofs-blocks.mediawiki
```diff
@@ -0,0 +1,97 @@
...
+|-
+| BlockTime || Millisecond-precision timestamp when the block was created.
+|-
+| Timestamp || Second-precision timestamp when the block was uploaded.
+|}
+
+These attributes are stored as NeoFS object metadata; the list of attributes MAY be extended.
+The object MAY include additional data after the block.
+
+An example of a [https://rest.fs.neo.org/v1/objects/3RCdP3ZubyKyo8qFeo7EJPryidTZaGCMdUjqFJaaEKBV/by_id/Fu7yQzspvLJwSGJNK64xeeyMdWXtU5B5b1es6KSxUag1 block object]:
```
roman-khimov, line 78: Maybe use `Test cases` section for it?

neofs-blocks.mediawiki
```diff
@@ -0,0 +1,97 @@
...
+  signature: 0443ff20d15952759b5101a7d223d70eb992260fd9ad5aecb404a97448b2ea54bd1ad783b12ccddba0097dd6608b55ccac9215c5715f9589ec6a555542ead6dc00
+[Block Binary Data]
+</pre>
+
+===Synchronization Algorithm===
+
+Blocks stored in NeoFS and those synchronized via P2P both use the standard Neo block
+serialization format, ensuring compatibility across different synchronization methods.
+The NeoFS container serves as verified storage for downloading blocks. Blocks should be
+searched one by one directly by block attribute.
```
roman-khimov: It'd be dog slow with a per-object search, GE searches should be used. It also better be described in terms of NeoFS API (https://github.com/nspcc-dev/neofs-api/blob/master/object/service.proto).

Jim8y (dismissed):

Jim8y:
LGTM. I prefer to approve the general logic as long as its correct and complete, then polish the format later.

neofs-blocks.mediawiki
```diff
@@ -0,0 +1,104 @@
...
+|-
+| Block || String representation of the block index in base 10.
+|-
+| Hash || Hexadecimal string of the block’s hash.
+|-
+| PrevHash || Hexadecimal string of the previous block’s hash.
+|-
+| BlockTime || Millisecond-precision timestamp when the block was created.
+|-
+| Timestamp || Second-precision timestamp when the block was uploaded.
```
Wi1l-B0t: What is the difference between Timestamp and BlockTime?

Maybe it would be better to rename `Timestamp` to `UploadTime`.

neofs-blocks.mediawiki
```diff
@@ -0,0 +1,104 @@
...
+{| class="wikitable"
+! Attribute name !! Attribute value
+|-
+| Block || String representation of the block index in base 10.
+|-
+| Hash || Hexadecimal string of the block’s hash.
+|-
+| PrevHash || Hexadecimal string of the previous block’s hash.
+|-
+| BlockTime || Millisecond-precision timestamp when the block was created.
```
Wi1l-B0t, line 40: `timestamp` is unix timestamp?

neofs-blocks.mediawiki
Wi1l-B0t:
```suggestion
| BlockTime || Millisecond-precision unix timestamp when the block was created.
|-
| Timestamp || Second-precision unix timestamp when the block was uploaded.
```

`timestamp` is unix timestamp ?

roman-khimov: `Timestamp` just follows well-known NeoFS object attribute:
https://github.com/nspcc-dev/neofs-api/blob/b7548cfb16426256f0a92c6d70a3ec794a6facb0/object/types.proto#L245
(maybe it's worth mentioning this in the document)

AnnaShaleva: Yes, fixed.

Wi1l-B0t (approved):

AnnaShaleva:
@neo-project/core let's review this proposal one more time.

shargon (commented):
- `TimeStamp` for upload time seems very similar to BlockTime, maybe use `UploadedTime` or `StoredAt` instead could be more clear
- It's possible to recover a block range 1-1000 ? without 1000 requests?

roman-khimov:
> TimeStamp for upload time seems very similar to BlockTime, maybe use UploadedTime or StoredAt instead could be more clear

Impossible. It's a part of NeoFS protocol since like forever: https://github.com/nspcc-dev/neofs-api/blame/2fb0e588670ecbaf7b40fe105d1e84cdf91de7e3/object/types.proto#L251

> It's possible to recover a block range 1-1000 ? without 1000 requests?

1 SEARCH + 1000 GETs.

PR #197, opened 2025-05-22
[Standard] Decoupling Council Rank From GAS Distribution 

lllwvlvwlll:
The Neo platform is a proven innovator in many areas in the blockchain space. One particular differentiator is the
unique tokenomic model which rewards NEO holders for participating in our delegated governance process without a lock period. In our current implementation, we only distribute notable reward to voters who support active council members on the voting epoch. GAS is uniformly distributed to each council member's voting bin, then uniformly distributed to the voters in the bin. This process results in a GAS distribution to voters that is inversely proportional to the governance rank of council members, with marginal GAS being distributed to those who vote for candidates without a council seat.

In the field, we have experienced some issues with this distribution strategy. In particular, council "stickiness" has become a significant problem and has led to complacency within platform governance to the point where we increasingly risk being unable to reach the signature threshold for approving protocol level changes without direct Neo involvement.

This proposal will specifically target enhancements to the GAS distribution mechanism to align governance engagement and community participation motivations. The initial specification is a `minimalist` approach and is intended to initiate the discussion. It needs to be evaluated with equivalent scrutiny to any other approaches outlined in the following discussion.

Enhancements considering other areas of the governance process including consensus node succession and voting mechanics should be modularly handled in other proposals.


ethArek:
This is what can make NEO great again! I'm 100% in favor of this idea.

snakeyrob:
Had a couple discussion lately where I expressed a similar opinion as the proposal. It's definitely needed to be able to vote for candidates without the fear of losing $GAS. I understand there is a danger that people can be unmotivated to check whether they are voting for a candidate that is actually still active. I wonder how you view the issue of making sure that people check their votes every now and then.

Otherwise I'm all for this change as it's a definite improvement over the current model.

lllwvlvwlll:
@snakeyrob Thats an incredibly complicated topic that this proposal doesn't address. There have been some discussions and attempts to define a standard process for council members to yield updates to the ecosystem publically. Most recently, NNT were the only ones to follow the process and some council members weren't even made aware that it should be done (since its a permissionless platform so we cant even get confirmation that communications were received). From a product perspective, engagement tools like the wallets (in the case of COZ) can work harder here. 

Ultimately, uniformed voting is a major problem in democratic systems and will probably continue to be one. One of the hopes of this proposal is to remove the incentive for uniformed voters to behave as an effective "Public Action Committee" against the informed voting population who may have greater vote distribution. By removing the protocol-level incentive to participate in a PAC, the vote distribution for uniformed voters widens across candidates. This provides a better opportunity to get a signal from informed voters amongst the noise.

NeoRedPill101:
> Had a couple discussion lately where I expressed a similar opinion as the proposal. It's definitely needed to be able to vote for candidates without the fear of losing $GAS. I understand there is a danger that people can be unmotivated to check whether they are voting for a candidate that is actually still active. I wonder how you view the issue of making sure that people check their votes every now and then.
> 
> Otherwise I'm all for this change as it's a definite improvement over the current model.

What is with your proposal?... We think the pool idea to stake Neo and earn gas (with depending how much memetoken you hold) is a very good addition to this. Maybe it is possible to make something like this too in a opinion if someone want to make this or want to stake in normal governance 🤗

snakeyrob:
> What is with your proposal?... We think the pool idea to stake Neo and earn gas (with depending how much memetoken you hold) is a very good addition to this. Maybe it is possible to make something like this too in a opinion if someone want to make this or want to stake in normal governance 🤗

I'm sorry I don't really understand what you mean, I'm not talking about our Grantshares proposal, that's not related to this topic.

neocactus3:
I am eager to see something like this move forward. Thank you Centre Point council members for organizing. I'm not a huge Elastos fan, but they do have an interesting model for voting that encourages council members to stay active and involved, maybe worth checking out.

neocactus3:
What would ensure that a council member or vote recipient was maintaining a healthy consensus node?

lock9:
The problems described in this NEP are real, but the proposed solution is inadequate and likely to worsen the situation.

- It is already difficult to get users to engage actively with the platform. Most only adjust their votes when a council candidate drops in rank or when rewards decrease significantly. This proposal removes this 'incentive to interact'.
- Distributing rewards equally regardless of candidate performance promotes a “vote once, earn forever” mindset and eliminates the need for users to monitor or participate in governance.

This NEP effectively removes Neo users' last meaningful interaction with the platform, instead of the opposite.

neocactus3:
I agree that folks would “set it and forget it” even harder if all votes delivered equal gas. Ontology and elastos use voting epochs to incentivize re voting and interaction. Could be a time based vote that needs to be reactivated. Council members should actively communicate what they are actually doing to attract votes, whether that is sharing a portion of gas, creating a community fund, managing a wallet, bringing in x users…. Currently I have no idea what Nash is doing with all that gas, certainly not much for Neo.

lock9:
>Could be a time based vote that needs to be reactivated. 

@neocactus3 I just proposed something like it. Check #198

snakeyrob:
> > Could be a time based vote that needs to be reactivated.
> 
> @neocactus3 I just proposed something like it. Check #198

That proposal doesn't include anything regarding voting activity in my opinion. Also the only voting activity you see now is to make sure you keep getting GAS, so I'm not so sure if losing that is such a big deal.

While a mechanism to keep engaging in voting would be great.

In that sense timed votes that need a refresh every x days or you'd gradually lose rewards would make much more sense, but that's not a simple change.

deanragnarok:
Firstly, I want to say that I’m really happy this topic is now being discussed. It’s something we’ve been concerned about for a long time, and I want to encourage everyone to contribute to the conversation — even if it’s just giving a thumbs-up or thumbs-down to ideas or comments. All the energy helps.

I think we have a few different issues we need to address here:

1. We need to shift the incentivization away from voting purely for the best personal economic return and toward voting for what’s best for the health of the Neo ecosystem.
2. We want to encourage ongoing voter engagement. While “set and forget” is a catchy marketing angle, it’s not healthy for the long-term sustainability of the system.
3. We need council members to be active and accountable to their voter base. Just as we want voters to stay engaged, council members must also participate actively in the ecosystem.

@lllwvlvwlll's  initial proposal is a great starting point for solving the first issue — flattening GAS rewards helps redirect voter behavior away from economic maximization and toward other concerns, hopefully ecosystem stewardship.

@lock9's  proposal when combined with @lllwvlvwlll's, helps address the second point — requiring re-voting every 60–90 days can disrupt passive, long-term voting patterns and encourage voter re-engagement. But this only works if we flatten GAS distribution. Otherwise, the economic incentive to vote for the most profitable node still overrides everything else.

Additionally, even with the re-voting requirement, I can still envision voters casting blind votes just to meet the requirement, or automated services emerging to re-cast votes for them every cycle. Or, bNEO may just evolve into a token that just handles re-voting automatically. In these cases, we end up essentially back where we started.

So flatting GAS helps in part, re-voting can help in part — and both are improvements to the current situation — but I really think the piece that will help it all come together is solving the third issue: council member accountability. 

While we can ask council members to do things like provide reports, participate in discussions, and contribute to ecosystem growth through various activities — all of which help demonstrate that they belong on the council and are actively contributing — I believe we can and should still explore protocol-level economic mechanisms that enforce accountability and ensure they remain active and responsive to those who vote for them.

There are a large number of people who are simply voting or claiming their rewards without paying any attention to what the ecosystem is doing or what the council members are contributing. For this segment — which I believe represents a significant portion of voters — it doesn’t matter whether council members are providing reports or not. The only thing they care about is whether they’re making money. So unless council accountability is tied to something economic at the protocol level, we’ll never meaningfully influence that group’s behavior. There needs to be some sort of pain or inconvenience  to the voter if the council member is not engaged.

Here are a few ideas to address this, not as endorsements, but as potential jumping-off points. Hopefully someone smarter than me can come along and suggest more concrete solutions:

- Coupling voter GAS claims to council member activity: For example, if Neo News Today is a council member, maybe our voters do not get GAS until we have claimed ours. This forces engagement from the council and encourages voters to think more carefully about who they’re supporting.
- Automatic ejection due to inactivity: If a council member fails to claim GAS or sign a transaction within a set period, their seat becomes vacant. That could also invalidate GAS rewards for voters during that epoch.
- Decay curve on voter GAS rewards: If a council member is inactive, perhaps the amount of claimable GAS for their voters reduces by, say, 10% per day over the epoch.
- Time-lock on GAS claims: If a council member is inactive, voter GAS claiming is delayed — e.g., a 7-day lock, to add friction and encourage reconsideration of voting behavior.
- Split GAS reward logic: Perhaps 50% of voter GAS is always claimable, but the other 50% is unlocked only if the council member is active (i.e., signs a transaction or votes on-chain).

Of course, with many of these ideas, there’s the risk of council members automating behavior to simply appear active the same way I'm concerned regular NEO holders would (e.g., with scripts to auto-claim GAS). I don’t have any good suggestions for that yet, but I think it’s something we need to keep in mind.

The main point is that we should not overlook the need for on-chain accountability and engagement from council members, especially if we’re asking the same of voters. None of these ideas are silver bullets, but hopefully they spur some ideas more concrete than my own.

We can’t have council members sitting idly by and still reaping rewards. There needs to be provable activity — ideally embedded directly into the protocol — so that “set and forget” isn’t an option for them either.

neocactus3:
I agree that a combination of flattening gas rewards and time based voting would solve part of the problem. What if distribution moved from:

GAS generated on each block is distributed as follows (from lllwvlvwlll's commit)
* 10% - Uniform distribution to NEO token holders
* 10% - Uniform distribution to council members
* 80% - "Successful Voters"
** 40% uniform distribution to each council(14) member voting pool
** 40% uniform distribution to each consensus(7) member voting pool

to 

80% uniform distribution to voting neo.
10% uniform distribution to council members.
5% rewarded to council members based on participation vote held once a quarter. Neo holders vote on progress, ecosystem development, engagement etc... to reward the most active council members. Could be a summit or online presentation by each council member. Certain metrics of participation need to be hit to qualify throughout the quarter. Could be a juicy gas pot for smaller council members that are very active and don't receive as many votes. 
5% to Proposals put forth by council members in a Dao or something. Maybe to Grant Shares? Would encourage active council members to direct funds towards interesting projects and ecosystem development.

lock9:
>Additionally, even with the re-voting requirement, I can still envision voters casting blind votes just to meet the requirement, 

They won't cast blind votes if this affects their GAS rewards. That's why there should be a difference if you vote for candidate A or B.

>or automated services emerging to re-cast votes for them every cycle.

The only way to solve this is by changing Neo Burger or penalizing contract votes (reducing rewards or vote weight). 

>The only thing they care about is whether they’re making money. So unless council accountability is tied to something economic at the protocol level, we’ll never meaningfully influence that group’s behavior. There needs to be some sort of pain or inconvenience to the voter if the council member is not engaged.
>The main point is that we should not overlook the need for on-chain accountability and engagement from council members, especially if we’re asking the same of voters. None of these ideas are silver bullets, but hopefully they spur some ideas more concrete than my own.

Anything based on off-chain governance participation is unfeasible and cannot be enforced. A more creative on-chain mechanism is needed, such as allowing votes for multiple candidates, blocking consecutive votes for the same candidate, or implementing a 'vote snapshot' system. In this model, users can change votes freely during a period, but only the snapshot vote counts. That way, users could vote for the next cycle's council this month without affecting ongoing rewards.

This NEP raises valid concerns, but flattening the GAS distribution will make voting less impactful. I strongly recommend against it.

Only a "Neo Burger with governance steroids" could introduce off-chain accountability tied to node election.

deanragnarok:
> This NEP raises valid concerns, but flattening the GAS distribution will make voting less impactful. I strongly recommend against it.

Flattening the GAS distribution is about making it possible for voters to support someone outside the top 21 without being penalized for backing an up-and-coming council member before they reach that threshold. To me, GAS rewards should be a reward for participating in governance — not a reward for choosing a so-called “correct” council member, which is effectively how the current model operates. Without solving that issue it's impossible to change the status-quo. Anyone who has tried to join the council without a whale backing them has encountered this insurmountable problem.

As long as GAS rewards are tied to council status and rank, there will always be an economic incentive to vote for the member who maximizes personal returns, not necessarily the one who will be the most engaged or beneficial to the ecosystem. It means there is almost no fear for current members to be voted out and no feasible way for new members to be voted in.

I am totally open to alternatives to a flat GAS distribution, but as it stands, this is the simplest proposal thus far to address the issue of enabling people to vote for candidates outside the top 21 without taking an unreasonable economic hit, and shifting the incentive away from just choosing whichever node yields the most GAS. I have other ideas (such as delegated votes that only come into effect when the delegate candidate achieves enough votes to join the council) but they are all more complicated than the current proposal.

> They won't cast blind votes if this affects their GAS rewards. That's why there should be a difference if you vote for candidate A or B.

If we do this right, there should still be a difference between voting for candidate A and candidate B - as I said earlier there needs to be economic pain if the council member is not engaged, and that would be a key difference. But I think we can find ways to link that to other economic mechanics than the yield rate, and maybe towards things like how reliably or often GAS is made available to a voting block.

As kind of a side note not directly related to my comments above, it's always seemed weird to me that NEO holders have to actively claim GAS but council members don't.

lock9:
>Flattening the GAS distribution is about making it possible for voters to support someone outside the top 21 without being penalized for backing an up-and-coming council member before they reach that threshold.

I agree with the problem:  'the need to support someone outside the top 21 without being penalized', but flattening rewards won't solve it.

> To me, GAS rewards should be a reward for participating in governance 
>if we do this right, there should still be a difference between voting for candidate A and candidate B - as I said earlier there needs to be economic pain if the council member is not engaged

This is too abstract to be enforced on-chain. It's also risky to have people who aren't interested in making important decisions.  It will result in the same behavior among NEO holders: nodes participating in the governance just to earn.

If rewards are flattened, the following will happen:
- No more reasons to use a GAS APY calculator
- No more 'let's see if I should change my vote' 
- No more reasons to change their votes
- New voters will vote for 'the first item in the list', without developing any kind of research or building a relationship. At best, they will vote for the one with the logo or colors they like. This will harm developer teams, while benefiting 'marketers'

The first three are critical. They are the favorite activities of NEO holders, and changing them would be extremely damaging to the platform's overall relationship with holders. 

The only solution is Neo Burger or an equivalent dApp / native contract. Since it's a dapp, it could handle different types of rules to balance vote distribution (in a way that the elected nodes reflect the community's intent).

deanragnarok:
In my mind, there are three main groups of voters in the Neo ecosystem.

The first group is made up of those who **don't vote conscientiously — and likely never will**. They’ll either vote purely for maximum profit or just pick any candidate at random for the sake of claiming rewards. I doubt we’re going to change behavior within this group.

The second group consists of people who **would vote conscientiously** — meaning they’d vote for candidates they believe would do the most good for the ecosystem — **but right now, they’re incentivized not to**. The system nudges them toward changing their vote to the candidate that provides the highest GAS yields or the most consistent return. So they don’t vote according to their principles because the protocol actively discourages them from doing so.

Then there’s the third group — those who **do vote conscientiously**, but **only within the limits of the current system**. That means voting only for candidates they like, but only in the top 21, because supporting someone outside of that group is essentially throwing their vote away and sacrificing GAS rewards with no clear upside.

I believe we should be aiming to:

- Shift that middle group toward conscientious voting by removing the incentive not to, and
- Give that last group the freedom and confidence to vote outside the top 21, empowering them to support candidates they believe in without being economically punished for doing so.

When it comes to governance, there will always be an off-chain component. Council members should report on their work, contribute to ecosystem growth, build products, run events, support marketing, evangelize Neo — all of these efforts should be encouraged and factored in when selecting council members. But we all know that’s just one layer of governance. A lot of people will never pay attention to those things. And there will always be council members who campaign for a seat and then disappear.

That’s why I believe we need an on-chain mechanism — not to measure productivity, but to act as a kind of “proof of life”. Something that demonstrates the council member is still there and paying attention to Neo in some capacity. That could be as simple as regularly claiming GAS, or periodically signing an on-chain message or transaction — just something that proves they’re still engaged in _some_ capacity.

It’s not foolproof — but neither is off-chain governance. And if we have this on-chain mechanism in place, we can at least introduce a consequence for voters who support inactive council members: if that council member fails to meet the basic on-chain criteria, their voting base receives reduced or delayed GAS rewards.

>If rewards are flattened, the following will happen:

>- No more reasons to use a GAS APY calculator
>- No more 'let's see if I should change my vote'
>- No more reasons to change their votes

When I think about all the teams we’ve lost over the years — or the teams we still have — who could have done/can do so much more with the resources that come with being a council member — I don’t look at the GAS APY calculators and feel like that would be any great loss. If redirecting rewards enables more teams to build, grow, and contribute much more interesting things to do in the ecosystem than play with a GAS APY calculator, that seems like a worthwhile trade-off. And I say that as someone who runs one of the most popular GAS APY calculators which would be rendered useless.

I'm not convinced by the second and third point either, especially if we have the re-vote and "proof of life" mechanisms. You could make the argument that there would be more reasons to check on your vote than there is now.

>New voters will vote for 'the first item in the list', without developing any kind of research or building a relationship. At best, they will vote for the one with the logo or colors they like. This will harm developer teams, while benefiting 'marketers'

Lastly, I don’t see any meaningful difference between voters picking the first name on the list and voters today who just support the last node with the best returns. Or those who just let NeoBurger do it for them. Functionally, it’s the same behavior — and under this new system we’re discussing, it’s arguably less damaging, because at least those votes will expire every 60 to 90 days and can be refreshed or re-evaluated.

I know I’ve said a lot already today, so I should step back and give some space for other community members to chime in. I’m totally open to changing my position — I’d actually love to be convinced by a better idea — but as of now, I haven’t been convinced of the risks enough to move off it or seen an alternate proposal that addresses these issues more effectively than what we’ve outlined here so far.

Jim8y:
> The problems described in this NEP are real, but the proposed solution is inadequate and likely to worsen the situation.
> 
> * It is already difficult to get users to engage actively with the platform. Most only adjust their votes when a council candidate drops in rank or when rewards decrease significantly. This proposal removes this 'incentive to interact'.
> * Distributing rewards equally regardless of candidate performance promotes a “vote once, earn forever” mindset and eliminates the need for users to monitor or participate in governance.
> 
> This NEP effectively removes Neo users' last meaningful interaction with the platform, instead of the opposite.

In general i agree with this proposal, but in the meanwhile I sort of sharing this concern. reputation system should be involved at least, people should know the behavior of a node, and know who they are voting for....


Or voting for rules,

such as 

1. vote for only active node.
2. vote for top 21.
3. only vote for consensus node.
4. vote for A, if not active, then vote B.
5. vote for best profit.

alike.

NeoDashboard:
This proposal is nice and I think for example removing the `10% - Uniform distribution to NEO token holders` is a good idea.
However I think from a user perspective we need some more information to be able to have an opinion on this. The proposal is talking about `council stickiness` but is It due to NGD/NeoBurger votes? If It is because they want to see this top21 no matter what change is done they can just use more NEO for voting to keep the 21 nodes they want and nothing will be solved?
At first look I would be more happy to see a solution like what was thought initially with NoBug as It could allow to get a lot of NEO liquid to be used in DeFi and could solve the community engagement by governance with NoBug token instead of the `vote and forget` thing most people seems to be worried of. Could be nice to hear about NeoBurger devs if something blocked them to finish this intiative.
Curious to see everyone point of view as It's an important topic but also a difficult one because It's touching NEO governance which is a core component

DylanNNT:
> > The problems described in this NEP are real, but the proposed solution is inadequate and likely to worsen the situation.
> > 
> > * It is already difficult to get users to engage actively with the platform. Most only adjust their votes when a council candidate drops in rank or when rewards decrease significantly. This proposal removes this 'incentive to interact'.
> > * Distributing rewards equally regardless of candidate performance promotes a “vote once, earn forever” mindset and eliminates the need for users to monitor or participate in governance.
> > 
> > This NEP effectively removes Neo users' last meaningful interaction with the platform, instead of the opposite.
> 
> In general i agree with this proposal, but in the meanwhile I sort of sharing this concern. reputation system should be involved at least, people should know the behavior of a node, and know who they are voting for....
> 
> Or voting for rules,
> 
> such as
> 
> 1. vote for only active node.
> 2. vote for top 21.
> 3. only vote for consensus node.
> 4. vote for A, if not active, then vote B.
> 5. vote for best profit.
> 
> alike.

If there should be a reputation system involved, it should likely include two things: node uptime and contributions to the ecosystem.

One is quantitative, and the other is qualitative, but they're both equally important when someone is examining their best options of who to vote for.

I'd really like to see a) NEO voters get GAS for voting for someone outside of the Top 21, and b) Council members to prove their worth.

I haven't really put much more thought to paper yet, so at the moment, I don't have much more to add to my suggestions. Maybe the inspiration will come over the weekend.

vncoelho:
In the past we had discussed on chain statistics of CNs.

I think that publishing CNs stats,agreed during block generation, is a first step to have a public view that can be the basis of a true certified reputation system.

deanragnarok:
> However I think from a user perspective we need some more information to be able to have an opinion on this. The proposal is talking about council stickiness but is It due to NGD/NeoBurger votes? If It is because they want to see this top21 no matter what change is done they can just use more NEO for voting to keep the 21 nodes they want and nothing will be solved?

I certainly can’t speak on behalf of anyone else, but I can say a few things.

First, I don't think it's NGD/Neo Foundation votes that are keeping council members in the top 21 at this point. For the members they support, I _think_ they receive around 100,000 to 150,000 NEO in vote support — whereas the lowest-ranked council member currently has a little over 700,000-ish NEO in votes. So even if you removed the NF support the council would remain largely unchanged. (You can look at Neo SPCC's current status as a reverse example of this - they lost the support of a whale, but the remaining votes which I believe is the NF support, does not keep them in the council). Second, based on conversations I've had there is definitely interest on Neo’s side in promoting more diversity and activity within the Neo Council.

So as far as NeoBurger voting goes, I don’t think it’s the cause of new candidates being unable to join the Council. If anything, it’s more a symptom of the system’s current structure than a root cause of the problem.

lock9:
>First, I don't think it's NGD/Neo Foundation votes that are keeping council members in the top 21 at this point. For the members they support, I think they receive around 100,000 to 150,000 NEO in vote support — whereas the lowest-ranked council member currently has a little over 700,000-ish NEO in votes. 

DHF said that NF holds at least 25M NEO. Recently, they also deposited one million Neo into the bond address. The same is true for Neo Burger: it has almost 2 million Neo deposited. That's more than all current votes combined. 

>Second, based on conversations I've had there is definitely interest on Neo’s side in promoting more diversity and activity within the Neo Council.

It seems naive to think that NF/NGD/Neo Burger can't intervene. It looks more plausible that they don't want to. 

Decoupling the gas distribution from the top 21 would cause votes to change, but that alone is dangerous and insufficient. What would happen if four of the top seven elected nodes aren't online? Afaik, it would lead to the network becoming permanently bricked (if these nodes never become online again). 

Lastly, it's worth mentioning again that a reputation system is not feasible without a DAO or equivalent. You need NEOs to be elected, and that should not change. Changes to Neo Burger are the only thing that can solve the problems mentioned in this proposal, with minimum risk.

roman-khimov:
Many of these things were previously discussed in https://github.com/neo-project/neo/issues/1617. I'm not commenting on the economic part, it's hard to say what's more appropriate there, but my concerns wrt any changes in this area are:
 * consensus stability
 * implementation complexity
 * contract voting
 * the change itself

1. https://github.com/neo-project/neo/issues/2203 is still relevant and current scheme gives us more stable list of CNs. Stable list of CNs is good.
2. Not all schemes are easy to implement, even the current one we have required some time to make it good enough with adequate overhead.
3. I strongly disagree with any penalizing of contract votes. An account is an account.
4. Changing the rules of the game is bad on its own, especially given that we had already changed them during the switch from Legacy to N3.

Regards, node 22 owner.

neocactus3:
The current system does very little to support community accountability and contributions. Is it stable? Yes. Is it good for the future of Neo, doesn’t seem like it. Give Neo voters power to vote in new projects and we will see a much more thriving ecosystems this desperately needs to change.

deanragnarok:
> Many of these things were previously discussed in [neo-project/neo#1617](https://github.com/neo-project/neo/issues/1617). 

[This comment](https://github.com/neo-project/neo/issues/1617#issuecomment-667563865) from @Elean0rZ on that issue was prophetic:

>The result would be a system in which, over time, the committee becomes more and more entrenched. Even if other candidate nodes provide amazing services, the financial incentive would still be for voters to support the committee. In effect, it becomes a tacit agreement among voters: We'll all only vote for these 21 nodes, so everyone 'wins'. Well, everyone except the network, since there's no incentive for nodes to compete and improve. Over time, this means that committee nodes can become 'lazy' regarding the services they provide. Meanwhile, non-committee nodes would slowly be discouraged from even trying to offer superior services, since their efforts would rarely be rewarded with votes. And voters would 'set it and forget it', knowing that there was no reason for them to remain actively involved in a process that isn't really a 'process' so much as a completed decision.

It's a valid concern that we need to maintain a certain level of stability within the consensus nodes, and the current system does allow for that. That was also my recollection of why the system we have now was originally chosen. However, given the other concerns outlined here, it’s clear we need to find a better solution.

A few days ago, I kind of broke down the different issues that were emerging here — and I think we might want to add one more to that list. To recap:
1. The issue we’re specifically trying to address here is: How do we make it more practical for people to vote for any node they believe adds value to the network? In other words, how do we shift voting away from purely seeking the best personal economic return, and toward what’s best for the health of the Neo ecosystem? Is flattening the GAS distribution the best way? If not, what other proposals are out there?
2. How do we prevent “set-and-forget” voting behavior, so that voters stay engaged in the process and decisions remain dynamic?
3. How do we ensure that any candidate who may be elected is actually ready and prepared to operate a consensus node? How can we be confident they’re capable of providing reliable service to the network?
4. And finally, how do we keep council members active and accountable to their voter base? How do we ensure they continue contributing meaningfully to the ecosystem? My sense is that addressing this last issue — council engagement — may require a combination of both on-chain and off-chain solutions.

Maybe it’s best to scope each of these concerns as separate issues here on GitHub? I do think they deserve their own focused conversations. Each one likely has several paths forward, and I’m looking forward to seeing some proposals take shape.

It might be hard, but most things worth achieving are on the other side of hard.

DylanNNT:
> As kind of a side note not directly related to my comments above, it's always seemed weird to me that NEO holders have to actively claim GAS but council members don't.

This is probably where we can resolve Council Member accountability and reduce inactivity. Many PoS networks have slashing and jailing procedures for inactive validator nodes. Don't participate on the network? Income is burnt, or they simply lose their position as a validator. A handful of these "participation requirements" is simply claiming their income within "x" timeframe.

An issue I foresee with a slashing/jailing solution is that Neo doesn't have a deep enough roster of potential Council member replacements to swap out with the defunct and jailed "validators."

Examples of chains that have jailing and slashing include Polkadot, Tezos, Cosmos, Ethereum, NEAR, and Flow to various degrees. 

Whatever the solution is though, retail/individual NEO voters/wallets **shouldn't** be penalized and forced to participate in on-chain activities every "x" days to earn their GAS rewards. And they especially shouldn't lose their GAS rewards if they're inactive (unless I misinterpreted https://github.com/neo-project/proposals/pull/198). I've lost staking rewards on Cosmos chains because I didn't switch my vote to active validators, which should be incentive enough for individuals to keep up to date with who they're voting and if their validator/Council member is still actively contributing to Neo's governance mechanisms.

rkonfj:
> I'd really like to see a) NEO voters get GAS for voting for someone outside of the Top 21, and b) Council members to prove their worth.

Agree

fritzorama:
Thanks for starting this important discussion. I’ve been following the concerns about inactive council members and the current incentive structure that unintentionally favors voting for whoever is already in the top 21 — even if they're not contributing meaningfully.

I’d like to propose a **contribution-based activity score system** that can be used to:

- Surface council member **engagement** transparently  
- Encourage **ongoing participation**  
- Deter inactive members from coasting on historical votes  
- Make **voting more meaningful** for community participants  

---

## 🎯 Overview of the Proposal

Council members would be assigned an **Activity Tier**, based on a publicly visible and verifiable score. This score is composed of metrics that reflect their actual contribution to the Neo ecosystem, both on-chain and off-chain.

### 🔑 Key activity categories:

- **Node Health** – Is their node online and active? How consistent is uptime in the last 3–6 months?
- **GitHub Contributions** – Commits, issues, PRs, reviews — visible on public repos
- **Event Participation** – Attendance at meetups, conferences, and hackathons (tracked via NFI/NFT check-ins like those COZ is working on)
- **Media Engagement** – Contributions to ecosystem awareness through articles, tutorials, podcasts, videos, etc.

Each of these would contribute to a total **Contribution Score**, which would define the member's status:

| Status           | Description                                  |
|------------------|----------------------------------------------|
| 🟢 **Highly Active** | Regularly contributes across multiple areas  |
| 🟡 **Active**        | Meets minimum engagement thresholds          |
| 🟠 **Low Activity**  | Limited contributions, may be declining      |
| 🔴 **Inactive**      | Node down, no community/dev engagement       |
| ⚫️ **Unverified**    | No trackable activity                        |

---

## 🖼️ Visual Mockup

To help illustrate the concept, I created a quick mockup of what the voting interface might look like with these activity tags and breakdowns included.

This would help voters make **informed decisions**, not just based on vote count or name recognition, but on transparent participation.

**Mockups below:**

[Voting UI with Activity Tags]
Council Memebers

[Expanded View with Contribution Breakdown]
Council Memebers expanded

---

## 💬 Final Thoughts

This system could:

- Flatten the playing field for new, motivated contributors  
- Increase accountability  
- Reduce apathy in governance  
- Make being on the council a role of **ongoing responsibility**, not just election

neocactus3:
Where does this go from here?

snakeyrob:
What happened to this topic? Was it discussed in the dev meeting? What's the outcome?

PR #198, opened 2025-05-23
[Standard] Discard Unclaimed GAS Rewards After X Blocks of Inactivity

lock9:
This NEP proposes a mechanism to enforce voter participation in the NEO network by requiring users to 'act' (vote/claim gas) every 30-60 days (configurable via network policy). 

GAS rewards accumulated beyond this period become unclaimable, incentivizing continuous interaction with the network. The goals are to increase voter engagement and reduce the total GAS emitted.

This would still keep the votes active but reduce rewards. Removing the votes was considered, but not included, to keep the changes simple.

### DIFF
--- a/nep-99.mediawiki
+++ b/nep-99.mediawiki
@@ -1,0 +1,117 @@
+# NEP: <to be assigned>
+
+**Title**: Discard Unclaimed GAS Rewards After 30-60 Days of Inactivity  
+**Author**: Ricardo Prado  
+**Type**: Standard  
+**Status**: Draft  
+**Created**: 2025-05-22  
+
+## Abstract
+
+This NEP proposes a mechanism to enforce voter participation in the NEO network by requiring users to re-vote every 30-60 days (configurable via network policy). Claimed rewards accumulated beyond this period will be burned, incentivizing continuous interaction with the network. The goals are to increase voter engagement, reduce the total GAS emitted, and ensure that only active participants' votes influence governance.
+
+## Motivation
+
+The current NEO protocol allows users to vote for consensus nodes and accumulate GAS rewards indefinitely without requiring periodic re-engagement. This leads to:
+
+1. **Low voter interaction**: Users may vote once and never revisit their choice, reducing active participation in governance.
+2. **Inefficient GAS emission**: Inactive voters continue to accrue GAS, increasing the total supply without contributing to network activity.
+3. **Stale votes**: Votes from inactive users remain valid, potentially skewing consensus node selection.
+
+By enforcing periodic re-voting, this NEP aims to:
+
+- Encourage active participation in the network.
+- Reduce GAS emissions by burning rewards for inactive voters.
+- Ensure the network reflects the preferences of engaged users.
+
+## Specification
+
+This NEP introduces a new network policy parameter, `MaxVoterInactivityBlocks`, which defines the maximum number of blocks (approximately 30-60 days) after which a voter's accumulated GAS rewards are burned if they do not re-vote. The default value is set to 172,800 blocks (~30 days, assuming 15 seconds per block).
+
+### Changes to `NeoToken.cs`
+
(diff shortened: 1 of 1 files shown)


neocactus3:
Would definitely incentivize interaction which is needed. ☝️Hopefully will be discussed. Something needs to shift.

zlewe:
There's still an issue about "delegated" votes. i.e. bneo alike service that aggregate votes from small voters and performs routine activity to satisfy the requirements.

snakeyrob:
To me this solution does nothing to increase engagement other than trying to get some txes on the network. I don't see a real benefits as it's what moest are doing anyway and doesn't give any other reason to participate in voting. For small holders it can make sense to delay gas claims until it amounts to something usable.

lock9:
>There's still an issue about "delegated" votes. i.e. bneo alike service that aggregate votes from small voters and performs routine activity to satisfy the requirements.

Yes, I agree, but it should be a separate NEP. IMO, the solutions are either penalizing smart contract votes (reducing rewards or voting power) or finding some way to update Neo Burger. The second option seems ideal because it's an existing 'problem', and people won't stop using it. 

>To me this solution does nothing to increase engagement other than trying to get some txes on the network. I don't see a real benefits as it's what moest are doing anyway and doesn't give any other reason to participate in voting. For small holders it can make sense to delay gas claims until it amounts to something usable.

The main goal is to reduce gas emissions by reducing the reward for those who aren't 'active'. But you are right about smaller holders. In addition to the activity date, it could have a minimum distributed amount, ensuring anyone could wait long enough to make a transfer or equivalent.

snakeyrob:
> Yes, I agree, but it should be a separate NEP. IMO, the solutions are either penalizing smart contract votes (reducing rewards or voting power) or finding some way to update Neo Burger. The second option seems ideal because it's an existing 'problem', and people won't stop using it.

The thing is that we would still need bNEO for DEFI activity due to the design of neo not being divisible.
Tylers proposal at least removes the advantage bNEO has in getting voting rewards only leaving the collective claim advantage, however completely removing bNEO rewards will introduce some difficulties for a platform like Flamingo.

The only way to really mitigate such things would be to make Neo divisible so nobody would have a reason to use bNEO after the voting reward changes.

lock9:
>The thing is that we would still need bNEO for DEFI activity due to the design of neo not being divisible.
Tylers proposal at least removes the advantage bNEO has in getting voting rewards only leaving the collective claim advantage, however completely removing bNEO rewards will introduce some difficulties for a platform like Flamingo.

>The only way to really mitigate such things would be to make Neo divisible so nobody would have a reason to use bNEO after the voting reward changes.

The proposals aren't related, and one doesn't (shouldn't) affect the other.

snakeyrob:
> The proposals aren't related, and one doesn't (shouldn't) affect the other.

Any change can affect any other proposal. But as it stands as written here ignoring other proposals, I'd be opposed as I don't see anything it's supposed to fix.

lock9:
>Any change can affect any other proposal. But as it stands as written here ignoring other proposals, I'd be opposed as I don't see anything it's supposed to fix.

They aren't related. They don't affect each other, and have different purposes. This one is almost exclusively aimed at reducing GAS rewards to inactive voters. Whether they get more or less will make no difference.  

It doesn't include a solution to the problems mentioned in #197 . These NEPs are about different topics.

snakeyrob:
> > Any change can affect any other proposal. But as it stands as written here ignoring other proposals, I'd be opposed as I don't see anything it's supposed to fix.
> 
> They aren't related. They don't affect each other, and have different purposes. This one is almost exclusively aimed at reducing GAS rewards to inactive voters. Whether they get more or less will make no difference.
> 
> It doesn't include a solution to the problems mentioned in #197 . These NEPs are about different topics.

Like I said, I'm ignoring it and don't see the purpose of your proposal.

lock9:
>Like I said, I'm ignoring it and don't see the purpose of your proposal.

I'm not sure if I understand your point. The author suggested creating different NEPs for different topics:
Screenshot 2025-05-23 at 04 54 43

This proposal aims to decrease GAS emissions by reducing rewards for those who aren't 'active'.

snakeyrob:
> This proposal aims to decrease GAS emissions by reducing rewards for those who aren't 'active'.

I simply say I don't see the purpose in this. Imo it doesn't fix anything and just makes things unnecessaryly confusing.

Imo If you want to fix inactive behavior you should search for activity in voting not in claiming as claiming activity doesn't bring anything useful to Neo.

lock9:
>Imo If you want to fix inactive behavior you should search for activity in voting not in claiming as claiming activity doesn't bring anything useful to Neo.

It's not just about activity but also about reducing GAS inflation. But that's okay; I don't think we have to agree.

PR #204, opened 2025-08-30
[Standard] Fixed/Dynamic Gas Contract Method Pricing

cschuchardt88:
Now contract owners, such as a exchange may want a way to get a little gas for sending funds to external wallets. This would allow the contract to have one or several methods to have a fixed price, taking a bit of GAS from transactioner for calling the a method (Much like an opcode price works). For example, a LP token maybe want you to pay gas for withdrawing funds. Reason most LP/Fund contracts do this to pay for the LP contract pool's transaction fees for adjusting prices and other things.

### DIFF
--- a/nep-33.mediawiki
+++ b/nep-33.mediawiki
@@ -37,9 +37,9 @@ then will read the <code>JSON</code> string to see if the method exists. If the 
    - If the amount does exist. Than that unit amount will than be deducted.
    - Amount can't be negative. If is negative than this entry is ignored.
 1. Receiver Address
-   - If receiver doesn't exist. Than the contract itself must have the <code>onNEP17Payment</code> method, must support NEP17 and accept GAS.
-   - If reciever does exist. Than the reciever address will be used. The reciever address can be a NEP17 contract. However the contract must accept GAS.
-   - If an error occurs for any reason the gas is then burned and added as an VM fee as used GAS.
+   - If receiver doesn't exist. Then the contract itself must have the <code>onNEP17Payment</code> method, must support NEP17 and accept GAS.
+   - If reciever does exist. Then the reciever address will be used. The reciever address can be a NEP17 contract. However that contract must accept GAS.
+   - If an error occurs for any reason the gas is than burned.
    - If <code>0x0000000000000000000000000000000000000000</code> is used the amount is burnned no matter what.
 
 ==JSON==
--- a/nep-33.mediawiki
+++ b/nep-33.mediawiki
@@ -33,7 +33,7 @@ then will read the <code>JSON</code> string to see if the method exists. If the 
    - Method can't start with underscores <code>_</code>. This limits malicious activity. For example if you were to require a fee for <code>_initialize</code>.
    - Methods <code>verify</code> and <code>onNEP17Payment</code> will be ignored.
 1. Amount
-   - If the amount doesn't exist. Then the small unit of GAS is used (_Datoshi_). _For example: GAS's small unit in Datoshi <code>0.00000001</code>_
+   - If the amount doesn't exist. Then the smallest unit of GAS is used (_Datoshi_). _For example: GAS's small unit in Datoshi <code>0.00000001</code>_
    - If the amount does exist. Than that unit amount will than be deducted.
    - Amount can't be negative. If is negative than this entry is ignored.
 1. Receiver Address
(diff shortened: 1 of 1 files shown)


shargon:
I think that it's a good idea

vncoelho:
This is a good feature,@cschuchardt88 .
Quite useful!

Maybe we can have limits set by the storage of the contract with some fixed prefix that contract can update under its preference without redeploying.
So, prefixes are given for each method that is enabled.

nep-33.mediawiki
```diff
@@ -0,0 +1,98 @@
...
+==Motivation==
+
+Now contract owners, such as a exchange may want a way to get a little gas for sending funds to external wallets. This would allow the contract to have
+one or several methods to have a <code>fixed</code> price, taking a bit of GAS from transactioner for calling the a method (Much like an opcode price works).
+For example, a LP token maybe want you to pay gas for withdrawing funds. Reason most LP/Fund contracts do this to pay for the LP contract pool's 
+transaction fees for adjusting prices and other things.
+
+==Specification==
+
+Contract will have a <code>JSON</code> string in their storage at prefix <code>0xfe</code>. At this storage location the <code>JSON</code> string will
```
roman-khimov, line 24: Contracts use storage the way they like, some may already be using this prefix. Storage scheme is deeply internal for contracts, so we can't dictate anything there. Also, parsing JSON is never fun in runtime, this can be abused to raise contract call cost (in time).

nep-33.mediawiki
```diff
@@ -0,0 +1,98 @@
...
+then will read the <code>JSON</code> string to see if the method exists. If the method does exist than will follow the following.
+
+* Method
+** Method must exist within the contract's manifest.
+** If method doesn't exist within the manifest than entry is ignored.
+** Method can not start with a underscore <code>_</code>. This limits malicious activity. For example if you were to require a fee for <code>_initialize</code>.
+** Methods <code>verify</code>, <code>onNEP11Payment</code> and <code>onNEP17Payment</code> will be ignored.
+* Amount
+** If the amount doesn't exist. Then the smallest unit of GAS is used (Datoshi). For example: GAS's small unit in Datoshi <code>0.00000001</code>
+** If the amount does exist. Than that unit amount will than be deducted.
```
roman-khimov, line 37: Out of where? VM GAS? So it'd be like minting some amount of new GAS out of (previously burned) system fee?

nep-33.mediawiki
```diff
@@ -0,0 +1,98 @@
...
+** Amount can not be negative. If is negative than this entry is ignored.
+* Receiver Address
+** If receiver doesn't exist. Then the contract itself SHOULD have the <code>onNEP17Payment</code> method and SHOULD support NEP17 and accept GAS.
+** If reciever does exist. Then the reciever address will be used. The reciever address can be a NEP17 contract. However that contract SHOULD accept GAS.
+** If <code>0x0000000000000000000000000000000000000000</code> is used the amount is burnned no matter what.
```
roman-khimov, line 42: We have `System.Runtime.Burn.Gas` for this.

roman-khimov (commented):
While we have a similar problem with NeoFS withdrawals (we need some GAS to do the job, so we take a fee), this solution is not convincing to me:
 * it only allows GAS to be used, while contracts may want some different tokens, sending some amount with `onNEP17Payment` callback doesn't have this problem
 * it's not easy to follow transfers this way, also not a problem for `onNEP17Payment`

Effectively, if contract wants your money `onNEP17Payment` works fine already, it's just a part of the contract interface and for users it's the same "do smth good" button in whatever UI they use.

nep-33.mediawiki
cschuchardt88: This will be added in as a vm system fee within transaction, Thats why I chose to use only gas. All this really is an vm fee for using a method. Much like an opcode fee. It just makes `Call` more expensive when your calling a method that has a fee within a contract. We could always add this to manifest for a method and add property fee and receiver.

roman-khimov: This can be arranged in a simpler way then. Just like we have `System.Runtime.BurnGas` we can also have `System.Runtime.MintGas` (or `GrabGas`, or whatever we may call it) that would add a system fee (like `BurnGas`) and mint the amount to calling contract. Nothing else needed, test invocations will work fine.

But there is some potential for abuse there probably. Although it's all limited by system fee that user is already agreed to pay by signing a transaction.

shargon:
Why it was closed?

cschuchardt88:
https://github.com/neo-project/neo/issues/4238 will replace this PR

Issue #137, opened 2021-04-20
Transfer to Transaction NEP-TTT


igormcoelho:
Summary: this NEP called Transfer to Transaction tries to provide a practical manner to redistribute assets during Verification time on transaction, thus allowing implementation of practical mechanisms for "free transactions".

=====

**Transfer to Transaction**

This NEP considers a transaction as a valid (and temporary) token holder (as HASH256 identifier, not usual HASH160). 
It can be done with operation "contract.scheduleTransfer(from, target_tx, value)" that launches a "promiseNotification" (consumed before actual contract invocation).
Contracts can easily extract funds from transaction, for example, in a mint operation. Neo system can also consume funds from Transaction, for example, GAS funds to cover fee for operations. User can also specify a "contract scope" or "group scope" as the temporary holder (instead of global "transaction scope") and funds are consumed according to the scope.

Example 1:
- User transfers its own 0.5 GAS to TxA (on tx header) 
- User calls transfer operation for ContractY NEP17 (on tx invocation) and uses GAS stored on tx for operations

Example 2:
- User "requests" 0.5 GAS from ContractY to TxA (on tx header) 
- User calls transfer operation for ContractY NEP17 (on tx invocation) and uses GAS stored on tx for operations

This "request" operation at ContractY can do some interesting thing such "redistributing" GAS for user operation, in that specific scope, with some simple rule as: 
- gas_request < User_NEP17_Balance_at_ContractY / NEP17_Total_Supply * GAS_Balance_at_ContractY * (1 - RELATIVE_HOLD_TIME)
- gas_request < max_gas_request => (some contract-specific constant tighter than the constraint above)
- RELATIVE_HOLD_TIME = last_balance_change / contract_specific_hold_constant (blocks)
- last_balance_change > XYZ (blocks)

This means that giveaway operation only works after few blocks, and with limited gas supply (limited per block hold time and user shares on that contract).

igormcoelho:
Related to https://github.com/neo-project/neo/issues/1468

erikzhang:
So these two transaction must be included in the same block?

igormcoelho:
The idea is that this is a single and atomic transaction.

One way to implement this is to put some new verification script field on tx header (I'm avoiding this approach)
Another way is to have the script part that emits the `GAS.scheduleTransfer(sender, GetTransactionHash(), value)`  to be embedded in a co-signer.
[...]

PR #208, opened 2025-10-27
Custom Contract Fees Mechanism (GAS only, N4-aligned)

cschuchardt88:
# NEP: Custom Contract Fees Mechanism (GAS only, N4-aligned)

**NEP**: TBD  
**Title:** Custom Contract Fees with ABI Declaration and Runtime Enforcement (GAS)  
**Author:** Chris Schuchardt (revised per community feedback)  
**Type:** Standard  
**Status:** Draft  
**Created:** 2025-10-27 (revised: 2026-06-21)

---

## Abstract

This NEP introduces a standardized mechanism for smart contracts to declare **custom execution fees in GAS** (in addition to standard network gas). These fees are paid by the caller to the contract owner (or designated beneficiary) upon successful invocation. The fee structure is declared in the contract’s ABI using a new `fee` metadata field, supported by updated DevPack tooling and enforced **transparently** by the Neo VM at runtime.

This proposal is aligned with and targeted for **Neo N4**.

---

## Motivation

Many dApps require micro-transactions, subscription models, or usage-based billing directly within smart contracts. While GAS covers network costs, there is no native way to charge contract-level fees in GAS.

**Per reviewer consensus (@shargon, @erikzhang, @ixje, @vncoelho):**
- Restricted to **GAS only** for simplicity, transparency, and to strengthen GAS utility.
- Contracts that use their own tokens do not need this API.
- Dynamic fees must support **internal** (same-contract) calculation.
- Enforcement must be fully transparent to the calling transaction (no extra manual transfers or visible syscalls).

This enables pay-per-use contracts, premium features, decentralized SaaS, and fair compensation for maintainers.

---

### 1. ABI Extension: Fee Declaration

New optional `fee` object added to method definitions in the ABI.

```json
{
  "methods": [
    {
      "name": "queryData",
      "parameters": [],
      "returnType": "String",
      "offset": 123,
      "safe": true,
      "fee": {
        "amount": 100000000,
        "beneficiary": "NM7g6DAeN3Nx8iK53GqY7fpeyqX5bUq5fJ",
        "mode": "fixed" | "dynamic",
        "dynamicScriptHash": "0x..."
      }
    }
  ]
}
```

**Field Details**

| Field                | Type    | Required          | Description |
|----------------------|---------|-------------------|-------------|
| `amount`             | integer | Yes               | Fee in GAS smallest units (8 decimals) |
| `beneficiary`        | string  | No                | Recipient address. Defaults to contract owner (manifest) |
| `mode`               | string  | Yes               | `"fixed"` or `"dynamic"` |
| `dynamicScriptHash`  | string  | No                | External calculator. Omit or match self → **internal** `CalculateFee` on current contract |

### 2. Dynamic Fee Computation

For `mode: "dynamic"`, implement:

```csharp
public BigInteger CalculateFee(string methodName, object[] args);
```

- **Must be safe** (VM enforces gas limit, e.g. 100k gas).
- If `dynamicScriptHash` differs from target contract → VM calls external.
- If omitted or matches target → VM calls **internal** `CalculateFee` on the current contract (full private storage access).

This satisfies the same-contract request from @ixje and @shargon.

### 3. C# Decorators (DevPack)

```csharp
[Fee(Amount = 50000000)]                                    // 0.5 GAS, fixed
public static string QueryData(string key) { ... }

[Fee(Amount = 100000000, Mode = FeeMode.Dynamic)]           // internal dynamic
public static UInt160 GetReport(uint userId) { ... }

[Fee(Amount = 25000000, Mode = FeeMode.Dynamic, Calculator = "0xb2a4...")]
public static string ComplexOp(...) { ... }
```

### 4. Runtime Enforcement (Transparent – No Extra Syscalls)

**Flow (completely transparent to caller script):**

1. Standard `contract.method(args)` invocation.
2. VM reads ABI `fee`.
3. Deducts GAS (fixed or via `CalculateFee`).
4. Transfers to beneficiary.
5. Executes method.
6. On revert → refunds custom fee (network GAS consumed normally).

No manual pre/post transfers required. No visible extra syscalls. Handled internally by VM (N4 scope).

### 5. Owner & Beneficiary

Defaults to manifest owner; per-method override supported.

---

## Security Considerations

- Dynamic calculator gas-capped + safe requirement.
- ABI is part of signed manifest.
- Refund on revert ensures fairness.
- GAS-only design avoids divisibility issues and leverages mature mechanics.

---

## Backwards Compatibility

- No `fee` in ABI → no change.
- Old DevPack → ignores new field.
- Fully N3 compatible; **primary target is Neo N4** (“It’s part of N4”).

---

## Rationale

- **GAS only**: Strong reviewer consensus.
- **Internal dynamic support**: Directly requested and endorsed.
- **Transparent VM enforcement**: No extra burden on transaction scripts.
- Future native DEX / LP scenarios (@vncoelho) left open for later N4+ work.

---

## Test Cases & Implementation

- Fixed / internal-dynamic / external-dynamic paths
- Beneficiary defaults + overrides
- Revert refund semantics
- DevPack + VM integration (N4)


shargon:
In my opinion only gas should be used (at least now) it's easier to compute and attach the gas into the transaction in calculateFeeMethod, and it's good for the strength of the network, it make the gas more useful. Also, neo is not divisible, so I think there is no reason to pay one neo for one smart contract call.

Also I think we don't need to add any extra syscall, it will be transparent to the user who make the transaction.

erikzhang:
> In my opinion only gas should be used

Some projects may have their own tokens that need to be taken into account.

ixje:
I'd like to see a `dynamic` `mode` where `calculateFee` is in the same contract instead of being forced to call an "external" contract. That way it can also be based on internal storage instead of being forced to add public methods to get the data to calculate the fee from the "calculator contract".

shargon:
> > In my opinion only gas should be used
> 
> Some projects may have their own tokens that need to be taken into account.

This will add the requirement of do a transfer before (or after) the execution, so it's not transparent to the transaction script. Maybe if we use gas we can do it transparent.

erikzhang:
I thought about it and realized that if the contract uses its own token, then it doesn’t really need this API, so I agreed to support only GAS.

shargon:
> I'd like to see a `dynamic` `mode` where `calculateFee` is in the same contract instead of being forced to call an "external" contract. That way it can also be based on internal storage instead of being forced to add public methods to get the data to calculate the fee from the "calculator contract".

As @ixje said, if we use the same contract we reduce the configuration, the contract can implement the required method and call to another contract if they want.

vncoelho:
> I thought about it and realized that if the contract uses its own token, then it doesn’t really need this API, so I agreed to support only GAS.

I think that the design should be draw to support that in the future.

If we really move in a way to have a "native DEX", this  can become a possible scenario much easier.

Also, it would be interesting if Contracts create a Liquidity Pool only accessible for such cases, maybe the contract owner could provide liquidity himself. So, the pool is just used for such cases.

nep-33.md
```diff
@@ -0,0 +1,186 @@
...
+---
+
+### 2. **Dynamic Fee Computation Interface**
+
+Contracts declaring `mode: "dynamic"` must reference a fee calculator contract implementing:
+
+```csharp
+public interface IFeeCalculator
+{
+    BigInteger CalculateFee(ByteString method, object[] args);
```
roman-khimov: Methods are defined using NEP-25 JSON, not using C#, see any other NEP defining methods for example.

nep-33.md
```diff
@@ -0,0 +1,186 @@
...
+
+- Called by VM before execution
+- Must be **Safe** (no state changes)
+- Returns fee in smallest unit
+
+---
+
+### 3. **Decorators**
+
+#### C# Attributes
```
roman-khimov: Implementation detail not relevant to the VM/runtime spec.

nep-33.md
```diff
@@ -0,0 +1,186 @@
...
+
+[Fee(
+    Asset = "GAS",
+    Mode = FeeMode.Dynamic,
+    Calculator = "0xb2a4cff31913016155e38e474a2c06d08be296cf"
+)]
+public static UInt160 GetReport(uint userId) { ... }
+```
+
+#### DevPack Export Logic
```
roman-khimov: Same here.

roman-khimov (commented):
1. GAS-only fees can be implemented much easier with https://github.com/neo-project/proposals/pull/204#discussion_r2362170457
2. Generic token fees are just a specific kind of transfer to me. Contracts can call `transfer(someone, me, amount)` and this is guarded with witness scopes, but the mechanism of scopes currently gives permissions for arbitrary changes and that's the thing that can be improved. If we could express a permission to transfer X amount from A to B in scopes I think this all would be solved immediately, contracts would just use regular transfer calls for their needs and it'd not require any ABI changes. Related to #137.

cschuchardt88:
Don't need

shargon:
It's part of N4

cschuchardt88:
@neo-project/core

All reviewer comments addressed.

Please re-read 1st comment or check the `.mediawiki` file

PR #213, opened 2025-12-08
NEP-message-signing

ajara87:
Add message singing proposal

### DIFF
--- a/nep-XX.mediawiki
+++ b/nep-XX.mediawiki
@@ -1,0 +1,164 @@
+<pre>
+  NEP: XX
+  Title: Offline Message Signing and Verification Standard
+  Author: Adrian Fjellberg (atlazor), Alvaro Jara (ajara87)
+  Type: Standard
+  Status: Draft
+  Created: 2025-12-08
+</pre>
+
+==Abstract==
+
+This NEP specifies how the command `sign message` constructs and signs an offline message. It standardizes (1) the exact payload format (including an anti-transaction prefix/suffix and a random salt), (2) the use of the network magic as a domain separator via `GetSignData`, and (3) the minimum set of values that implementations expose so that signatures can be reproduced and verified by other tools.
+
+==Motivation==
+
+Neo users, node operators and external systems need to prove key ownership or authorize off-chain actions by signing messages instead of transactions. Without a standard, each implementation can invent its own payload and signing rules, making verification brittle and opening the door to cross-network replay.
+
+The goals of this NEP are:
+
+* Ensure that “message signatures” cannot be interpreted as valid transactions.
+* Bind signatures to a specific Neo network via the existing `networkMagic`.
+* Make it clear what fields a wallet or CLI should output (payload, keys, signatures, salt) so that other tools can verify them.
+
+==Specification==
+
+===Context and inputs===
+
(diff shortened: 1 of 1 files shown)


nep-XX.mediawiki
```diff
@@ -0,0 +1,164 @@
...
+
+===Context and inputs===
+
+The standard applies to the CLI command:
+
+<pre>
+wallet sign message &lt;message&gt;
+</pre>
+
+A conforming implementation MUST:
```
roman-khimov, line 34: The standard is written in a way that ensures there is a single implementation. NEPs shouldn't specify CLI commands.

ajara87: thank you Roman. I'll rewrite it.

ixje: I'd like to add that it be described in a language agnostic way. I see a lot of C# code in there

nep-XX.mediawiki
```diff
@@ -0,0 +1,164 @@
...
+var buffer = new byte[4 + UInt256.Length];
+BinaryPrimitives.WriteUInt32LittleEndian(buffer, networkMagic);
+hash.Serialize(buffer.AsSpan(sizeof(uint)));
+signData = buffer;
+</pre>
+
+That is:
+
+* 4 bytes: `networkMagic` in little-endian.
+* Followed by 32 bytes: the serialized `hash` (`UInt256`).
```
ixje, line 114: Not sure why UInt256 got involved here. 

Can't we just say something along the lines of
___
The data to be signed using `secp256r1` is a byte array consisting of; `a` + `b` where
 
* `a` is the networkMagic encoded as a 32-bit little-endian unsigned integer
* `b` is the `sha256(sha256(payload))` digest where `payload` is formatted as described in `<section xyz>`
___
All these C# code references from the `neo` project make no sense for other languages and should be irrelevant for a specification.

Instead, I would suggest to add a section with test vectors so implementers can verify their implementations.

ajara87: thank you @ixje  I'm rewriting the proposal. It will be uploaded next weak.

shargon (commented):
My data Idea:

NetworkMagic (`GetSignData already added`) + Ocode.RET + 4 bytes for ValidUntilBlock + 4/8 bytes for nonce + message

This allow:
- Avoid to use the message as script because the first RET and transaction wrapper.
- Refuse a signature if the block was reached.
- Refuse a signature if it's the wrong network.

roman-khimov:
Wallets do no work this way. Wallets sign the thing described in the proposal. It's not new, it just tries to outline things that are known for years, de facto standard.

erikzhang:
I believe the message signature should always be verifiable and should not include a `ValidUntilBlock`.

roman-khimov:
> Refuse a signature if the block was reached.
> Refuse a signature if it's the wrong network.

Abstract message is not a transaction, these things are out of scope to me. It's message and its interpretation that matters wrt this. The main thing this standard tries to do is to avoid signing a valid transaction.

shargon:
> I believe the message signature should always be verifiable and should not include a `ValidUntilBlock`.

But if we include the blockchain protection (using `GetSignData`) we will be incompatible with the wallet implementation also

PR #215, opened 2025-12-26
Draft NEP: GAS-only contract fees via system fee minting

Jim8y:
## Summary
- add Draft NEP in mediawiki format for GAS-only contract fees via System.Runtime.MintGas
- note token fees are out of scope for now
- link to reference implementation in neo-project/neo#4396

## Testing
- not run (doc-only change)

### DIFF
--- a/nep-33.mediawiki
+++ b/nep-33.mediawiki
@@ -1,0 +1,64 @@
+<pre>
+  NEP: TBD
+  Title: GAS-Only Contract Fees via System Fee Minting
+  Author: Chris Schuchardt <cschuchardt88@gmail.com>, Roman Khimov <roman.khimov@gmail.com>
+  Type: Standard
+  Status: Draft
+  Created: 2025-10-27
+</pre>
+
+==Abstract==
+
+This NEP defines a minimal protocol addition that enables GAS-only contract fees without ABI changes. It introduces a syscall that mints GAS to the executing contract while increasing the transaction system fee. Token-based fees are explicitly out of scope for now.
+
+==Motivation==
+
+Contracts often need to charge fees for services such as withdrawals, premium data, or subscription access. Existing approaches either require custom ABI metadata or force users to pre-transfer tokens, both of which add friction and reduce transparency. This NEP keeps fee logic inside contracts while avoiding ABI changes and keeping authorization explicit at the transaction level, but limits the scope to GAS-only fees.
+
+==Specification==
+
+===System.Runtime.MintGas===
+
+Add a new syscall for GAS-only fees:
+
+<pre>
+void System.Runtime.MintGas(BigInteger amount)
+</pre>
+
+Semantics:
+
+* <code>amount</code> MUST be non-negative.
+* If <code>SystemFee + amount</code> exceeds the transaction <code>MaxSystemFee</code>, the VM MUST FAULT.
+* On success, <code>SystemFee</code> is increased by <code>amount</code> and the same <code>amount</code> of GAS is credited to the executing contract (<code>Runtime.ExecutingScriptHash</code>).
+* The effect is equivalent to <code>System.Runtime.BurnGas(amount)</code> plus a native GAS mint to the executing contract.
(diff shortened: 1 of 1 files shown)


Jim8y:
replace https://github.com/neo-project/proposals/pull/208

nep-33.mediawiki
```diff
@@ -0,0 +1,64 @@
...
+  Title: GAS-Only Contract Fees via System Fee Minting
+  Author: Chris Schuchardt <cschuchardt88@gmail.com>, Roman Khimov <roman.khimov@gmail.com>
+  Type: Standard
+  Status: Draft
+  Created: 2025-10-27
+</pre>
+
+==Abstract==
+
+This NEP defines a minimal protocol addition that enables GAS-only contract fees without ABI changes. It introduces a syscall that mints GAS to the executing contract while increasing the transaction system fee. Token-based fees are explicitly out of scope for now.
```
vncoelho, line 12: Good explanation @Jim8y

nep-33.mediawiki
```diff
@@ -0,0 +1,64 @@
...
+Contracts often need to charge fees for services such as withdrawals, premium data, or subscription access. Existing approaches either require custom ABI metadata or force users to pre-transfer tokens, both of which add friction and reduce transparency. This NEP keeps fee logic inside contracts while avoiding ABI changes and keeping authorization explicit at the transaction level, but limits the scope to GAS-only fees.
+
+==Specification==
+
+===System.Runtime.MintGas===
+
+Add a new syscall for GAS-only fees:
+
+<pre>
+void System.Runtime.MintGas(BigInteger amount)
```
vncoelho, line 25: I think that MintGas is not a good name for this.

This should be something like `ChargeGas`

nep-33.mediawiki
```diff
@@ -0,0 +1,64 @@
...
+</pre>
+
+Semantics:
+
+* <code>amount</code> MUST be non-negative.
+* If <code>SystemFee + amount</code> exceeds the transaction <code>MaxSystemFee</code>, the VM MUST FAULT.
+* On success, <code>SystemFee</code> is increased by <code>amount</code> and the same <code>amount</code> of GAS is credited to the executing contract (<code>Runtime.ExecutingScriptHash</code>).
+* The effect is equivalent to <code>System.Runtime.BurnGas(amount)</code> plus a native GAS mint to the executing contract.
```
vncoelho, line 33: So, the workaround would be Burn systemfee as expected?
Then, mint GAS for the executing contract?

vncoelho: If so, I think it would be an acceptable solution for N3 only, not N4.

nep-33.mediawiki
```diff
@@ -0,0 +1,64 @@
...
+
+===Usage Pattern===
+
+* GAS fees: The contract computes the fee and calls <code>System.Runtime.MintGas(fee)</code>.
+
+No ABI changes are required.
+
+==Rationale==
+
+* GAS-only fees can be implemented with a minimal syscall (<code>System.Runtime.MintGas</code>) rather than ABI metadata.
```
vncoelho, line 46: Again, I think this should be `ChargeGas` instead of Mint.

vncoelho (commented):
@Jim8y ,

It looks good.
I did not understand the Semantics very well.

I think that the syscall should be renamed to "ChargeGas" instead of "MintGas".

Regarding the implementation , I think that `NativeContract.GAS.Mint(this, contract.Hash, amount, false).GetAwaiter().GetResult();` is acceptable for N3 but not for N4.
Anyway, it will work...If the majority think that it is good like this, it can be kept like this.
I just think it is a little strange to Mint from NativeContract for this reason.
But, it is a doable way to do it.

When I see this implementation/workaround, I straightly think that we need some another extra feature on N4 to make this a real charge of gas and not a mix of Burn and Mint.

Jim8y:
Closing this for now. The current approach and naming need a clearer direction before continuing the proposal.

PR #218, opened 2026-03-19
Draft: Contract-based Verification Script Standard

Jim8y:
## Summary
- generalize the proposal from a fixed `verify(accountId)` account pattern to a generic contract-based verification script standard
- require a statically parseable `System.Contract.Call` target with constant verifier contract hash, method name, ReadOnly call flags, and argument count
- allow arbitrary public verifier method names, arbitrary parameter counts, and non-empty invocation scripts for signatures or other dynamic witness data

## Review feedback addressed
- removes the fixed `verify()` requirement
- removes the one-argument `accountId` restriction
- documents the static app-call instruction shape needed by tooling
- allows invocation scripts to provide witness data for transaction signing and verification

## Verification
- git diff --check


nep-aa-verification-script.mediawiki
```diff
@@ -0,0 +1,86 @@
...
+* <code>accountId</code>: an opaque <code>ByteString</code> chosen by the Abstract Account system.
+* <code>entryContract</code>: the <code>Hash160</code> of the Abstract Account entry contract.
+* <code>accountAddress</code>: the script hash of the verification script defined by this NEP.
+
+===Verification script===
+
+A compliant Abstract Account verification script MUST be semantically equivalent to:
+
+<pre>
+System.Contract.Call(entryContract, "verify", ReadOnly, [accountId])
```
roman-khimov: 1. `verify()` is NEP-30, is must not be used here.
2. Limiting the name is not required. This should be a generic contract-based verification script standard.
3. Called function can have any number of parameters. Think of `verifyAccount(network, namespace, id, sig)`
4. Specific instructions should be mentioned, it's crucial for interoperability (refs https://pkg.go.dev/github.com/nspcc-dev/neo-go/pkg/smartcontract/scparser#GetAppCallFromContextNonStrict)

nep-aa-verification-script.mediawiki
```diff
@@ -0,0 +1,86 @@
...
+
+* MUST derive the Abstract Account address from the full verification script.
+* MUST NOT derive the address from <code>accountId</code> alone.
+
+===Witness rules===
+
+When a transaction uses an Abstract Account address defined by this NEP as a signer:
+
+* the witness verification script MUST be the deterministic verification script for that Abstract Account;
+* the invocation script SHOULD be empty;
```
roman-khimov: How are you going to sign/verify transactions then?

nep-aa-verification-script.mediawiki
```diff
@@ -0,0 +1,86 @@
...
+The verification script:
+
+* MUST embed <code>accountId</code> as a constant.
+* MUST embed <code>entryContract</code> as a constant.
+* MUST call the method named <code>verify</code> on <code>entryContract</code>.
+* MUST return the Boolean result of that call.
+
+===Address derivation===
+
+The public address of the Abstract Account MUST be the script hash of the verification script defined above.
```
roman-khimov: Addresses are _always_ derived from verification scripts, there is no way around it for anyone, not worth mentioning.

nep-aa-verification-script.mediawiki
```diff
@@ -0,0 +1,86 @@
...
+
+==Abstract==
+
+This NEP defines a deterministic verification script for Neo Abstract Accounts.
+The script embeds an opaque <code>accountId</code> and an Abstract Account entry-contract hash, calls <code>verify(accountId)</code> on that entry contract, and uses the script hash of this verification script as the public address of the Abstract Account.
+
+==Motivation==
+
+Current Abstract Account designs can derive an account address without deploying a new contract for every user.
+That property is useful only if wallets, explorers, SDKs, relayers, and indexers derive the same address from the same inputs.
```
roman-khimov: That's not the problem.

roman-khimov (commented):
I'd happily drop all references to abstract accounts here except for motivation. We need a standard for contract-based verification scripts (for wallets/RPC to process them correctly wrt estimations/signatures), then it can be used to provide any accounts you'd like.

Jim8y:
Refocused this draft around contract-based account verification scripts. Abstract Accounts are now only listed as one use case in the motivation, while the normative sections define the deterministic witness script and address derivation that wallets, RPC tooling, SDKs, explorers, relayers, and indexers can process consistently.

Jim8y:
Updated the draft to address the contract-based verification script feedback. It no longer fixes the method name to `verify`, no longer limits the verifier call to one argument, allows non-empty invocation scripts for witness data such as signatures, and documents the static `System.Contract.Call` shape tooling can parse, including non-strict argument parsing when invocation-supplied values are used.

PR #219, opened 2026-03-19
Draft: Transferable Abstract Account Standard

Jim8y:
## Summary
- add a draft NEP for a transferable Abstract Account mode
- model one single-controller AA as one non-divisible NEP-11 token
- let NEP-11 transfer semantics handle sending and marketplace trading

## Context
This draft is based on the current `neo-abstract-account` address-first model and keeps the transferability scope intentionally narrow.

### DIFF
--- a/nep-aa-transferable-account.mediawiki
+++ b/nep-aa-transferable-account.mediawiki
@@ -1,0 +1,90 @@
+<pre>
+  NEP: <to be assigned>
+  Title: Transferable Abstract Account Standard
+  Author: <to be completed>
+  Type: Standard
+  Status: Draft
+  Created: 2026-03-19
+  Requires: 11
+</pre>
+
+==Abstract==
+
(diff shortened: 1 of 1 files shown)


roman-khimov (commented):
Not sure what's the real use case and whether it's needed at all. An account is an account, contract account plays by contract rules, what does "owner" mean in this context?

erikzhang:
We need an `AccountManagement` native contract.

roman-khimov:
We don't. Contract-based accounts (and I don't mean NEP-30) are available today with regular contracts, there is nothing to do for native contract.

Jim8y:
Closing this for now. The ownership and transferability model is not well defined enough for a standard, and contract-based accounts can already follow contract-defined rules. Further account-specific behavior should wait until the verification-script scope is settled.

PR #231, opened 2026-05-11
Nep 33 -> Update from NEP 23

shargon:
Updated NEP number from 23 to X and added compatibility note for NEP-23. Modified the description for error code -510 and added new error codes -514 and -515.

**When is accepted, we can preserve the nep-23 file, like this is easier for check the differences.**


nep-23X.mediawiki
```diff
@@ -157,13 +158,17 @@ used by compliant implementations.
...
 |-
 | -511 || Insufficient funds || Sender doesn't have enough GAS to pay for all currently pooled transactions.
 |-
 | -512 || Invalid verification function|| Contract doesn't have a <code>verify</code> method or this method doesn't return proper value.
 |-
 | -513 || Conflicts          || There is a conflict with another transaction already in memory pool caused by one or both of them having a Conflicts attribute.
+|-
+| -514 || Expired            || Transaction's ValidUntilBlock value is already in the past.
+|-
+| -515 || NotYetValid        || Transaction's ValidUntilBlock value is in future.
```
ajara87: The description for -515 just says "ValidUntilBlock value is in future" but that's true for any unconfirmed transaction. The description should be more precise something like "Transaction's ValidUntilBlock value exceeds the maximum allowed forward window (height + MaxValidUntilBlockIncrement)  making it not yet eligible for inclusion."

nep-23X.mediawiki
```diff
@@ -157,13 +158,17 @@ used by compliant implementations.
 |-
 | -509 || Invalid size       || Transaction or its script is too big.
 |-
-| -510 || Expired            || Transaction's ValidUntilBlock value is already in the past or NotValidBefore attribute value is in future.
+| -510 || Expired *          || Transaction's ValidUntilBlock value is already in the past or NotValidBefore attribute value is in future. Note: this code was moved to -514.
```
ajara87: `-510` covered two cases: `ValidUntilBlock` in the past AND `NotValidBefore` attribute in the future. The new `-515` only mentions `ValidUntilBlock`. What error code should nodes emit when a `NotValidBefore` attribute check fails? Should there be a `-516`?

ajara87: Is -510 reserved (never emit again) or deprecated (may still emit for compatibility)?

superboyiii:
I suggest not moving `Expired` from `510` to `514`. Exchanges are already using that for production, we should keep it compatible. Just modify the remark is OK. Then `514` for `NotYetValid`.

AnnaShaleva (commented):
> I suggest not moving Expired from 510 to 514

Makes sense. But then from 510 code the user can't really distinguish is it "the old" 510 error combining Expired&NotYetValid cases or "the new" 510 error that only matches the Expired case.

nep-23X.mediawiki
shargon: done

nep-23X.mediawiki
shargon: @copilot fix this

superboyiii:
> > I suggest not moving Expired from 510 to 514
> 
> Makes sense. But then from 510 code the user can't really distinguish is it "the old" 510 error combining Expired&NotYetValid cases or "the new" 510 error that only matches the Expired case.

"The new  510" is a subset of "the old 510", so it should be safe. They just need to know "the new 514" and to see if they need handle it. As I know they don't handle future tx before, so it's not a problem. `510` is still `510` as `Expired`.

superboyiii:
Need modify these parts.
b74e064dedd97f5dda5af9d86165e123

superboyiii:
I don't see any code on this, are we missing?
8950dece5e84b0d95688774572151095
Proposal and code are inconsistent? Maybe we need delete this part of description.
6462014c8fc6962622d4cbfe459ab480

ajara87:
> I don't see any code on this, are we missing? 8950dece5e84b0d95688774572151095 Proposal and code are inconsistent? Maybe we need delete this part of description. 6462014c8fc6962622d4cbfe459ab480

https://github.com/neo-project/proposals/pull/231#discussion_r3223291798

superboyiii:
> 
> [#231 (comment)](https://github.com/neo-project/proposals/pull/231#discussion_r3223291798)

It's another story. `NotValidBefore` is an existed attribute mostly for Notary. But I don't see any relationship with `510` in our current code (C#) which makes me confused.

Jim8y (changes requested):
The current document is still published as `nep-23X.mediawiki` with the header `NEP: X`, while the PR title indicates this is NEP-33. If this merges as-is, the proposals repository will contain a final NEP with an unresolved placeholder number and a non-standard filename. Please rename the file to the assigned NEP number and update the header before merge.

Issue #234, opened 2026-06-20
CompoundType (Array/Map/Struct) and complex VM objects should support binary serialization (JSON support exists)


cschuchardt88:
### Description

Neo.VM already provides **JSON serialization** for all `StackItem` types, including full support for `CompoundType` (`Array`, `Map`, `Struct`, etc.). This is great for debugging and logging.

**What’s missing** is native **binary serialization** for compound and complex VM objects.

### Why binary serialization is important

- **Performance & size** — Binary is dramatically more efficient than JSON for storage, snapshots, network transfer, test fixtures, tooling, indexers, and high-volume use cases.
- **Determinism** — A canonical binary format ensures consistent round-tripping, which is critical for a blockchain VM.
- **Ecosystem consistency** — Primitives already expose easy binary views (`GetSpan()` / `Memory`). Extending this cleanly to compounds would eliminate the need for everyone to write their own recursive serializers.

A working, high-quality implementation of binary serialization for all VM types (including compounds) already exists and works nicely in https://github.com/Rapid-Loop/neo-platform. (_Also works on circular references_)

### Current pain point in official VM

Operations on `CompoundType` (e.g. anything involving `CAT`, slicing, or similar byte-oriented behavior) currently throw `InvalidCastException` in the official VM. This makes it harder to treat compounds in flexible ways and complicates full-featured serialization/deserialization support.

### Proposed solution

Add native binary (de)serialization to the core `neo-project/neo-vm`:

- `byte[] ToBinary(ExecutionEngineLimits? limits = null)`
- `static StackItem FromBinary(ReadOnlySpan<byte> data, ExecutionEngineLimits? limits = null, ...)`  
  (or streaming `Serialize`/`Deserialize` variants)

**Requirements**:
- Versioned, deterministic binary format with type tags and proper recursion
- Full support for nested compounds, `Null`, limits, reference counting, etc.
- Comprehensive tests for round-tripping complex structures
- Keep the existing JSON support untouched

This would be a major improvement for anyone embedding or building tools on top of Neo.VM.

PR #235, opened 2026-06-20
Neo vm/add/binary serialization compound types

cschuchardt88:
# NEP: <to be assigned>

**Title:** Binary Serialization for CompoundType and Complex StackItem in NeoVM  
**Author:** cschuchardt88@gmail.com  
**Discussions-To:** https://github.com/neo-project/proposals/issues/234  
**Status:** Draft  
**Type:** Standard  
**Created:** 2026-06-20  
**Requires:**

## Abstract

This NEP defines a canonical, deterministic, versioned binary serialization format for all `StackItem` types in NeoVM, with full native support for `CompoundType` (`Array`, `Map`, `Struct`) and other complex objects.

It reuses the existing `StackItemType` byte values as type tags and adds `ToBinary()` / `FromBinary()` methods to the public API.

## Motivation

JSON serialization already exists, but binary is needed for performance, compactness, determinism, and tooling.

The current lack of native binary support for compounds forces duplication and makes operations like `CAT` or slicing on compounds throw `InvalidCastException`.

A standardized format aligned with the existing type system solves this cleanly.

### API Additions

Add to `Neo.VM.Types.StackItem`:

```csharp
public byte[] ToBinary(ExecutionEngineLimits? limits = null);
public ReadOnlyMemory<byte> ToBinaryMemory(ExecutionEngineLimits? limits = null);

public static StackItem FromBinary(
    ReadOnlySpan<byte> data,
    ExecutionEngineLimits? limits = null,
    IReferenceCounter? referenceCounter = null);
```

Streaming variants with `BinaryWriter`/`BinaryReader` are recommended for large objects.

### Binary Format Specification (Version 1)

All multi-byte integers are **big-endian**.

#### Overall Header

```
[Version: uint8 = 0x01] [Root Item Payload]
```

#### Item Encoding

Every item starts with its **StackItemType** byte (from `Neo.VM.Types.StackItemType`):

```csharp
public enum StackItemType : byte
{
    Any = 0x00,
    Pointer = 0x10,
    Boolean = 0x20,
    Integer = 0x21,
    ByteString = 0x28,
    Buffer = 0x30,
    Array = 0x40,
    Struct = 0x41,
    Map = 0x48,
    InteropInterface = 0x60,
}
```

**Detailed Payloads:**

- **Null / Any (0x00)** — No additional data (treated as Null for serialization).

- **Pointer (0x10)** — Not serializable (throw on attempt) or serialize as a special marker (TBD — recommended to disallow).

- **Boolean (0x20)**  
  `[0x20] [Value: uint8]` — 0 = false, 1 = true.

- **Integer (0x21)**  
  `[0x21] [Length: uint8 (0-32)] [Data: Length bytes]`  
  - Minimal big-endian two's complement representation.  
  - Zero = length 0 (no data bytes).  
  - No leading zero bytes (except for zero).

- **ByteString (0x28)**  
  `[0x28] [Length: uint32] [Data: Length bytes]`

- **Buffer (0x30)**  
  Same as ByteString: `[0x30] [Length: uint32] [Data: Length bytes]`

- **Array (0x40)**  
  `[0x40] [Count: uint32] [Item 0] ... [Item Count-1]` (recursive)

- **Struct (0x41)**  
  Same as Array: `[0x41] [Count: uint32] [Item 0] ...`

- **Map (0x48)**  
  `[0x48] [Count: uint32] { [Key Item] [Value Item] } × Count`

**Map Determinism:**
- Pairs **must** be serialized sorted by the **serialized binary form of the key** (lexicographical byte order).
- This guarantees the same map always produces identical output.

- **InteropInterface (0x60)** — Not serializable (throw exception).

#### Limits & Validation (Mandatory)

Deserialization **must** enforce `ExecutionEngineLimits`:
- Total size
- Nesting depth (recommended: configurable, default 128+)
- Number of items created
- Integer size ≤ 32 bytes

Reject malformed input immediately (invalid type, length overrun, etc.).

## Rationale

Reusing `StackItemType` bytes ensures consistency with the existing VM type system and simplifies implementation.

The format is:
- **Simple & fast** to parse
- **Deterministic** (especially maps)
- **Safe** (strict limits and early validation)
- **Extensible** (version byte + reserved types)

This design mirrors successful patterns in other Neo implementations while making it official and canonical.

## Backwards Compatibility

Fully additive. No impact on existing code, JSON serialization, or VM execution.

## Test Cases

- All primitive types round-trip
- Deeply nested compounds
- Map key ordering determinism
- Limit exhaustion cases
- Malformed data rejection
- Performance benchmarks (binary vs JSON)
- Cross-language compatibility tests

## Implementation

Reference implementation exists in https://github.com/Rapid-Loop/neo-platform. The official version should live in `neo-project/neo-vm` with full tests and documentation of this wire format.

This NEP serves as the authoritative specification for the binary format.

### DIFF
--- a/nep-XXX-binary-serialization-compoundtype.mediawiki
+++ b/nep-XXX-binary-serialization-compoundtype.mediawiki
@@ -1,0 +1,75 @@
+NEP: <to be assigned>
+Title: Binary Serialization for CompoundType and Complex StackItem in NeoVM
+Author: cschuchardt88@gmail.com
+Discussions-To: https://github.com/neo-project/proposals/issues/234
+Status: Draft
+Type: Standard
+Created: 2026-06-20
+Requires:
+
+== Abstract ==
+
+This NEP proposes adding native binary serialization and deserialization support for all <code>StackItem</code> types in NeoVM, with full support for <code>CompoundType</code> (<code>Array</code>, <code>Map</code>, <code>Struct</code>) and other complex VM objects.
+
+While JSON serialization is already available, binary serialization provides a compact, deterministic, and high-performance alternative that is essential for state snapshots, tooling, testing, cross-process communication, and efficient host applications.
+
+== Motivation ==
+
+Primitives in NeoVM (<code>ByteString</code>, <code>Integer</code>, etc.) already expose easy binary representations via <code>GetSpan()</code> and <code>Memory</code>. However, <code>CompoundType</code> and nested complex objects have no equivalent native binary format.
+
+This creates several problems:
+* Every consumer (Neo core, Neo Express, debuggers, indexers, testing frameworks, custom hosts) must implement its own recursive serializer.
+* JSON is convenient but verbose, slow for large structures, and less suitable for deterministic storage or high-performance scenarios.
+* Certain operations on <code>CompoundType</code> currently throw <code>InvalidCastException</code> (e.g. behaviors related to <code>CAT</code> or slicing), making it difficult to support flexible and complete serialization.
+
+A working high-quality implementation already exists and performs well. Adding this capability to the official <code>neo-project/neo-vm</code> would eliminate duplication and provide a canonical, limit-aware, deterministic binary format.
+
+== Specification ==
+
+Add the following methods to the <code>StackItem</code> base class (with proper overrides in <code>CompoundType</code>, <code>Array</code>, <code>Map</code>, <code>Struct</code>, etc.):
+
+* <code>byte[] ToBinary(ExecutionEngineLimits? limits = null)</code>
+* <code>ReadOnlyMemory<byte> ToBinaryMemory(ExecutionEngineLimits? limits = null)</code>
+* <code>static StackItem FromBinary(ReadOnlySpan<byte> data, ExecutionEngineLimits? limits = null, IReferenceCounter? referenceCounter = null)</code>
+
+A streaming variant using <code>BinaryWriter</code> / <code>BinaryReader</code> is also recommended for very large objects.
+
+The binary format must be:
+* Versioned (with a clear header)
+* Deterministic
+* Length-prefixed and recursive for nested compounds
+* Respect <code>ExecutionEngineLimits</code> (depth, size, etc.)
+* Handle <code>Null</code>, reference counting, immutability flags, and all current <code>StackItemType</code> values
+* Include type discriminators
+
+The format should be documented in detail (similar to how NEF or storage serialization is documented) so that other language implementations can interoperate.
+
+== Rationale ==
+
+Binary serialization is a natural extension of the existing primitive binary access patterns. It closes a clear gap in the VM object model without affecting existing JSON support or contract execution semantics.
+
+Design decisions:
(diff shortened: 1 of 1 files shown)


shargon:
We already have `BinarySerializer.Serialize` so we should use the same format

cschuchardt88:
> We already have `BinarySerializer.Serialize` so we should use the same format

Yes, that's correct—`Neo.dll` currently depends on `Neo.VM`, which creates a circular concern if we're trying to keep `Neo.VM` lightweight and independent.

I believe we could reference `Neo.IO` from the `Neo.VM` project instead (as it already contains much of the serialization infrastructure). However, I need to double-check whether all the necessary classes/interfaces required by `BinarySerializer.Serialize` (and related helpers) are fully available in `Neo.IO`. 

If any core pieces are missing, we may need to extract/move them into `Neo.IO` (or a shared layer) to avoid pulling in heavier dependencies for the VM. 

I'll verify this shortly and report back with the exact gaps (if any) and a proposed path forward.

cschuchardt88:
Thanks for the feedback and discussion on this (including the points around `BinarySerializer.Serialize` and `Neo.IO`).

After further review and confirmation of the ecosystem patterns (NEP-1 style analysis + lookups across org:neo-project repos/issues/code), **the API surface (e.g., the specific `ToBinary`/`FromBinary` methods on `StackItem`) doesn't really matter** for the core goal here.

**What truly matters is the canonical binary wire format** for the objects: the versioned, deterministic, type-tagged encoding for all `StackItem` types (primitives + full `CompoundType` support with map key sorting, limits enforcement, etc.). This ensures interoperability, determinism, performance wins, and eliminates duplication across tools, indexers, snapshots, other VMs/implementations, and the broader Neo N3/N4 stack — while staying consistent with existing patterns (including `Neo.IO` primitives for streaming/compatibility where used in the full node/ecosystem).

The API details are secondary and can evolve or be adjusted later (e.g., via extensions, streaming helpers tied to `Neo.IO.BinaryReader`/`BinaryWriter`, or delegation in higher-level serializers like `BinarySerializer` in `Neo.SmartContract`).

I'll update the PR shortly by **removing the specific API proposal section** (and any related implementation notes that tie too tightly to it). This keeps the focus squarely on the **binary format specification** as the authoritative part of the NEP. The VM can expose whatever is natural for the format, and the ecosystem can build on top cleanly.

This should also fully address the circular dependency / `Neo.IO` confirmation concerns we discussed — the wire format stands alone, with `Neo.IO` confirmed for integration in the broader stack without forcing changes to VM dependencies.

Updated PR incoming soon. Let me know if there are other adjustments needed before we move toward merging / NEP number assignment.

cschuchardt88:
Implementation PRs related to this NEP (cycle-safe compound bytes / CAT on compounds):

- https://github.com/neo-project/neo-vm/pull/594 — `GetSafeSpan` / `ComputeSpan` / `VmFeatures.CompoundSpan`
- https://github.com/neo-project/neo/pull/4742 — host `LimitsFor` keeps `CompoundSpan` off splice opcodes (draft)

